PR #4, opened 2026-07-18
fix: harden public SDK prompt and CI boundary

mldangelo-oai:
## Summary

Fix the concrete prompt, CI, and artifact-boundary issues found immediately after the TypeScript bootstrap merged:

Paired monorepo projection/source change: [openai/openai#1153264](https://github.com/openai/openai/pull/1153264).

- Address the [Codex P2 review finding](https://github.com/openai/codex-security/pull/3#discussion_r3607536516) without unsafe shell substitution: keep runtime paths out of prompt text, reject output paths that Codex would render as injected model context (including safe-looking output or custom-temp-root symlinks that resolve to unsafe canonical paths), and preserve caller `inherit`/`exclude`/custom `set` restrictions while augmenting both top-level and existing named-profile `include_only` policies with required variables, including the isolated `CODEX_HOME`. Preserve all other profile settings and never synthesize a permissive profile policy that would replace a restrictive inherited policy. Move large scoped-path input out of the environment and into a unique read-only file in the isolated Codex home; exclude Codex's default temp write roots while retaining the scan directory as the explicit writable root, and restore owner-read permission even under a restrictive `umask`. Serialization occurs before final no-follow directory validation, and only the file path is forwarded. One helper invocation now inventories all requested files/directories (including explicit `Dockerfile`, lockfile, and binary targets with empty previews) without fully reading large binaries, skips file symlinks and candidates that resolve outside the repository, and safely binds overlapping/empty scopes into manifest and coverage without printing repository-controlled names; legacy `--scope ~/...` expansion remains supported while `--scopes-file` names stay literal, and diff/working-tree scans use resolved commit IDs.
- Fix the flaky fake-login/process boundary exposed by the post-merge Ubuntu run: preserve the existing Windows inherited-pipe fallback, make login fakes exit deterministically, tolerate platform-specific pipe-close errors, and cap the CI Test step at 10 minutes so a future leaked handle cannot wedge the workflow.
- Close artifact-boundary gaps found during prepublish review: reject malformed tar directory entries including ustar-prefix collisions, internal Slack/Linear/GitHub/local-path/license/registry markers, quote/punctuation-delimited links and go-links, UTF-16/plain or compressed payloads, trailing streams, and oversized gzip/Brotli expansion. Keep the checker compatible with the paired Copybara whole-tree content guard without weakening marker detection. Pin the single permitted PNG to the known-public logo SHA-256, closing malformed/encoded-metadata and pixel-payload bypasses without maintaining a partial PNG parser.

## Validation

- `pnpm run types`, formatting, build, npm pack, executable launcher smoke, and package inspection pass; the artifact contains the expected **146 entries**.
- Full projected SDK suite: **148 passed, 1 intentional integration skip, 0 failed**; the Python-backed helper regression uses `-B` and leaves no bytecode in the package tree.
- The exact Ubuntu-failing test, `clears a cached API key after successful ChatGPT login`, passes **10 consecutive attempts**. Auth cleanup, API, and source packaging/workflow regressions pass.
- Prompt/helper regressions cover comma/newline/Unicode and literal `~name` filenames, unsafe output-directory rejection, a **>128 KiB scoped target list**, restrictive `include_only`/`umask`, default-temp-root exclusion, external and in-repository file-symlink exclusion, post-write scan-directory symlink swaps (10 consecutive passes) and completed/failed inventory cleanup, a real POSIX-shell round trip, combined explicit file/directory scopes, sparse **256 MiB binaries** including source-like files without a full read even when a directory scope encounters them first, overlapping/empty manifest scopes, and valid Unicode Git refs. Artifact regressions reject prior tar/PNG PoCs, UTF-16/plain or compressed text, UTF-16LE/BE inside Base64 or percent encodings, every short single-wrap split, literal tabs/CR and JSON-escaped CRLF, recursively nested Base64, nested percent/Unicode/octal escapes, decoder depth/breadth/byte exhaustion, tampered/truncated/appended PNG payloads via the exact-logo pin, concatenated Brotli members, and oversized expansion.

The canceled post-merge run was [29636505450](https://github.com/openai/codex-security/actions/runs/29636505450): Windows/macOS passed; Ubuntu timed out in the login test and remained alive until canceled. No package is published by this change.


mldangelo-oai:
@codex review

ianw-oai (approved):

PR #9, opened 2026-07-22
test: make CLI tests Windows-compatible

mldangelo-oai:
## Summary

Validates the Windows fixes for the first CopyBerry projection run, which failed three CLI tests on `windows-latest`:

- set both `HOME` and `USERPROFILE` when testing MCP registration so Node resolves the isolated home on every platform;
- assert the JSON-escaped native skill path instead of a POSIX-only path;
- pass the ESM preload to Node as a `file:` URL and keep the complete exit/stdout/stderr contract visible on failure.

The source-of-truth change will land in `openai/openai` and flow back through CopyBerry.

Failing run: https://github.com/openai/codex-security/actions/runs/29960300785

## Test plan

- [x] `bun test --timeout 30000 tests-ts/cli.test.ts` (66 passed)
- [x] `prettier --check tests-ts/cli.test.ts`
- [ ] `node-ci` on Linux, macOS, and Windows

### DIFF
--- a/sdk/typescript/tests-ts/cli.test.ts
+++ b/sdk/typescript/tests-ts/cli.test.ts
@@ -14,6 +14,7 @@ import {
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { Writable } from "node:stream";
+import { pathToFileURL } from "node:url";
 import { describe, expect, test } from "bun:test";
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
@codex review

ianw-oai (approved):
Looks good.

PR #15, opened 2026-07-27
fix: pin customer container images and restore SDK CI

GrantBirki:
## Summary

Pin the Dockerfile frontend and both Node 22 build and runtime stages to verified, multi-platform SHA-256 image digests.

The latest public-repository projection also omitted the candidate-normalization script required by the published plugin contract and reverted the SDK's cross-platform API, authentication, and Git-hook regression tests. Restore those exact previously green files so the packaged CLI, native Linux container, and Ubuntu, macOS, and Windows SDK checks all exercise the shipped plugin successfully.

## Validation

- All four exact-head GitHub checks passed: the native Linux container and the Ubuntu, macOS, and Windows Node 22 SDK matrix.
- 344 SDK tests passed; the remaining three skips are expected.
- Node 22 generated-model checks, TypeScript, and Prettier passed.
- Built and smoke-tested the pinned customer container and its bulk-scan command.
- Verified that the installed package contains all 94 required bundled plugin files.
- Confirmed the restored public tree differs from the last fully green main only in the three immutable Dockerfile digest pins.

## Canonical backport

[openai/openai#1195368](https://github.com/openai/openai/pull/1195368) backports the identical three pins to the source-owned public Dockerfile.

### DIFF
--- a/.github/workflows/container-ci.yml
+++ b/.github/workflows/container-ci.yml
@@ -1,0 +1,155 @@
+name: container-ci
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - .dockerignore
+      - .github/workflows/container-ci.yml
+      - Dockerfile
+      - Dockerfile.dockerignore
+      - compose.yaml
+      - docker/**
+      - sdk/typescript/**
+  pull_request:
+    paths:
+      - .dockerignore
+      - .github/workflows/container-ci.yml
+      - Dockerfile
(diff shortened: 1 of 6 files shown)


mldangelo-oai (dismissed):
Thank you

mldangelo-oai:
@codex review

mldangelo-oai (approved):

PR #17, opened 2026-07-28
release: publish protected multi-architecture Codex Security images

dc-oai:
## Summary

Prepare the official `0.1.1` Codex Security image for externally distributed, noninteractive, resumable CSV repository scans.

- Preserve the existing supported public package version; do not overwrite the separately merged `0.1.1` SDK release.
- Build and validate `linux/amd64` and `linux/arm64` on native runners.
- Keep pull request validation registry-free; image publication cannot run from a pull request.
- Preserve the normal no-host-setup Compose workflow and the existing customer-facing security controls: nonroot execution, dropped Linux capabilities, `no-new-privileges`, narrowed seccomp, private mounted authentication, and persistent scan results.
- Provide a separately selectable, enforcing AppArmor profile and Compose override for restricted Ubuntu hosts. Do not require the override on other hosts.
- Keep Codex’s preferred sandbox when the explicit AppArmor profile is active; retain the compatibility fallback only when it is actually required.
- Bind a clean pinned repository scan to its authoritative Git revision without weakening manifest, coverage, target, or workbench validation.
- Keep interactive repository discovery out of scope.

## Protected publication

- Require the source commit to belong to protected `main`, and require the image tag to match the exact public CLI version.
- Fail closed before any platform image is pushed unless `ghcr.io/openai/codex-security` has been created, made publicly readable, and granted access to this repository.
- Require independent approval through the protected `container` environment.
- Verify anonymously pullable platform images and the multi-architecture candidate before promotion.
- Generate provenance in a separate, minimal-permission job before promoting the versioned, commit, and `latest` tags.
- Never overwrite an existing stable version.

## Verification

- All public checks pass on Linux `amd64`, Linux `arm64`, Ubuntu, macOS, and Windows.
- The companion monorepo Buildkite run is green: **380 SDK tests**, **346 Copyberry integration tests**, and an actual clean public projection all pass.
- Native restricted and unrestricted Ubuntu hosts verify the intended hardened customer sandbox and Compose paths.
- Native Linux customer regression suite: **365 passed, 3 expected platform/integration skips, 0 failed**.
- Independently built and clean-installed the exact `@openai/codex-security@0.1.1` tarball; verified the public SDK import, installed CLI, all **94 bundled scanner-plugin files**, all **178 expected package entries**, and absence of unexpected internal content.
- Release and pull request workflows pass `actionlint`; TypeScript and formatting validation run against the exact `0.1.1` package.
- Fresh authenticated pinned-repository, restricted-Ubuntu full-scan verification is being checked separately; do not interpret sandbox-only success as a completed scan.

Opening or merging this PR does **not** create the GHCR package or publish an image. First publication still requires an authorized package administrator and independent protected-environment approval.

Source-of-truth companion: https://github.com/openai/openai/pull/1199105.


.github/workflows/container-release.yml
```diff
@@ -0,0 +1,407 @@
...
+
+      - name: Verify customers can pull without GitHub credentials
+        env:
+          IMAGE: ${{ needs.authorize.outputs.image }}
+          VERSION: ${{ needs.authorize.outputs.version }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          docker logout ghcr.io
+          if ! docker pull "$IMAGE:$VERSION"; then
```
mldangelo-oai: Could we explicitly document or preflight the first-time public-package bootstrap here? Newly created GHCR packages start out private, so an initial publication can reach this anonymous pull only after platform images and tags have already been pushed, then fail. Making the visibility/bootstrap prerequisite explicit before promotion would give the first release a clear, predictable success path.

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,407 @@
...
+          references=()
+          for digest in "${digest_files[@]}"; do
+            if [[ ! "$digest" =~ ^[[:xdigit:]]{64}$ ]]; then
+              echo 'The release contains an invalid platform digest.' >&2
+              exit 1
+            fi
+            references+=("$IMAGE@sha256:$digest")
+          done
+
+          docker buildx imagetools create \
```
mldangelo-oai, line 462: Would it make sense to create or select the manifest by digest first, run attestation and the public-pull/Compose checks against that candidate, and promote the version and latest tags only after everything succeeds? At the moment a later failure can leave the user-facing tags published even though the release failed.

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,407 @@
...
+      - compose.yaml
+      - docker/**
+      - sdk/typescript/**
+  push:
+    tags:
+      - "container-v*"
+  workflow_dispatch:
+
+concurrency:
+  group: container-release-${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.ref }}
```
mldangelo-oai: Small release-race concern: the concurrency key uses the ref, so a manual main run and a version-tag run can both publish to the same image/version/latest at the same time. Could we use a shared repository/image-level group for non-PR publication, and avoid overwriting an already published version? Promptfoo’s public Docker workflow uses a shared publishing group for this reason.

.github/workflows/container-release.yml
```diff
@@ -0,0 +1,407 @@
...
+          compression-level: 0
+
+  manifest:
+    name: publish-and-verify-multiarchitecture-image
+    needs:
+      - authorize
+      - publish-platform
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+    permissions:
```
mldangelo-oai, line 405: Could we move provenance generation into its own minimal job, passing in the verified manifest digest as an output? That would keep provenance-signing permissions separate from repository checkout, registry promotion, and Compose execution, and follows the narrowly scoped attestation-job pattern in Promptfoo’s public workflow.

mldangelo-oai (commented):
Thanks for putting this together — there is a lot to like here, especially the native amd64/arm64 validation, SHA-pinned actions, digest-based platform builds, SBOM generation, and the updated host-aware AppArmor fallback with regression coverage.

I took another pass on the latest head and compared the publicly visible release flow with Promptfoo’s public Docker workflow. I left a few inline suggestions that I think would make the first release safer and more predictable: explicitly handle first-time GHCR visibility, only promote stable image tags after attestation and smoke checks have passed, serialize all publication paths while protecting version-tag stability, and keep provenance signing in a small dedicated job.

As an optional follow-up, a deterministic successful bulk-scan smoke test would also provide stronger coverage alongside the existing empty-CSV regression.

Really appreciate the attention to hardened defaults and the improvement to sandbox selection — this is close, and I think the small release-flow adjustments would make it much more robust.

dc-oai:
Thanks, Michael — addressed the GHCR visibility preflight, verify-and-attest-before-promotion flow, shared release concurrency and immutable version tags, and dedicated least-privilege signing job. I’ll keep the successful bulk-scan smoke test as a follow-up.

README.md
```diff
@@ -48,4 +48,45 @@ console.log(result.reportPath);
 await security.close();
 ```
 
+## Containerized bulk scans
```
mldangelo-oai, line 66: can we move most of this to the developer docs

README.md
```diff
@@ -48,4 +48,45 @@ console.log(result.reportPath);
...
+
+Create `repositories.csv` with one full, immutable Git commit per repository:
+
+```csv
+id,repository,revision
+payments,https://github.com/example/payments.git,0123456789abcdef0123456789abcdef01234567
+```
+
+Once the approved image has been published, prepare private results and
```
mldangelo-oai: double check this

mldangelo-oai (dismissed):
Please see the comments on the docs. Willing to give this a try. Thank you so much @dc-oai

mldangelo-oai:
@codex review

mldangelo-oai (dismissed):
Reviewed exact head 263059affc74b66eaf0e6e37ee04645e942029b9. Preserves current scan-target protections, fails closed on GHCR version lookup errors, keeps immutable image tags, moves detailed container instructions into developer documentation, and builds Codex Security 0.1.4. Verified 478 passing tests, workflow/shell lint, container build, nonroot execution, and hardened-host sandbox fallback.

mldangelo-oai:
@codex review

mldangelo-oai (approved):
Reviewed exact head 00c61201141bf03816c23930e7416e36d5c67013, including both Codex findings. Each published amd64/arm64 image is now smoke-tested by its exact immutable digest on its native runner; restricted Ubuntu hosts preserve existing compatible Landlock overrides and reject incompatible values. The release-version preflight remains fail-closed, main target protections are preserved, and all focused regression tests and workflow/shell/type checks pass.

PR #19, opened 2026-07-28
fix: make ChatGPT/API-key auth conflicts actionable

mldangelo-oai:
## Problem

A successful ChatGPT login can be silently overridden by an ambient `OPENAI_API_KEY` or `CODEX_API_KEY`. The subsequent scan then reports a generic authentication or model-access error, and the existing recovery hint is hidden in non-interactive terminals. This makes users think that ChatGPT login is broken even though a different credential was actually selected.

## Change

- Warn immediately after successful ChatGPT or device-code login when an environment API key will take precedence.
- Identify the exact effective environment-variable source in `401` and `403` scan failures.
- Include the ChatGPT recovery instructions in the final failure message so they are visible in CI and other non-TTY environments.
- Preserve existing environment-key precedence, intentional API-key login, stored-credential behavior, existing exit codes, and credential/provider-context redaction.
- Add targeted regressions for both key variables, normal and device-code ChatGPT login, unsuccessful login, and non-interactive authentication and model-access failures.

This is intentionally limited to the P0 authentication issue: no TAC gate, policy changes, token-refresh write-back, cost changes, or unrelated documentation changes.

## Verification

- Full public-CI test suite using the pinned Bun `1.3.14`: **348 passed, 3 expected platform/integration skips, 0 failed**.
- Focused authentication suite: **11 passed, 0 failed**.
- Full TypeScript typecheck and SDK/CLI production build.
- Package-wide Prettier and generated-model checks.
- Packed and validated the actual public npm tarball: **178 validated entries**.
- Successfully installed the packed release and verified public SDK import, executable CLI, and all **94 bundled plugin files**.
- Eight additional compiled, zero-network authentication regressions covering both keys, `401`, `403`, stored credentials, and device-code login.


ianw-oai (dismissed):

daneschneider-oai:
Superseded by #22, which is now merged. The subprocess-level effective-API-key regression from this PR is preserved in follow-up #23, alongside the access-token login warning.

ianw-oai (dismissed):

sdk/typescript/src/cli.ts
```diff
@@ -1450,10 +1446,16 @@ export async function main(
...
               .map(([name]) => name);
+            const loginKind = options.withAccessToken
+              ? "Access-token"
+              : "ChatGPT";
+            const credential = options.withAccessToken
+              ? "access token"
+              : "ChatGPT sign-in";
             errorOutput.write(
-              "ChatGPT login succeeded. Interactive scans will ask which account to use; " +
+              `${loginKind} login succeeded. Interactive scans will ask which account to use; ` +
```
zcrab-oai, line 1456: Non-blocking concern: with `OPENAI_API_KEY` or `CODEX_API_KEY` set, a successful `login --with-access-token` says an interactive scan will ask which account to use. However, that scan only prompts when `hasStoredChatGPTSignIn()` sees `chatgpt` in `codex login status`. Codex reports access-token credentials as `Logged in using access token` or `Logged in using personal access token`, so the prompt never appears and the environment API key is used instead. Could we avoid promising an interactive choice for access-token logins, or teach the credential check to recognize access-token logins? The access-token-specific wording in #23 already addresses this. Flagging for visibility only; this is not blocking my approval.

zcrab-oai (approved):
Approved. I left one non-blocking inline note about the access-token login warning promising an interactive credential choice that the current ChatGPT-only status check does not provide. The access-token-specific wording in #23 already addresses the concern.

sdk/typescript/src/cli.ts
mldangelo-oai: Thanks we'll resolve in 23

PR #64, opened 2026-07-29
fix(windows): normalize canonical scan path casing

hoquanganh09:
## Summary

- Improve handling of mixed-case Windows paths in Deep Scan, scan-finalization, and workbench canonical-path checks.
- Keep scan finalization fail-closed: the earlier fallback for missing manifests, findings, and coverage was removed after review.
- Continue the platform-safe implementation and cross-platform regression coverage in #85, which preserves both original contributor-authored commits.

## Maintainer follow-up

Thank you to @hoquanganh09 for the original contribution and for incorporating the early review feedback. The maintainer follow-up in #85 preserves that work, adds explicit Windows path-casing and POSIX symlink-security regression tests, and carries the change forward onto the latest main branch.


hoquanganh09:
Thanks for the automated review!

I have updated the PR to focus strictly on the core **Windows Path Case-Insensitive Normalization Fix**. 

- Removed the fallback draft JSON synthesis from \_prepare_scan_finalization\ so incomplete or crashed scans properly fail instead of sealing zero-finding results.
- The PR now cleanly fixes the Windows canonical directory checks (\
esolved.as_posix().lower() != scan_dir.as_posix().lower()\) across \workbench_db.py\, \inalize_scan_contract.py\, and \deep_scan_workbench.py\.

mldangelo-oai (commented):
Thank you for the contribution, and for taking the time to remove the synthetic-manifest fallback after the earlier feedback! I reviewed the updated head and reproduced one security issue that needs to be addressed before this is ready.

On a case-sensitive filesystem, changing native `Path` equality to `.as_posix().lower()` allows a symlink such as `/scan/prompt -> /scan/Prompt` through `deep_scan_path`. I checked both implementations: the current base correctly rejects it, while this PR accepts it. The same unconditional case folding also weakens the canonical directory, artifact, and output-parent protections in the other two files.

Python already handles Windows path casing in native comparisons: `PureWindowsPath("D:/Repo/Scan") == PureWindowsPath("d:/repo/scan")` evaluates to `True`. Could you retain native `Path` equality or, if there is a specific Windows failure requiring normalization, use platform-aware `os.path.normcase(os.fspath(path))`? Regression tests for both the mixed-case Windows path and the case-differing POSIX symlink would make the fix safe across platforms.

Once addressed, I would be happy to take another look. Thank you again for working on this!

mldangelo-oai:
Thank you again for the contribution and for quickly addressing the original scan-manifest feedback! I prepared #85 as a maintainer follow-up so we can run the complete Windows, macOS, Linux, and container checks from the OpenAI repository. It retains both of your original commits and adds the platform-safe path normalization and Windows/POSIX symlink regression tests we discussed. I also updated this PR description to reflect the current scope and make sure your contribution remains clearly credited. Really appreciate you working on this!

mldangelo-oai (approved):

PR #65, opened 2026-07-29
fix: include personal account in bulk-scan discovery

### LINKED ISSUE #28: bulk-scan omits the personal account when the user belongs to an organization
## Summary

Interactive `bulk-scan` discovery cannot select repositories owned by the authenticated personal account when that user belongs to one or more GitHub organizations.

`selectGitHubOwner` first calls `GET /user/orgs`. If exactly one organization is returned, it selects that organization automatically. If multiple organizations are returned, the prompt contains only those organizations. `GET /user` is called only when the organization list is empty, so the personal account is unreachable for users with any organization membership.

## Environment

- `@openai/codex-security` 0.1.1 / current `main` at `e94d6bef9797a192febfde89a26ec7f831bc09b2`
- macOS
- Bun 1.3.14

## Reproduction

1. Authenticate `gh` with a GitHub user that owns at least one personal repository and belongs to at least one organization.
2. Run `npx codex-security bulk-scan`.
3. Attempt to select the personal account as the repository owner.

With one organization, discovery immediately uses that organization. With multiple organizations, only organization names are offered.

A targeted test using the existing discovery test harness also reproduces this: configure `organizations: ["acme"]`, request `personal-account`, and assert that `/user` is queried and the GraphQL owner is `personal-account`. The assertion fails because `/user` is never queried; all ten existing discovery tests still pass.

## Expected behavior

The authenticated personal login should be included alongside organization logins. If the combined list contains more than one owner, the wizard should prompt for the account or organization; if it contains only the personal login, it can select it automatically.

## Relevant code

https://github.com/openai/codex-security/blob/main/sdk/typescript/src/bulk-scan-discovery.ts#L195-L215

The current test explicitly covers prompting only when several organizations exist, but does not cover a personal account plus organization membership:

https://github.com/openai/codex-security/blob/main/sdk/typescript/tests-ts/bulk-scan-discovery.test.ts#L336-L350

JustasMonkev:
## Summary

- Include the authenticated personal GitHub account alongside organization owners.
- Prompt for an owner when both personal and organization accounts are available.
- Add regression coverage for selecting the personal account.

Fixes #28.

### DIFF
--- a/sdk/typescript/src/bulk-scan-discovery.ts
+++ b/sdk/typescript/src/bulk-scan-discovery.ts
@@ -197,22 +197,20 @@ async function selectGitHubOwner(
   prompt: BulkScanPrompt,
   signal?: AbortSignal,
 ): Promise<string> {
-  const { data } = await github.request("GET /user/orgs", {
-    per_page: 100,
-    request: { signal },
-  });
-  const organizations = data.map(({ login }) => login).sort();
-  if (organizations.length > 1) {
+  const [orgs, user] = await Promise.all([
+    github.request("GET /user/orgs", { per_page: 100, request: { signal } }),
+    github.request("GET /user", { request: { signal } }),
+  ]);
+  const personal = user.data.login;
+  const owners = [personal, ...orgs.data.map(({ login }) => login)].sort();
+  if (owners.length > 1) {
     return await prompt.select(
       "Which account or organization should we scan?",
-      organizations.map((owner) => ({ label: owner, value: owner })),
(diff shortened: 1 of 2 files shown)


mldangelo-oai (approved):
Thank you for the thoughtful contribution! I verified this against the latest main, including personal-account discovery, existing organization selection, the single-account flow, CSV generation, the complete SDK test suite, and a clean packaged CLI on Linux. The fix is focused and preserves the existing organization workflow. LGTM, and thank you for helping improve Codex Security!

PR #77, opened 2026-07-29
fix: preserve scans when reusing archived output

ianw-oai:
## Summary
- Prevent `UNIQUE constraint failed: scans.scan_dir` when rerunning a scan with `--archive-existing`.
- Preserve the previous scan and its archived artifacts in the same SQLite transaction that registers the replacement.
- Refuse to displace running scans and reject unsafe archive paths.
- Forward the actual archive path from the TypeScript SDK into the bundled workbench.

## Validation
- Full public SDK suite: 375 passed, 3 expected skips.
- Four real-SQLite regression checks covering completed archives, running scans, empty failed scans, and unsafe paths.
- TypeScript, generated-model, formatting, and whitespace checks.


mldangelo-oai (dismissed):

mldangelo-oai (approved):
Validated the exact head after addressing both archived-output review findings. Artifact paths are preserved transactionally, retries cannot strand existing archived artifacts, both focused regression tests pass, and the full suite passes (389 tests; 3 expected skips). Typecheck, formatting, build, and npm package validation also pass.

PR #81, opened 2026-07-29
fix: keep completed security scan output concise

ianw-oai:
## Summary
- Print the concise completion summary for ordinary security scans instead of dumping the complete manifest and findings.
- Keep findings, coverage, elapsed time, tokens, estimated cost, report path, and results directory visible.
- Preserve complete output for explicit `--json`, `--format`, and `--full-output` requests; keep dry-run output unchanged.
- Port the matching regression coverage for explicit output formats and worker fallback.

## Validation
- Full public SDK suite: 385 passed, 3 expected platform/integration skips, 0 failed.
- CLI regression suite: 75 passed.
- TypeScript and generated-model checks.
- Prettier and committed-patch whitespace checks.


mldangelo-oai (dismissed):

mldangelo-oai (approved):
Verified the exact head after fixing both concise-scan-output review findings. Explicit equals-style output formats and token count, limit, and offset are covered by passing regression tests; all 389 local tests pass (3 expected skips), and the TypeScript typecheck, formatting, and package build are clean.

mldangelo-oai (approved):

PR #84, opened 2026-07-29
fix: harden Node CI and Windows package smoke

mldangelo-oai:
## Summary

- Extend Node CI to Node 24 and Node 26 without introducing a full operating-system-by-Node-version cross-product: preserve Node 22 on Ubuntu, macOS, and Windows and add Node 24 and Node 26 on Ubuntu.
- Install the exact project-defined pnpm release through a full-SHA-pinned action and configure separate lockfile-keyed pnpm and npm caches, so Node 26 does not rely on the removed bundled Corepack executable and package smoke tests do not repeatedly cold-download dependencies.
- Cancel only superseded pull-request CI runs, retain read-only workflow permissions, and bound Node CI jobs without canceling main-branch or release runs.
- Retain a direct SDK and CLI smoke test for each selected Node runtime.
- Fix the Windows fresh-install package smoke without dropping Windows coverage: reuse the integrity-checked npm cache with `--prefer-offline`, and use a shared platform-aware timeout contract that preserves the existing 120-second command and 150-second supervisor limits on Linux and macOS while allowing 180 and 210 seconds, respectively, on Windows.
- Disable Bun executable caching in the protected npm release to remove the identified cache-poisoning finding while preserving release provenance, `gitHead` verification, protected environments, and OIDC publishing.
- Fix the pre-existing container-workflow shellcheck warning without changing its Compose behavior.

## Root cause

The current `main` Windows runner reaches package inspection successfully, but its clean npm consumer installation can exceed the smoke script's shared 120-second subprocess deadline. Each smoke test also created and immediately discarded its own private npm cache, forcing every matrix job to cold-download the same dependency graph. The parent inspector independently terminates the smoke at 150 seconds, so increasing the inner limit alone cannot fix the failure safely.

The smoke test now keeps the consumer installation isolated while reusing the lockfile-keyed npm download cache and preferring verified cached packages. A shared timeout helper coordinates both subprocess limits and gives additional time only to Windows. Regression tests assert the exact timeout budgets, the supervisor safety margin, and the default-platform behavior.

## Validation

- Focused Windows package-smoke regression tests.
- Complete TypeScript SDK test suites on Node 22, Node 24, and Node 26.
- Type-check, formatting, build, direct SDK import, CLI version/help smoke, package creation, and complete package inspection.
- Fresh npm consumer installation, SDK import, CLI execution, and bundled-plugin validation on Node 22, Node 24, and Node 26; the warmed shared cache reduced repeated 77-package installations from approximately two minutes to six seconds.
- GitHub Actions syntax and shell analysis with `actionlint`.
- GitHub Actions security analysis with `zizmor`; no reportable findings.
- Semantic assertions confirming exactly five matrix jobs, immutable action pins, safe PR-only concurrency, least-privilege permissions, and preserved npm OIDC release settings.
- `git diff --check`.

### DIFF
--- a/.github/workflows/container-ci.yml
+++ b/.github/workflows/container-ci.yml
@@ -25,6 +25,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   container:
     name: linux-amd64
@@ -96,7 +100,8 @@ jobs:
           mkdir -p results state
           chmod 700 results state
           printf 'id,repository,revision\n' > repositories.csv
-          export CODEX_SECURITY_USER="$(id -u):$(id -g)"
+          CODEX_SECURITY_USER="$(id -u):$(id -g)"
+          export CODEX_SECURITY_USER
           docker compose config --quiet
           docker compose run --rm codex-security --version
 
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -8,14 +8,25 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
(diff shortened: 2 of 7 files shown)


ianw-oai (approved):
Looks good. The supported Node versions, Windows packaging, and package smoke tests are covered.

PR #92, opened 2026-07-29
fix: trust git when installing hooks

GautamSharma99:
## Summary

  Harden `codex-security install-hook` against repository-controlled Git execution and Git environment overrides.

  Previously, the command invoked `git` by name through `execFileSync`. When run from an untrusted checkout, a repository-controlled directory placed early in `PATH` could
  supply a malicious Git executable. That executable would run during hook installation and could return an arbitrary hook destination.

  ## Changes

  - Resolve Git through the existing trusted-executable boundary before invoking it.
  - Protect both:
    - the directory from which the CLI was invoked;
    - the target repository passed to `install-hook`.
  - Extend `resolveTrustedExecutable` to support multiple protected roots.
  - Remove inherited `GIT_*` environment variables before resolving and invoking Git.
    - This prevents variables such as `GIT_CONFIG_COUNT` and `GIT_CONFIG_KEY_*` from overriding `core.hooksPath`.
  - Invoke the canonical trusted Git path instead of the string `"git"`.
  - Validate that the path returned by `git rev-parse --git-path`:
    - is non-empty;
    - is absolute;
    - contains no control or Unicode line-separator characters.
  - Preserve legitimate repository-configured `core.hooksPath` behavior.

  ## Security impact

  This closes an arbitrary-code-execution path where an attacker-controlled `git` shim could be selected from `PATH` while installing a hook in an untrusted checkout.

  It also prevents inherited Git environment variables from redirecting the generated hook to an attacker-selected location.

  ## Tests

  Added regression coverage confirming that:

  - executable resolution removes `PATH` entries beneath every protected root;
  - the real trusted executable is selected after unsafe entries are removed;
  - inherited `GIT_CONFIG_*` variables cannot override the repository’s configured hook location.

  Verification performed:

  - `pnpm run types`
  - `pnpm run format`
  - `pnpm run build`
  - `pnpm run test`

  Full test result:

  - 396 passed
  - 3 skipped
  - 0 failed


mldangelo-oai:
@codex review exact head 92ab7abbb495ed3d242862c16dc659b402f26e58

PR #98, opened 2026-07-29
fix: publish multiscan locks atomically

GautamSharma99:
## Maintainer tracking

Related to #102. The follow-up in #196 preserves this original contributor-authored change.

  ## Summary

  Make multiscan lock publication atomic and recover incomplete or malformed legacy locks.

  Previously, lock acquisition performed two separate operations:

  1. create `.lock/`;
  2. write `.lock/owner.json`.

  A crash between those operations left a visible lock without ownership metadata. Subsequent campaigns attempted to read and parse `owner.json`
  unconditionally, permanently preventing automatic recovery.

  Truncated or structurally invalid owner files caused the same problem.

  ## Changes

  ### Atomic lock publication

  The supervisor now:

  1. creates a uniquely named pending lock file;
  2. writes its complete owner record;
  3. flushes the owner record with `fsync`;
  4. atomically hard-links it to `.lock`;
  5. removes the pending name.

  The published owner record contains:

  - the supervisor PID;
  - a unique ownership token.

  Because hard-link creation has no-clobber semantics, only one contender can publish `.lock`. The lock never becomes visible with partially
  written ownership metadata.

  Crash behavior is now:

  - before publication: no `.lock` exists, so the next campaign can start;
  - after publication: `.lock` contains a complete owner record and can be recovered using the recorded PID.

  ### Legacy lock compatibility

  Recovery continues to recognize the previous directory layout:

  ```text
  .lock/
  └── owner.json

  Legacy locks are treated as stale when ownership metadata is:

  - missing;
  - truncated or malformed JSON;
  - structurally invalid;
  - oversized;
  - associated with a process that no longer exists.

  Valid legacy locks belonging to a running process still block another supervisor.

  ### Race-safe recovery

  Lock recovery now:

  - rechecks a live owner before reporting contention;
  - atomically moves stale state aside before removing it;
  - revalidates the moved owner;
  - restores and preserves the lock if it belongs to a live process;
  - retries acquisition if another contender changes the lock concurrently.

  EPERM from process probing is treated as evidence that the owner exists rather than as a stale lock.

  ### Ownership-safe release

  A supervisor removes .lock only when the published ownership token still matches its own token.

  This prevents a finishing supervisor from deleting a replacement lock published by another process.

  Pending and stale files created during normal acquisition and recovery are cleaned up.

  ## Impact

  An interruption during lock creation can no longer permanently wedge a bulk-scan campaign.

  Users can resume normally after:

  - a process kill between old lock-directory creation and owner publication;
  - a machine restart;
  - a truncated owner.json;
  - malformed or invalid owner metadata.

  At the same time, concurrent live supervisors remain mutually exclusive.

  ## Tests

  Added regression coverage confirming that:

  - a live lock is a fully published regular file;
  - its owner record contains the current PID and a unique token;
  - no pending lock remains after publication;
  - a second live supervisor is rejected;
  - the previous directory-style lock remains recoverable;
  - missing legacy owner.json is recovered;
  - truncated owner JSON is recovered;
  - structurally invalid owner metadata is recovered;
  - stale and pending recovery files are cleaned up;
  - valid dead-owner recovery and checkout cleanup continue to work.

  Verification performed:

  - pnpm run types
  - pnpm run format
  - pnpm run build
  - focused multiscan tests
  - pnpm run test

  Full test result:

  - 408 passed
  - 5 expected platform/integration skips
  - 0 failed


GautamSharma99:
@mldangelo-oai  @ianw-oai 
I updated multiscan locking so ownership is published atomically using a fully written and synced lock file. The new flow prevents crashes from leaving incomplete locks that permanently block future campaigns. It also recovers missing, malformed, truncated, or dead-owner legacy locks, while ownership tokens and PID checks ensure active supervisors remain protected and cannot remove another process’s lock. Regression tests cover atomic publication, stale recovery, cleanup, and concurrent supervisor rejection.

mldangelo-oai:
Thank you for the original multiscan locking work. The follow-up in #196 preserves the contributor author and the same stable patch after rebasing, and is the working branch for #102. Its recovery/ownership race findings still need to be addressed, so neither branch should be merged yet; keeping this PR open preserves the original contribution and context.

mldangelo-oai:
@codex review exact head 558ad1b9925ed299f7e4c9c159d2a8f5ba9f939a

PR #101, opened 2026-07-29
fix: support managed Codex credential keyrings

mldangelo-oai:
## Summary

Fix the case where an enterprise-managed credential-store policy makes `codex-security login` succeed but a subsequent scan fail with `No credentials were found`.

- Respect Codex's native `auto` and system-keyring credential storage instead of forcing file-backed authentication.
- Share one private, persistent Codex home between login, login status, logout, scans, and finding comparisons.
- Verify keyring-backed authentication through the native Codex account status even when no `auth.json` exists.
- Keep persistent credentials read-only to the scan sandbox, reject unsafe or symlinked credential homes, and retain temporary isolation for environment-provided API keys.
- Reuse an already-installed plugin and safely upgrade it through the supported native Codex plugin commands without deleting existing credentials.
- Add regression coverage for managed authentication, sequential and parallel scans, credential isolation, API-key fallback, and upgrades with the real bundled Codex executable.

## Testing

- Full CI-matched Bun 1.3.14 suite: **419 passed, 0 failed, 5 expected platform or opt-in integration skips**.
- `node scripts/generate-models.cjs --check`.
- `tsc --noEmit`.
- Full Prettier formatting check.
- Production TypeScript build against current `main` and `@openai/codex-security@0.1.3`.
- Full packed-tarball validation, including a fresh npm consumer installation, public SDK import, executable CLI, and all 95 bundled plugin files.
- Built CLI ChatGPT-auth dry run against an isolated state directory.

## Managed-device acceptance

After installing the patched build, validate the actual enterprise-managed ChatGPT/keyring flow on an affected device:

```sh
codex-security login
codex-security login status
codex-security scan . --auth chatgpt
```

The live managed-device login and model-backed scan require an affected, authenticated device and are not represented as having been run in local automated tests.


mldangelo-oai:
@codex review

Please focus on managed and MDM-enforced keyring policies; stable Codex home identity and login, status, scan, and comparison consistency; API-key precedence; persistent credential permissions and sandbox isolation; concurrent scans; and native plugin reuse and upgrades.

sdk/typescript/src/api.ts
```diff
@@ -1043,33 +1063,45 @@ export class CodexSecurity {
...
-    const codexHome = await createIsolatedHome(temporaryRoot, validateLocation);
+    const processEnvironment = selectedScanEnvironment(
+      this.#dependencies.environment,
+      auth,
+    );
+    const persistentCredentialHome =
+      scanAuthentication(this.#dependencies.environment, auth).method ===
+      "stored_credentials";
+    const codexHome = persistentCredentialHome
+      ? await prepareCodexSecurityCredentialHome(
```
ianw-oai, line 1220: This switches scans to a persistent credential home, but the existing ambient-auth import still runs every time. It overwrites an explicit login and restores ambient credentials after logout. Import ambient credentials only when initializing an empty home; preserve an explicit logout.

sdk/typescript/src/runtime.ts
```diff
@@ -784,13 +866,60 @@ export async function bootstrapPlugin(
 ): Promise<PluginInstall> {
   const root = await realpath(pluginRoot);
   const { name, version } = await pluginMetadata(root);
-  const marketplace = await createMarketplace(codexHome, root, options.signal);
+  const existingMarketplace = join(codexHome, "sdk-marketplace");
```
ianw-oai, line 1286: Parallel scans now bootstrap the same marketplace and credential home without coordination. The auth copy also reuses a PID-and-timestamp temporary path; eight simultaneous imports all failed in reproduction. Serialize shared initialization and use unique temporary files.

ianw-oai (changes requested):
Two blockers: scans replace an explicitly selected account with ambient credentials and undo logout; concurrent scans race while preparing the shared credential home. Both are reproduced. Please preserve explicit authentication and serialize shared initialization.

ianw-oai (approved):

PR #110, opened 2026-07-29
fix: keep scan output dirs private across shared parents

### LINKED ISSUE #109: Scan output under a shared parent can be replaced before completion
## Summary

Scan output directories are created `chmod 700` and ownership-checked once at prepare time. Completion and contract loading later resolve the scan directory by path string and only require a canonical non-symlink directory.

On a shared parent that is group/world-writable without the sticky bit, another local user can rename the private directory aside and create a replacement tree at the same path before `complete-scan`. The replacement is then accepted as the completed scan.

## Expected

- Refuse `--output-dir` when an ancestor is group/world-writable without the sticky bit (same class of protection `/tmp` already provides).
- Whenever the scan directory is resolved for registration, completion, or contract load, re-assert that it remains private to the current user.

## Actual

Privacy is checked only at prepare. `require_canonical_scan_directory` / contract root loading do not re-check mode or ownership.

## Notes

- Sticky `/tmp` and private home directories are unaffected.
- Aligns with the existing “output must not be accessible to other users” rule.

## Environment

- Package: `@openai/codex-security` (main)
- Area: TypeScript SDK output prep + workbench scan-dir resolution

batmnnn:
## Summary

- Fixes #109: reject non-sticky group/world-writable output parents at prepare time.
- Re-check scan-directory privacy (mode + owner) in the workbench canonical resolver and TypeScript contract loader so a mid-scan rename/replace under a shared parent cannot supply forged completion artifacts.

## Test plan

- [x] `bun test --timeout 60000 ./tests-ts/runtime.test.ts ./tests-ts/contract.test.ts ./tests-ts/api.test.ts` (129 pass)
- [ ] Confirm CI `node-ci` passes on this branch

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sdk/typescript/_bundled_plugin/scripts/workbench_db.py
+++ b/sdk/typescript/_bundled_plugin/scripts/workbench_db.py
@@ -3469,6 +3469,16 @@ def require_canonical_scan_directory(scan_dir: Path) -> Path:
         scan_dir
     ):
         raise SystemExit("Scan directory must be an existing canonical non-symlink directory.")
+    # Re-check privacy on every resolution so a mid-scan rename/replace under a
+    # shared parent cannot substitute another user's forged artifact tree.
+    if os.name != "nt":
+        if stat.S_IMODE(metadata.st_mode) & 0o077:
+            raise SystemExit(
+                "Scan directory must not be accessible to other users (chmod 700)."
+            )
+        geteuid = getattr(os, "geteuid", None)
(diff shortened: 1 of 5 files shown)


batmnnn:
@mldangelo should I close this ?

mldangelo-oai:
Thanks for working on output-directory privacy. A stronger version of the same protection has already landed on `main`, and this older branch now conflicts with it. I'll close the PR as superseded. We'd be glad to review another contribution.

PR #112, opened 2026-07-30
fix: bind immutable diff scans to snapshot digests

createsomethingtoday:
## Summary

- compute a deterministic binary Git diff digest for immutable commit and range targets
- persist that digest during both setup and direct CLI scan registration
- require the same snapshot digest when completing and verifying scan manifests
- cover stability against unrelated untracked files and the real CLI registration path

## Why

Two real exact-commit scans completed 52-56 minutes of model work but could not save because scan.target.snapshotDigest was empty. Working-tree diffs already carried a digest; immutable commit and range diffs did not. The first implementation exposed a second missing handoff: direct CLI registration constructed the immutable target without persisting the computed digest. The added registration regression failed with an undefined digest before the fix.

## Validation

- pnpm types
- pnpm format
- pnpm test: 470 passed, 6 skipped, 0 failed
- focused immutable-digest and CLI-registration regressions passed

### DIFF
--- a/sdk/typescript/_bundled_plugin/scripts/workbench_db.py
+++ b/sdk/typescript/_bundled_plugin/scripts/workbench_db.py
@@ -97,6 +97,7 @@ from workbench_target import (
     directory_content_digest,
     directory_snapshot_regular_file_count,
     git_command,
+    git_diff_content_digest,
     git_output,
     git_revision,
     git_submodule_paths,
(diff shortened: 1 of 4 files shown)


createsomethingtoday:
A second real 52:43 exact-commit run confirmed the initial patch missed direct CLI registration: normalization and completion were fixed, but register-cli-scan still created refs targets without contentDigest. The retained scan had complete coverage and a valid finding, but save again failed. Added a red/green registration regression, persisted the digest in that path, reran the full suite (470 pass, 6 skip, 0 fail), and force-updated this PR.

createsomethingtoday:
Follow-up hardening is now included in 6723b8b: the CLI regression no longer stops at asserting the registration response. It copies a real draft bundle, prepares completion, verifies the authoritative snapshot digest was persisted into the sealed manifest, and completes the scan through the workbench. Validation remains green: 470 passed, 6 platform skips, 0 failed; types and formatting also pass.

mldangelo-oai:
Thanks for looking closely at how diff scans are tied to a repository snapshot. The existing full Git commit IDs already identify the immutable content for these scans, while the additional digest would require buffering the complete binary diff. We don't have a separate use case that justifies that cost, so I'm going to close this PR. We'd welcome a focused follow-up if you find a concrete case the current identifiers do not cover.

Issue #43, opened 2026-07-29
Scan failure messages are stored without credential redaction


BryanTegomoh:
`cliErrorMessage()` at `src/cli.ts:2898` strips API keys, bearer tokens and URL passwords out of every error the CLI prints. It is used at 22 call sites, all of them on the way to stderr.

The same string gets written to the scan database without it.

`src/api.ts:733`:

```ts
"fail-scan",
"--scan-id",
activeScan.id,
"--message",
(failure instanceof Error
  ? failure.message
  : String(failure)
).slice(0, 2400),
```

That value comes off the model event stream (`api.ts:1191`, `api.ts:1196`). It is stored in `scans.failure_message` (`workbench_db.py:1777`) and printed back out by `scan-history show` (`scan-history-renderer.ts:213`). Nothing on that path redacts it.

So a credential the CLI hides from the terminal still lands on disk.

### What it looks like

I copied `cliErrorMessage` verbatim and ran four realistic upstream errors through it. First line is what you see, second is what gets stored.

```
stderr : error sending request for url (https://[redacted]@proxy.internal/v1/responses)
stored : error sending request for url (https://svc-account:s3cr3tP4ss@proxy.internal/v1/responses)

stderr : Incorrect API key provided: [redacted]
stored : Incorrect API key provided: sk-proj-ABCDEF1234567890

stderr : upstream rejected header Authorization: Bearer [redacted]
stored : upstream rejected header Authorization: Bearer eyJhbGciOiJIUzI1NiJ9padding

stderr : fetch to https://api.internal/v1/chat?api_key=[redacted] failed
stored : fetch to https://api.internal/v1/chat?api_key=SUPERSECRETVALUE123 failed
```

The first one is the easy path to hit. Point the tool at a proxied model provider whose base_url carries userinfo and let the provider be unreachable. `classifyConnectionFailure` (`api.ts:1543`) reads that as a network error, so you get the canned "The model service could not be reached" from `cli.ts:2499` and never see the URL. The raw string with the password is already in `workbench.sqlite3`.

### Why it leaves the machine

`Dockerfile:52` sets `CODEX_SECURITY_STATE_DIR=/output/.codex-security-state` and `compose.yaml:28` mounts `/output` to the host results directory. The database sits inside the bundle people archive and pass around.

The db file is `0600` (`workbench_db.py:237`), so I am not claiming local disclosure. The problem is that the value is stored unredacted, printed again by `scan-history show`, and carried along in the results directory.

### Fix

[...]

PR #119, opened 2026-07-30
fix: fail the scan on every turn.failed payload

### LINKED ISSUE #37: A failed turn is reported as a successful scan when turn.failed has an unexpected error payload
## Summary

A `turn.failed` event is only treated as a failure when its payload is an object with a string `message`. Any other payload shape falls through every branch of the event loop, and if a `turn.completed` was already observed the failed turn is reported to the caller as a **successful scan** and recorded in the workbench with `complete-scan`.

The branch requires both `isRecord(event["error"])` and a string `message`:

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/api.ts#L1187-L1192

There is no `else`, so `error: null`, `error` as a plain string, or a non-string `error.message` all leave `status` as `"completed"` and the loop continues to `onFinalize`.

- macOS 26.5.2 (build 25F84)
- Node.js v24.5.0
- Bun 1.3.14

## Steps to reproduce

No API call is required. Using the injected-dependency pattern from `tests-ts/api.test.ts`, drive the event stream with `thread.started`, `turn.completed`, then a `turn.failed` whose error payload varies. The scan directory holds a valid artifact bundle so the only variable is the failure event.

```ts
createCodex: () => ({
  startThread: () => ({
    id: null,
    async runStreamed() {
      async function* stream() {
        yield { type: "thread.started", thread_id: "t" };
        yield { type: "turn.completed", usage: {} };
        yield { type: "turn.failed", error: null };   // <- payload under test
      }
      return { events: stream() };
    },
  }),
}),
```

Observed results:

```text
[A] turn.failed with a string message (handled correctly)
    outcome:  rejected: CodexSecurityError: model refused the turn
    workbench: ["register-cli-scan","fail-scan"]

[B] turn.failed with error: null
    outcome:  RESOLVED (scan treated as successful)
    workbench: ["register-cli-scan","complete-scan"]

[B] turn.failed with error as a plain string
    outcome:  RESOLVED (scan treated as successful)
    workbench: ["register-cli-scan","complete-scan"]

[B] turn.failed with error.message not a string
    outcome:  RESOLVED (scan treated as successful)
    workbench: ["register-cli-scan","complete-scan"]

[B] turn.failed with error.code only
    outcome:  RESOLVED (scan treated as successful)
    workbench: ["register-cli-scan","complete-scan"]
```

The well-formed case behaves correctly, which isolates the payload shape as the only difference.

## Expected behavior

`turn.failed` should be terminal regardless of the shape of its error payload. When the message cannot be extracted, the scan should still fail, with a generic message rather than silent success.

## Actual behavior

`run()` resolves with a `ScanResult`, the CLI exits 0, and the workbench records `complete-scan`. A user running with `--fail-on-severity` in CI gets a passing build from a turn the model reported as failed.

## Impact

This is a silent failure rather than a visible one. The scan directory can hold a stale or partial artifact bundle from an earlier phase while the run is reported as a clean pass, so neither the exit code nor `scans list` shows anything wrong.

Whether this is reachable in practice depends on what the Codex SDK can emit for `turn.failed`; the SDK types it loosely enough that the SDK cannot rely on the shape, and the defensive `isRecord` check in this very branch suggests the same conclusion.

## Suggested direction

Make `turn.failed` terminal in all cases, extracting the message when possible:

```ts
} else if (event.type === "turn.failed") {
  const error = event["error"];
  const message =
    isRecord(error) && typeof error["message"] === "string"
      ? error["message"]
      : "Codex Security reported a failed turn without a usable error message.";
  throw new CodexSecurityError(message);
}
```

Haiduongcable:
## Summary

Fixes #37.

A `turn.failed` event was only treated as a failure when its payload was an object with a
string `message`. Every other payload shape fell through every branch of the event loop, and
because a `turn.completed` had already been observed the failed turn was returned to the
caller as a successful scan and recorded in the workbench with `complete-scan`. The `else if`
following that branch tests a different event type, so there was no fallback.

- Treat every `turn.failed` event as terminal, regardless of its error payload shape.
- Reuse a non-blank `error.message` verbatim so existing diagnostics and expectations are
  unchanged, and fall back to a generic message for every other shape.
- Deliberately do not forward or stringify any other payload shape: this message reaches
  `fail-scan --message` and is stored in `scans.failure_message` without redaction, so
  widening what is copied out of the payload would add a new credential-disclosure path into
  persistent scan history. Only `error.message` is reused because it is the single shape the
  previous code already surfaced, which keeps this change free of new data on that path.
- Cover the blank-message case, which now fails with the generic message instead of raising an
  error whose message is whitespace.
- Add regression coverage for seven payload shapes against the real event loop, each with a
  complete valid artifact bundle so the failed turn is the only variable.

Left for a separate change: the redaction gap itself, reported in #43. Once the failure-storage
path redacts, this helper can safely be widened to improve diagnostics for unusual payloads.

## Validation

- Reproduced the failure on `main` before applying the fix: the new tests fail with
  `Expected promise that rejects / Received promise that resolved`, which is the reported bug.
  17 passed and 2 failed before the source change; 19 passed after.
- Confirmed `null`, bare-string, non-string `message`, array, `undefined`, and blank-message
  payloads were each returned as successful scans and recorded with `complete-scan` before the
  fix, and are each terminal after it.
- Confirmed the new branch cannot shadow a later one: the only subsequent condition tests
  `event.type === "error"`, which is never true for a `turn.failed` event.
- Full public SDK suite with the CI-pinned Bun 1.3.14: 472 passed, 6 expected skips, 0 failed.
- TypeScript type check, generated-model check, Prettier, and whitespace checks.
- The new tests use `copyCompletedScan`, so the assertions cannot be satisfied incidentally by
  artifact validation.

Reporter credit: @mariohercules, who identified the fall-through and the affected branch in #37.

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -380,7 +380,9 @@ including cached input and cache writes; fees and surcharges are not included.
 
 Use `--max-cost USD` to stop a scan, including its delegated workers, when its
 running cost exceeds the limit. Partial results are preserved. Requests
-already in progress can finish above the limit.
+already in progress can finish above the limit. Cost tracking accepts Codex
+session events up to 1 MiB; an oversized event stops the scan because its
+running cost can no longer be verified safely.
 
 Run `npx @openai/codex-security scan --help` or `npx @openai/codex-security bulk-scan --help`
 for the complete CLI references.
--- a/sdk/typescript/_bundled_plugin/examples/completed-scan/findings.json
+++ b/sdk/typescript/_bundled_plugin/examples/completed-scan/findings.json
@@ -42,6 +42,12 @@
       "remediation": "Normalize destinations and reject entries that escape the extraction root.",
       "validation": null,
       "attackPath": null,
+      "remediationTests": [
+        "Assert that extracting an archive entry named `../escape.txt` fails without writing outside the extraction root."
+      ],
+      "preventiveControls": [
+        "Route all archive extraction through one helper that normalizes and validates destination paths."
+      ],
       "provenance": {
         "source": "local_plugin"
       },
--- a/sdk/typescript/_bundled_plugin/examples/completed-scan/scan-manifest.json
+++ b/sdk/typescript/_bundled_plugin/examples/completed-scan/scan-manifest.json
@@ -30,7 +30,7 @@
     "artifacts": [
       {
         "path": "findings.json",
-        "sha256": "db5ce8533c1b6cd9131f9e12e8bb705f63474caa2a3f7dd21207666b3a8da777",
+        "sha256": "a6dc4521d6478828224fbafc33585401a47bfeb0e56e07b5d2886e8a29937f2f",
         "mediaType": "application/json"
       },
       {
--- a/sdk/typescript/_bundled_plugin/references/finding-detail-fields.md
+++ b/sdk/typescript/_bundled_plugin/references/finding-detail-fields.md
@@ -23,7 +23,7 @@ The finding detail view is a decision-focused projection of the canonical findin
 - dataflow source, meaningful transformations, dangerous sink, and concrete outcome;
 - realistic attacker, entry point, access requirements, preconditions, and attacker outcome;
 - severity rationale plus the specific evidence that would raise or lower the rating;
-- the minimal remediation invariant, regression tests, and preventive controls.
+- the minimal remediation invariant (`remediation`, a single string), plus `remediationTests` and `preventiveControls`, each an array of short strings with one regression test or preventive control per entry.
 
 Keep background exposition, alternate exploit research, full PoC instructions, representative command output, and long source walkthroughs in the detailed write-up. Do not copy them into canonical fields merely to make the workspace report longer. The workspace should stay self-contained enough to support triage while avoiding duplicated or speculative prose.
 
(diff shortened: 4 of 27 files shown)


Haiduongcable:
@codex review this PR

Haiduongcable:
Rebased onto `9c7634b` now that #115 and #116 have merged. The diff is unchanged: `sdk/typescript/src/api.ts` and `sdk/typescript/tests-ts/api-events.test.ts`, +96/-6.

Full CI is green on this exact commit via a fork run:

- `ubuntu-latest` / node 22, 24, 24.0.0, 26, 26.0.0
- `macos-latest` / node 22
- `windows-latest` / node 22
- `linux-amd64` container

The local `node-ci` step sequence also passes: `audit:prod` (no known vulnerabilities), `types`, `test` (472 passed, 6 expected skips, 0 failed), `format`, and `build`.

Two notes for review:

- The fall-through is still present on `main` at `sdk/typescript/src/api.ts:1399-1403`.
- #115 touched only `sdk/typescript/src/cli.ts`, so it does not interact with this change.

mldangelo-oai:
Thanks for taking the time to look into this failure case. There is no longer a code change in the PR, and the behavior is already covered on `main`, so there is nothing left for us to review here. I'm going to close it, but we'd be happy to see another focused contribution if you spot a gap.

PR #120, opened 2026-07-30
fix: contain scan matching failures to a single batch

mldangelo:
## Summary

`codex-security scans match --all` aborts the entire campaign when one batch produces conflicting confirmed and uncertain findings, and it does so *after* earlier batches have already been written.

`matchAllScans` throws at `src/cli.ts:2006` while building `comparisons` inside `beforeScans.map(...)`. That is before the failing batch's own save loop, but after every earlier iteration of the outer `for (const ... of batches)` loop already called `save-scan-comparison`. The result is a nonzero exit with partially persisted match state and no resume path — rerunning re-derives the unsaved pairs but the operator has no way to tell how far the previous run got.

## What this changes

Each batch is now contained. A batch that cannot be matched is counted, warned about on stderr, and skipped; the remaining batches still save.

If **nothing** matched, the first failure is rethrown. This keeps a systemic problem — unwritable workbench, rejected credential — from being downgraded to a warning on an apparently successful no-op, and it preserves today's behavior exactly in the single-batch case.

The summary gains `unmatchedBatches` when nonzero, rendered in the TTY match-all view as `N scans could not be matched; rerun to retry` alongside the existing `scans unavailable` line.

## What this deliberately does *not* change

The conflict check itself. Refusing to persist conflicting confirmed and uncertain findings is intentional and test-covered by `tests-ts/cli-workbench.test.ts` → *"does not save conflicting confirmed and uncertain matches"*. **That test passes unmodified** — that was the acceptance signal for this change, and I'd treat any need to edit it as a sign the fix had overreached.

## Open question for maintainers (not addressed here)

There is a real disagreement between two layers that this PR leaves alone:

`src/scan-comparison.ts:246-257` with `allowHistoricalUncertainty: true` **deliberately permits** an `uncertain` pair whose `afterOccurrenceId` is already claimed by a confirmed match — it only ever rejects the before-side collision via `matchedBefore`. The flag exists so an after-occurrence can be confirmed against one earlier scan and uncertain against another. But `validateComparison` has no notion of scan boundaries, so it also admits the collision **within a single before-scan**, which `matchAllScans` then treats as unusable.

That shape is not self-contradictory — *"after-finding A definitely corresponds to X, and may also correspond to Y"* is a coherent claim about two earlier findings merged into one — so a well-behaved model can produce it and get a skipped batch. Either `scan-comparison` should reject it or `matchAllScans` should drop the pair. I did not pick, because inverting a documented product invariant isn't mine to decide. Happy to follow up either way.

## Testing / QA instructions

Full gate, from `sdk/typescript`. Run pnpm from inside this directory, not the repo root — from the root, `pnpm --dir` resolves your global pnpm instead of the pinned `packageManager` and the next in-directory command tries to purge `node_modules`:

```bash
cd sdk/typescript
CI=true pnpm run types
CI=true pnpm run test
CI=true pnpm run format
CI=true pnpm run build
```

### The assertion that matters most — pre-existing behavior is intact

```bash
CI=true bun test --timeout 30000 ./tests-ts/cli-workbench.test.ts -t "does not save conflicting"
```
Expect `1 pass / 0 fail`. This file is unmodified apart from added tests and one added import; if this fails, the containment scope is wrong.

### New coverage

```bash
CI=true bun test --timeout 30000 ./tests-ts/cli-workbench.test.ts -t "after one batch conflicts"
CI=true bun test --timeout 30000 ./tests-ts/cli-workbench.test.ts -t "no batch could be matched"
```

1. **`keeps matching later scans after one batch conflicts`** — two batches, the first conflicting. Asserts exactly one `save-scan-comparison` call, that it is for the *healthy* batch (`before-ok`), that stderr names the skipped scan (`after-conflict`), and that the JSON summary reports `matchedPairs: 1, unmatchedBatches: 1`. On `main` this exits 2 and saves nothing.
2. **`reports the underlying failure when no batch could be matched`** — every `save-scan-comparison` throws a workbench-database error. Asserts exit 2 and that the underlying message reaches stderr rather than being swallowed into a zero-exit summary.

### Manual check of the rendered line

The `unmatchedBatches` line only renders on a TTY with `--format toon` (default). To see it without a real workbench, call the renderer directly:

```bash
cd sdk/typescript && bun -e '
console.log(renderScanHistory(
  { repository: "/repo", scanCount: 3, matchedPairs: 1, findingMatches: 1, unavailableScans: 0, unmatchedBatches: 1 },
  "match-all", { color: false }));'
```
Expect a `1 scans could not be matched; rerun to retry` line. Confirm `--format json` output is unaffected (the field is simply absent when zero).

### Regression surface to watch

- `--format json` consumers now see an extra optional key. It is omitted entirely when zero, so existing `toEqual` assertions on clean runs are unaffected (verified: the pre-existing `matches every unmatched pair` test asserts the full object with `toEqual` and still passes).
- Partial failure *within* a batch is unchanged: if save #2 of 3 in one batch fails, saves #1 persisted. That pre-existing behavior is out of scope here.


mldangelo-oai:
@codex review Please review exact head 87781f3bdfc8250f4a948287f98447d36714ae7f, focusing on isolating scan-matching failures to one batch, preserving unrelated comparisons, renderer behavior, and the current-main merge.

mldangelo-oai:
@codex review exact head c1f208b45f683d5797fee93723e9764f6b4940d4

mldangelo-oai:
@codex review

kmbroai (commented):
## Critical review

Reviewed `96b8110bcb3b758bbabb65c43c06268113074c54`. **Recommendation: fix the two failure-path issues before landing.** Batch-local recovery is useful, but the catch currently both fails to recover and has the wrong persistence boundary.

### Correctness

1. **[P1] The warning path calls a removed helper.** At [cli.ts:2261](https://github.com/openai/codex-security/blob/96b8110bcb3b758bbabb65c43c06268113074c54/sdk/typescript/src/cli.ts#L2261), `cliErrorMessage` is undefined; this head imports `redactedErrorMessage` instead. Any batch error throws a new `ReferenceError` from inside the catch, so later healthy batches never run and the original error is lost. I reproduced this with the checked-in `cli-workbench.test.ts`: **10 passed, 3 failed**, including both new recovery tests. Use the existing redaction helper, not another formatter. The earlier review comment remains applicable at this exact head.

2. **[P1] Do not treat database-write failures as recoverable matching failures.** The same `try` contains [the persistence loop](https://github.com/openai/codex-security/blob/96b8110bcb3b758bbabb65c43c06268113074c54/sdk/typescript/src/cli.ts#L2240). Once any pair saves, `matchedPairs > 0` disables the final rethrow. A later `save-scan-comparison` failure therefore produces a normal return even though persistence is incomplete. In an isolated control of the actual function, with only the intended error formatter supplied to get past issue 1, the first save succeeded, the second threw a synthetic database-write error, and the result was `{matchedPairs: 1, unmatchedBatches: 1}` rather than a rejection. This is distinct from a conflicting model-generated match and also remains unresolved from the earlier review.

Keep recovery for an individual matcher/validation failure. Narrow the `try/catch` to producing and validating that batch's comparisons; on that failure, warn and `continue`. Save the validated comparisons **outside** the catch so persistence failures propagate normally. That is a simpler boundary than inferring error severity from whether an earlier save happened. Previously saved rows already survive an exception; the useful feature here is continuing to later independent batches, not preventing those rows from being discarded.

Keep `unmatchedBatches` as an additive diagnostic. Add one regression for a save failure **after a successful save**, which the all-failed test cannot cover.

### Verification

Ran `bun test --timeout 30000 tests-ts/cli-workbench.test.ts` on this head with Bun 1.3.14/Linux and locally available dependencies, plus the isolated persistence control described above. This was not a clean dependency-install or full platform-matrix run. No source fixes were applied to the reviewed head.

PR #121, opened 2026-07-30
fix: keep scan history rendering total over workbench payloads

mldangelo:
## Summary

`runWorkbench` (`src/runtime.ts`) validates only that the workbench response parsed as JSON and `isRecord(result)`. Nothing checks the shape between the Python workbench and `renderScanHistory`, and the renderer then cast every field unconditionally. Any drift — a plugin/database version skew, a partially migrated sqlite row, a hand-edited `workbench.sqlite3` — surfaced as an unhandled `TypeError` with a stack trace rather than a usable view or an actionable error.

Three crashes confirmed by calling the renderer directly:

| Payload | Before |
|---|---|
| finding with `severity` absent | `TypeError: undefined is not an object (evaluating 'severity.level')` |
| finding with `severity: null` | same — `typeof null === "object"` |
| `scans` row without `progress` | `TypeError: undefined is not an object (evaluating 'scan.progress.status')` |
| `knownSince: "not-a-timestamp"` | `RangeError: Invalid time value` from `Intl.DateTimeFormat.format` |

## Scope — please read before reviewing as a severity call

**These are not reachable with the plugin as currently bundled.** I checked `_bundled_plugin/scripts/workbench_db.py`: `severity` is always emitted as `{level: ...}` and `progress` is always populated, and every timestamp comes from `now()` = `datetime.now(timezone.utc).isoformat()`, so `knownSince` is always ISO-8601. The renderer also only runs on a TTY with `--format toon`, so `--format json` was never affected.

So this is drift resilience, not a live bug. I'm raising it because the *class* is one bad row away, and a read-only history view crashing is a poor failure mode for a command whose whole job is to tell you what happened.

## Approach: make the renderer total, don't validate at the boundary

I considered adding a schema check in `runWorkbench` instead. I did not, for two reasons: the workbench payload has a large optional surface, so a schema would carry real risk of rejecting *valid* data and breaking working installs; and a display layer's correct response to a field it cannot read is to omit it, not to refuse the whole view.

So every read goes through three small accessors — `record()`, `records()`, `text()` — and the behavior is:

- missing/null severity → blank badge, default color
- missing `progress.status` → `UNKNOWN`
- unparsable `knownSince` → the raw value, bounded to 32 chars (information preserved rather than dropped)
- unrecognized severity → sorts **below** the known levels. Previously `Object.keys(SEVERITY_COLORS).indexOf(...)` returned `-1`, which sorted an unknown severity **above CRITICAL** — a separate real ordering bug fixed here via `severityRank()`.

## Testing / QA instructions

Baseline before this branch: **470 pass / 6 skip / 0 fail**. After: **475 pass / 6 skip / 0 fail** (five added tests).

```bash
CI=true pnpm run types
CI=true pnpm run test
CI=true pnpm run format
CI=true pnpm run build
```

### The regression guard that matters most

```bash
CI=true bun test --timeout 30000 ./tests-ts/scan-history-renderer.test.ts
```
Expect `11 pass / 0 fail`. The six pre-existing tests in this file assert **exact rendered text** (via `stripVTControlCharacters`) across compare/list/show/match-all, narrow and wide widths, and dim-styling rules. They are unmodified and still pass, which is the proof that output for well-formed payloads did not change. If any of those six fail, the accessors changed real behavior and the diff is wrong.

### New coverage — `scan history renderer resilience`

1. `renders comparison findings with a missing or null severity` — both shapes; asserts the title and `src/search.ts:10` still render.
2. `sorts an unrecognized severity below every known severity` — asserts CRITICAL appears before the unknown-severity finding in the output. Fails on `main` (unknown sorts first).
3. `lists scans that carry no progress record` — asserts the scan id renders and status shows `UNKNOWN`.
4. `falls back to the raw value for an unparsable knownSince` — asserts `Known since not-a-timestamp` rather than a `RangeError`.
5. `renders every command from an empty payload` — `{}` through all four commands, asserting none throw. This is the broad net for the casts I did not individually reproduce (`result["coverage"]`, `result["findings"]`, `result["severityCounts"]`, `basename(result["targetPath"])`, `basename(result["repository"])`).

### Verifying the crashes exist on main

```bash
git stash && cd sdk/typescript && bun -e '
try { renderScanHistory({ beforeScanId:"b", afterScanId:"a", coverage:{afterCompleteness:"complete"},
  summary:{resolved:1}, findings:[{ status:"resolved", title:"x", locations:[{path:"a.ts"}] }] }, "compare"); }
catch (e) { console.log("CRASH:", e.constructor.name, e.message); }'
# then: git stash pop
```

### Note for maintainers

This file is also touched by PR "fix: align the severity badge column in finding tables". The two diffs are in different regions (accessors + helper functions here; badge padding and its dependent indents there) but both are small, so whichever lands second may need a trivial rebase.


mldangelo-oai:
@codex review Please review exact head 0282909557b7e99eac4cd65019e3d05d09050bc1, focusing on total scan-history rendering over malformed optional workbench values, safe presentation, and the current-main merge.

mldangelo-oai:
@codex review exact head 0dff0e5ded15ca0ddaa0c1152abf37c8edfcfe09

mldangelo-oai:
@codex review

kmbroai (commented):
## Critical review

Reviewed `6fa53e5932d36fe0812bf0190fc83f891ce889ca`. **Recommendation: retain the targeted crash fixes, but tighten the boundary between optional presentation fields and unreadable result data.** The current implementation does not yet meet its stated total-rendering contract.

### Correctness and remaining scope gaps

1. **[P2] Optional values still reach operations that throw.** The [knowledge-base cast](https://github.com/openai/codex-security/blob/6fa53e5932d36fe0812bf0190fc83f891ce889ca/sdk/typescript/src/scan-history-renderer.ts#L314) accepts a nonempty string through the `length` check and then calls `.map()`. I reproduced `renderScanHistory({recipe: {knowledgeBasePaths: "policy.md"}}, "show", {color: false})` throwing `TypeError`. The [artifact path code](https://github.com/openai/codex-security/blob/6fa53e5932d36fe0812bf0190fc83f891ce889ca/sdk/typescript/src/scan-history-renderer.ts#L323) similarly passes a truthy non-string `scanDir` to `relative()`; `{scanDir: 1, artifacts: {report: "/scan/report.md"}}` throws `ERR_INVALID_ARG_TYPE`. These are residual gaps in the behavior this PR promises, not newly discovered regressions in the original renderer. Guard the string array and path at their consumption sites, and include these actual shapes in the tests. The existing knowledge-base review comment is still applicable.

2. **[P2] Invalid findings should not silently disappear into a successful-looking view.** [Coercing `findings` through `records()`](https://github.com/openai/codex-security/blob/6fa53e5932d36fe0812bf0190fc83f891ce889ca/sdk/typescript/src/scan-history-renderer.ts#L335) makes an invalid container indistinguishable from an empty array. A payload with `progress.status: "complete"`, `findingCount: 3`, and `findings: "malformed"` renders a normal `COMPLETE` header with no findings section or diagnostic. This fallback is newly introduced. Preserve the rest of the display if useful, but visibly say that finding details could not be read; do not present malformed core data as an ordinary empty result. The same distinction matters when filtering invalid entries from an otherwise nonempty array.

Missing severity and invalid timestamps justify small presentation guards. A blanket promise to accept every arbitrary object is substantially broader: it adds repeated coercion without preserving the distinction between absent optional metadata and a broken response. Prefer a small, explicit normalization boundary and direct rendering afterward, or keep the patch focused on the demonstrated optional-field crashes. Do not add a generic recursive validation framework.

The new fallback-aware scalar accessor is also used inconsistently: an empty `show` currently prints literal `undefined` for the ID and mode. Use explicit fallbacks for optional labels. The unknown-severity sort adjustment is reasonable, but independent of crash containment.

### Verification

The checked-in renderer suite passes **11/11** on Bun 1.3.14/Linux. The additional synthetic calls above reproduce both remaining exceptions and the silent-loss behavior. No native-platform behavior or full package validation is claimed.

PR #122, opened 2026-07-30
feat(cli): align severity badges in finding tables

mldangelo:
## Summary

`src/scan-history-renderer.ts` pads the severity badge with `severity.padEnd(8)`. That is a no-op for `INFORMATIONAL`, which is thirteen characters, so informational findings push their title five columns right of every other severity:

```
    HIGH      Short severity row
    INFORMATIONAL  Long severity row
```

`INFORMATIONAL` is a normal value, not an edge case — it is the longest key in `SEVERITY_COLORS` and it appears in `DISPLAY_SEVERITIES` in `src/cli.ts`. Any scan that reports an informational finding renders a broken table in `scans compare` and `scans show`.

The underlying problem is that the badge width and the two indents that must line up beneath it were three independent literals — `padEnd(8)`, `wrap(title, 14, ...)`, and a hardcoded 14-space path indent — plus two more at +2/+4 for the nested `LINKED FINDINGS` block. They agreed only for as long as every label happened to fit in eight columns.

## What this changes

All five are now derived from one place:

```ts
const SEVERITY_LABELS: Record<string, string> = { INFORMATIONAL: "INFO" };
const SEVERITY_BADGE_WIDTH = Math.max(
  ...Object.keys(SEVERITY_COLORS).map((s) => (SEVERITY_LABELS[s] ?? s).length),
);
const FINDING_INDENT = 4 + SEVERITY_BADGE_WIDTH + 2;
```

**Both values are unchanged: width 8, indent 14.** So the only difference in rendered output is the label itself, `INFORMATIONAL` → `INFO`. Adding a longer severity in future automatically widens the column and moves every dependent indent with it, rather than silently breaking alignment again.

## The one judgment call — please confirm

I chose to abbreviate rather than widen the column to 13. Widening is the alternative and it keeps the full word, but it costs every finding title five characters: at the 48-column minimum, titles drop from 32 usable characters to 27, and finding titles are routinely 40–60 characters, so they would wrap considerably more. `INFO` is unambiguous and the badge is a compact status column, so the trade seemed clearly worth it — but it *is* a user-visible label change, so say the word and I will switch to widening instead.

This affects only the TTY history view. SARIF/CSV/JSON exports go through a different path and are untouched.

Known remaining limitation: an *unrecognized* severity longer than eight characters still misaligns. I deliberately did not truncate arbitrary values, since mangling a future legitimate severity name seemed worse than misaligning an anomalous one. (A related ordering bug for unknown severities — they sorted above CRITICAL because `indexOf` returns `-1` — is fixed in the companion PR below.)

## Testing / QA instructions

```bash
CI=true pnpm run types
CI=true pnpm run test
CI=true pnpm run format
CI=true pnpm run build
```

### Confirm the new tests actually catch the bug

The two added tests must fail against `main`'s renderer. Verify by reverting only the source and keeping the tests:

```bash
cd sdk/typescript
cp src/scan-history-renderer.ts /tmp/fixed.ts
git checkout main -- src/scan-history-renderer.ts
CI=true bun test --timeout 30000 ./tests-ts/scan-history-renderer.test.ts   # expect 6 pass / 2 fail
cp /tmp/fixed.ts src/scan-history-renderer.ts
CI=true bun test --timeout 30000 ./tests-ts/scan-history-renderer.test.ts   # expect 8 pass / 0 fail
```

Note that plain `git stash` does **not** work for this check — the tests live in the same file, so stashing removes them too and you get a misleading `6 pass / 0 fail`.

### New coverage — `severity badge column`

1. **`starts every finding title at the same column`** — renders all five severities and asserts every title begins at the same offset, across `color: false`/`true` and widths 48/96/120. Also asserts the path line beneath each badge lines up with the title above it, which is the indent coupling.
2. **`keeps the badge column as wide as its widest label`** — asserts the exact string `    INFO      Informational finding` and that `INFORMATIONAL  ` no longer appears.

### Existing tests are unmodified

All six pre-existing tests in this file pass untouched. They assert relative *ordering* via `text.indexOf(...)` rather than badge text, which is why the label change does not churn them. Their exact-text assertions on other rows are the guard that indent 14 really did stay 14.

### Manual check

```bash
cd sdk/typescript && bun -e '
console.log(renderScanHistory({ beforeScanId:"b", afterScanId:"a",
  coverage:{afterCompleteness:"complete"}, summary:{new:2}, findings:[
    {severity:"high",title:"Short severity row",status:"new",locations:[{path:"a.ts",startLine:1}]},
    {severity:"informational",title:"Long severity row",status:"new",locations:[{path:"b.ts",startLine:2}]},
  ]}, "compare", {color:false}));'
```

### Note for maintainers

This file is also touched by PR "fix: keep scan history rendering total over workbench payloads". Different regions (badge padding and indents here; field accessors there), but whichever lands second may need a trivial rebase.


mldangelo-oai:
@codex review exact head fc7123dfc4ec2f7a286c0ad4f306b0ff8735ec39

mldangelo-oai:
@codex review

kmbroai (commented):
## Critical review

Reviewed `1723a3bdb0b3cccebd3f216869d8832d1f43bcaa`. **No blocking correctness issue found in this diff. This is useful display polish, not a necessary scan feature.**

### Necessity

The existing `padEnd(8)` does not truncate `INFORMATIONAL`, so that row starts its title later than the other severities while continuation and path lines still assume a 14-character indent. Abbreviating the display label to `INFO` fixes a real alignment inconsistency without changing the underlying severity. I checked current main (`227fc32ac47c9ac5d78f6dd246d8899673a67858`); the longer label is still rendered there, so this is not already redundant.

This can be prioritized below ingestion, matching, and scan-correctness fixes. It should not become a new formatting option or severity schema change.

### Implementation and simplification

The [shared badge width and finding indent](https://github.com/openai/codex-security/blob/1723a3bdb0b3cccebd3f216869d8832d1f43bcaa/sdk/typescript/src/scan-history-renderer.ts#L33) are a sensible small abstraction: the same geometry is reused for titles, paths, linked findings, and reasons. I would keep that relationship explicit instead of restoring separate magic indentation values. Computing the width from the five known labels is harmless; no generic terminal-table framework is needed.

One non-blocking test improvement: the new width/color matrix uses short titles that fit without wrapping. It demonstrates starting-column alignment, but does not exercise the continuation indentation changed by the patch. Replace some redundant short-title cases with **one long title that wraps at 48 columns and one linked-finding reason**; assert those continuation columns align after stripping ANSI codes. This gives more relevant coverage without growing the matrix.

The renderer-resilience work in #121 touches this same function. Whichever change lands second should preserve both the label geometry and the optional-field handling; neither requires merging the two concerns into a larger redesign.

### Verification

Ran `bun test --timeout 30000 tests-ts/scan-history-renderer.test.ts` at this head: **8 passed, 0 failed**, Bun 1.3.14/Linux. The test covers all five known severities with and without ANSI color. This is not a full package/platform-matrix result.

PR #117, opened 2026-07-30
fix: bound knowledge-base discovery and document extraction

mldangelo-oai:
## Summary

- Adopt the original knowledge-base hardening contribution from PR #95 while preserving the original substantive commit author, kartikshukla17, and acknowledging GautamSharma99 for submitting and maintaining the source PR.
- Bound discovery depth, entries, document count, per-document and aggregate input, extracted text, PDF pages, and cancellation.
- Preserve secure source handling, private temporary outputs, and existing public SDK behavior.

## Verification

- Full current-main SDK suite: 474 passed, six expected platform/integration skips, zero failures.
- Focused knowledge-base and orchestration coverage: 78 passed.
- TypeScript and formatting checks passed.

PR #123, opened 2026-07-30
fix: keep unusable DOCX numeric references as literal text

### LINKED ISSUE #40: Out-of-range numeric XML entity in a DOCX aborts the whole knowledge base and the scan
## Summary

`decodeXml` passes the parsed value of a numeric XML entity straight to `String.fromCodePoint` with no magnitude bound. Any entity above `U+10FFFF` raises `RangeError`, which propagates out of `extractDocx` and fails the whole `prepareKnowledgeBase` call, aborting the scan before it starts.

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/knowledge-base.ts#L206-L224

The regex admits `#\d+` and `#x[\da-f]+` with no upper limit, so `&#x110000;` and `&#99999999999999;` both reach `String.fromCodePoint` and throw.

- macOS 26.5.2 (build 25F84)
- Node.js v24.5.0
- Bun 1.3.14

## Steps to reproduce

Build a minimal `.docx` (a zip containing `word/document.xml`) whose body text contains an out-of-range numeric entity, then pass it to `prepareKnowledgeBase`:

```ts
const xml =
  '<?xml version="1.0" encoding="UTF-8"?>' +
  '<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main">' +
  "<w:body><w:p><w:r><w:t>Boundary &#x110000; case.</w:t></w:r></w:p></w:body>" +
  "</w:document>";
await writeFile(path, zipSync({ "word/document.xml": strToU8(xml) }));
await prepareKnowledgeBase([path]);
```

Observed results across four documents that differ only in body text:

```text
[ordinary text] prepared successfully

[valid entity &#65;] prepared successfully

[out-of-range &#x110000;] FAILED
   message: Cannot extract text from knowledge base DOCX: /.../threat-model.docx
   cause:   RangeError: Arguments contain a value that is out of range of code points

[huge decimal &#99999999999999;] FAILED
   message: Cannot extract text from knowledge base DOCX: /.../threat-model.docx
   cause:   RangeError: Arguments contain a value that is out of range of code points
```

## Expected behavior

An entity that cannot represent a code point should be left as literal text (or dropped), the same way an unrecognized named entity is already returned unchanged by the `entities[name.toLowerCase()] ?? entity` fallback on the line above.

## Actual behavior

The whole knowledge base fails to prepare, so `--knowledge-base` aborts the scan before Codex starts. The reported message — "Cannot extract text from knowledge base DOCX" — suggests the file is unreadable or corrupt, which sends the user looking in the wrong place; the actual `RangeError` is only visible via the error's `cause`.

## Impact

Modest, and I want to be straightforward about it: Word itself does not emit out-of-range entities, so this is unlikely to fire on a document authored normally. It is reachable through documents produced by conversion tools or hand-edited XML, and by any deliberately malformed file. The consequence is disproportionate to the cause — a single stray entity anywhere in one document blocks the entire scan, including all the other knowledge base files that were fine.

Note that this failure mode is asymmetric with the surrounding code, which is otherwise careful: unsupported file types, symlinks and unreadable entries are all rejected with targeted errors rather than propagating a runtime exception.

## Suggested direction

Clamp to the valid range and fall back to the literal entity text:

```ts
const codePoint = Number.parseInt(name.slice(hexadecimal ? 2 : 1), hexadecimal ? 16 : 10);
return Number.isFinite(codePoint) && codePoint <= 0x10ffff
  ? String.fromCodePoint(codePoint)
  : entity;
```

mldangelo:
Fixes #40.

## Summary

`decodeXml` in `src/knowledge-base.ts` passed the parsed value of a numeric character reference straight to `String.fromCodePoint` with no magnitude bound. The regex admits `#\d+` and `#x[\da-f]+` with no upper limit, so `&#x110000;` and `&#99999999999999;` both reached it and threw `RangeError`.

That propagated out of `extractDocx` as `Cannot extract text from knowledge base DOCX: <path>` — which reads like a corrupt or unreadable file and sends you looking in the wrong place, with the real `RangeError` visible only via the error's `cause` — and failed the entire `prepareKnowledgeBase` call. One stray reference in one document aborted the scan before Codex started, taking every other knowledge-base file with it.

## What this changes

A reference that cannot name a Unicode scalar value is left as literal text, exactly matching the `entities[name.toLowerCase()] ?? entity` fallback for unrecognized *named* entities on the line above. This is the direction suggested in the issue.

I also excluded surrogates (`U+D800`–`U+DFFF`), which the issue's suggested snippet does not cover. They do **not** throw — `String.fromCodePoint(0xD800)` happily returns a lone surrogate — so this is a separate, quieter bug: XML forbids surrogates in the `Char` production, and the lone surrogate would then be written out by `writeFile(..., "utf8")` as `U+FFFD`, silently corrupting the extracted text rather than failing. Same root cause, same one-line guard, so it seemed wrong to leave it.

```ts
function isUnicodeScalarValue(codePoint: number): boolean {
  return (
    Number.isInteger(codePoint) &&
    codePoint >= 0 &&
    codePoint <= MAX_CODE_POINT &&
    !(codePoint >= 0xd800 && codePoint <= 0xdfff)
  );
}
```

`Number.isInteger` covers the `NaN` case; huge decimals like `&#99999999999999999999;` parse to `1e20`, which *is* an integer by float semantics but is caught by the upper bound.

## Deliberately out of scope

**The broader policy question.** The issue's Impact section objects that "a single stray entity anywhere in one document blocks the entire scan." With the decode fixed, that no longer happens for *this* cause — the document extracts normally. But `prepareKnowledgeBase` still aborts wholesale for a genuinely unextractable document, and I did not change that, because failing loudly may well be the right behavior for a security knowledge base: silently proceeding with a document missing is arguably worse than refusing to start. If you want skip-and-warn semantics instead, that is a deliberate product change and I am happy to do it separately.

**The full XML `Char` production.** `&#0;` and other C0 controls still decode, so a NUL can reach the extracted text. XML forbids those too (`Char ::= #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]`), and implementing the production in full would be a few more lines. I left it out because it is neither a crash nor a silent corruption, and it is not what the issue reports. Flagging it as a possible follow-up.

## Testing / QA instructions

```bash
CI=true pnpm run types
CI=true pnpm run test
CI=true pnpm run format
CI=true pnpm run build
```

### Confirm the tests reproduce the issue

Both added tests must fail against `main`'s source. Revert only the source, keeping the tests (plain `git stash` will not work — the tests are in a different file, but reverting both defeats the check):

```bash
cd sdk/typescript
cp src/knowledge-base.ts /tmp/fixed.ts
git checkout main -- src/knowledge-base.ts
CI=true bun test --timeout 30000 ./tests-ts/knowledge-base.test.ts
#   expect 7 pass / 2 fail, with:
#   RangeError: Arguments contain a value that is out of range of code points
cp /tmp/fixed.ts src/knowledge-base.ts
CI=true bun test --timeout 30000 ./tests-ts/knowledge-base.test.ts   # expect 9 pass / 0 fail
```

### New coverage

Both tests build a real `.docx` through the existing `docx()` helper (`fflate` `zipSync` of `word/document.xml`) and assert on the extracted text, so they exercise the whole `prepareKnowledgeBase` → `extractDocx` → `decodeXml` path rather than the decoder in isolation.

1. **`keeps DOCX numeric references that cannot name a code point as literal text`** — seven cases, covering both the regression and the no-regression side:

   | Input | Expected |
   |---|---|
   | `&#x110000;` | left literal |
   | `&#1114112;` | left literal |
   | `&#99999999999999;` | left literal |
   | `&#xD800;` (lone surrogate) | left literal |
   | `&#65;` | decodes to `A` |
   | `&#128512;` | decodes to 😀 |
   | `&#x10FFFF;` | decodes (boundary must still work) |

   The last three matter as much as the first four — the bound must not clip valid astral characters or the max code point.

2. **`keeps one unusable reference from failing the other knowledge-base documents`** — a directory holding one good `.md` and one `.docx` with an out-of-range reference. Asserts **both** documents are extracted. This is the user-visible impact from the issue: on `main` the `.md` is lost too.

### Manual reproduction from the issue

```bash
cd sdk/typescript && bun -e '
import { zipSync, strToU8 } from "fflate";
import { writeFile, mkdtemp } from "node:fs/promises";
import { tmpdir } from "node:os"; import { join } from "node:path";
import { prepareKnowledgeBase } from "./src/knowledge-base.js";
const dir = await mkdtemp(join(tmpdir(), "kb-"));
const xml = `<?xml version="1.0"?><w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"><w:body><w:p><w:r><w:t>Boundary &#x110000; case.</w:t></w:r></w:p></w:body></w:document>`;
await writeFile(join(dir, "threat-model.docx"), zipSync({ "word/document.xml": strToU8(xml) }));
const kb = await prepareKnowledgeBase([join(dir, "threat-model.docx")]);
console.log("prepared:", kb.path); await kb.cleanup();'
```
Expect `prepared: ...` on this branch; on `main` it throws `Cannot extract text from knowledge base DOCX`.

### Note for maintainers

PRs #117 and #95 are competing rewrites of this same file (both add `MAX_DOCUMENTS`/`MAX_DIRECTORY_DEPTH`/size caps). Neither fixes this — I checked, they only rename the `decodeXml(...)` call site to `const text = decodeXml(...)`. This PR touches only `decodeXml` and one new constant, so it should rebase cleanly onto whichever of those you land, but merge order matters.


mldangelo-oai:
@codex review Please review exact head be2157343c07fa0d7219e6c72a0e93c9531bd3cb, focusing on DOCX numeric entity bounds, Unicode scalar validation, invalid references, preserving unrelated documents, and the current-main merge.

mldangelo-oai:
@codex review exact head 65f8b1273224cf305f0eea8e3ddf70314116ecba

mldangelo-oai:
@codex review

kmbroai (commented):
## Critical review

Reviewed `4090aa5ac766b4b888335045a35c7be2b6a2715f`. **The change is justified and appropriately small; no blocking correctness issue found.**

### Necessity and correctness

An out-of-range numeric reference reaches `String.fromCodePoint` in the existing decoder and aborts preparation of the entire knowledge base, not just the affected text. Current main (`227fc32ac47c9ac5d78f6dd246d8899673a67858`) still has the unguarded conversion. This is a concrete correctness fix rather than speculative input hardening.

The [scalar-value guard](https://github.com/openai/codex-security/blob/4090aa5ac766b4b888335045a35c7be2b6a2715f/sdk/typescript/src/knowledge-base.ts#L237) covers the exception-producing upper bound and the separate lone-surrogate corruption case. It still accepts supplementary characters and the maximum scalar value. Retaining the original reference when it cannot be decoded is consistent with the decoder's existing literal fallback and avoids silently deleting input text.

### Simplification and scope

I would keep this as a local decoder fix. Do **not** add document-count/size caps, a general XML parser, or skip-and-warn behavior for genuinely unreadable documents to solve this issue. Those would change unrelated contracts. Failing to load a requested knowledge-base document and recovering from one undecodable reference are different decisions.

The named predicate and bound are already simple enough; inlining them or removing the integer check offers little practical benefit. The long explanatory comment could be shortened to the literal-fallback and surrogate rationale, but that is non-blocking and not a reason to delay the fix.

The acknowledged XML-control-character question is a separate follow-up: being a Unicode scalar does not mean being legal in XML. This PR should continue to describe its guarantee as safe scalar conversion, not full XML validity. I would not expand that policy here without an observable requirement.

### Verification

Ran `bun test --timeout 30000 tests-ts/knowledge-base.test.ts` at this head: **9 passed, 0 failed**, Bun 1.3.14/Linux using locally available dependencies. This exercises real synthetic DOCX archives, invalid-reference preservation, valid astral/max-boundary decoding, and retaining the other document in a mixed knowledge base. No clean-install or full platform-matrix validation is claimed.

PR #124, opened 2026-07-30
docs: correct runtime and scan guidance

GautamSharma99:
## Summary

  Update the TypeScript package documentation so its runtime requirements, defaults, filesystem paths, and scan behavior match the current
  implementation.

  ## Changes

  - Corrected the documented `cli_auth_credentials_store` default from `"file"` to `"auto"`.
  - Clarified which operations require Python:
    - scans;
    - bulk scans;
    - exports;
    - scan-history commands;
    - saved-finding commands.
  - Documented how Python interpreter selection differs across interfaces:
    - `--python` for scan, bulk-scan, and export commands;
    - `pythonPath` for the SDK;
    - `PYTHON` for all Python-backed commands.
  - Documented that working-tree snapshots retain files from untracked nested Git repositories.
  - Clarified that initialized submodules must have clean worktrees because their recorded commits define the parent repository snapshot.
  - Added the default bulk-scan worker count and clarified that `--max-attempts` controls attempts for pending repositories during each command
  invocation.
  - Documented the exact workbench database location and precedence:
    - `$CODEX_SECURITY_STATE_DIR/workbench.sqlite3` when the override is configured;
    - `$CODEX_HOME/state/plugins/codex-security/workbench.sqlite3` otherwise;
    - `~/.codex` when `CODEX_HOME` is unset.
  - Added a regression assertion ensuring the credential-store value shown in the README stays synchronized with `DEFAULT_CODEX_CONFIG`.

  ## Verification

  - `pnpm run format`
  - `pnpm run types`
  - `pnpm run build`
  - Focused configuration, runtime, CLI, and scan-recovery tests:
    - 168 passed
    - 1 expected Windows skip
    - 0 failed
  - Final CLI suite:
    - 81 passed
    - 0 failed

  The full suite encountered an unrelated timeout in the authentication orchestration test `keeps ambient credentials available after successful
  ChatGPT login`. Package smoke testing reached dependency installation but was blocked because the configured npm registry could not resolve
  the pinned `fast-uri@3.1.4` package.

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -16,10 +16,11 @@ npx @openai/codex-security --version
 ```
 
 The package supports macOS, Linux, and Windows and requires Node.js 22.13.0 or
-later in the 22.x release line, Node.js 24.x, or Node.js 26.x. Scanning and
-exporting findings also require Python 3.10 or later. If you use Python 3.10,
-install the `tomli` package. Select another interpreter with `--python`,
-`pythonPath`, or `PYTHON` when needed.
+later in the 22.x release line, Node.js 24.x, or Node.js 26.x. Scans, bulk
+scans, exports, scan-history commands, and saved-finding commands also require
+Python 3.10 or later. If you use Python 3.10, install the `tomli` package.
+Scan, bulk-scan, and export commands accept `--python`; the SDK accepts
+`pythonPath`; and every Python-backed command honors `PYTHON`.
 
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thanks for this. I carried your original commit into #358, merged current main, resolved the conflicts, and updated the runtime and deep-scan guidance. The full suite passes: 1,037 tests, with 11 expected skips.

PR #126, opened 2026-07-30
fix: redact credentials before storing scan failures

MBemera:
## Problem

Issue #43 identifies a credential-handling gap.

Raw model failure messages can reach `fail-scan`, SQLite, and scan history.
Terminal output is redacted, but stored failures and bulk-scan receipts do not
always receive the same protection.

## Fix

- Add one shared credential-redaction helper.
- Redact failures before truncation and storage.
- Use the helper for CLI output and bulk-scan receipts.
- Remove the duplicate bulk-scan redactor.
- Preserve protection for short authorization values.

Tests cover SQLite persistence, bulk-scan receipts, API keys, URL passwords,
query credentials, and short authorization values.

### DIFF
--- a/sdk/typescript/src/api.ts
+++ b/sdk/typescript/src/api.ts
@@ -32,6 +32,7 @@ import {
   OutputDirectoryError,
   OutputInsideProtectedRootError,
   type ProtectedScanPathKind,
+  redactedErrorMessage,
   ScanCostLimitExceededError,
   ScanInterruptedError,
 } from "./errors.js";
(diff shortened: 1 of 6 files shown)


mldangelo-oai:
Thank you for taking credential leakage seriously. `main` already has a broader centralized redaction path that covers the cases in this PR, so adopting this older version would duplicate the existing fix. I'm closing it as superseded. We'd welcome future security improvements.

MBemera:
@mldangelo-oai no problem thanks letting know and not just closing it out blind. -MB

PR #128, opened 2026-07-30
fix: prefer exact PATH during trusted executable resolution

### LINKED ISSUE #41: resolveTrustedExecutable can pick a lowercase `path` variable instead of PATH on POSIX
## Summary

`resolveTrustedExecutable` selects the search path with a case-insensitive scan that takes the **first** matching variable:

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/trusted-executable.ts#L18-L20

On Windows, environment variables are case-insensitive and this is correct. On POSIX they are not: `path` and `PATH` are two distinct variables, and `execve` consults only `PATH`. If a lowercase `path` appears earlier in the environment object, the helper searches it instead — and then installs it as the child's sanitized `PATH`.

The sibling resolver in this codebase gets this right, preferring the exact key and only then falling back to a case-insensitive scan ([`runtime.ts` lines 97-102](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/runtime.ts#L97-L102)), which is what makes this look like an oversight rather than a deliberate choice.

- macOS 26.5.2 (build 25F84)
- Node.js v24.5.0
- Bun 1.3.14

## Steps to reproduce

Place a decoy `git` in a directory referenced only by a lowercase `path`, and call the resolver directly:

```ts
const environment = {
  path: decoyDirectory,                       // contains an executable named `git`
  PATH: ["/usr/bin", "/bin"].join(delimiter), // the real search path
};
await resolveTrustedExecutable("git", environment, protectedRoot);
```

Observed results, varying only the presence and position of the lowercase variable:

```text
[PATH only]
   resolved executable: /usr/bin/git
   sanitized PATH:      /usr/bin:/bin

[lowercase `path` listed before PATH]
   resolved executable: /.../cs-path-tAHjEO/decoy-bin/git
   sanitized PATH:      /.../cs-path-tAHjEO/decoy-bin

[PATH listed before lowercase `path`]
   resolved executable: /usr/bin/git
   sanitized PATH:      /usr/bin:/bin
```

The third case shows the outcome depends purely on insertion order in the environment object, which for `process.env` mirrors `environ` order.

## Expected behavior

On POSIX the resolver should use `PATH`. Preferring the exact key, with the case-insensitive scan as a fallback, would make this consistent with `codexSecurityStateDirectory` and correct on both platforms.

## Actual behavior

A lowercase `path` that happens to precede `PATH` silently redirects executable resolution for `git` (`src/targets.ts`) and for the plugin Python interpreter (`src/runtime.ts`), and the redirected directory is what gets written into the child process's `PATH`.

## Impact

I want to be measured about severity here. Reaching this requires an environment that defines a lowercase `path` on a POSIX system, which is unusual — it is not something an attacker controls from a scanned repository. The realistic paths are an SDK embedder passing its own environment object, or a shell/tooling setup that exports a lowercase `path`.

That said, the consequence is that a function whose entire purpose is to pin a trusted, out-of-repository executable can be steered to a different one, and the benign version of the same bug is a confusing "Git is not available on a trusted PATH." on a machine where git is plainly installed.

## Suggested direction

Mirror the existing lookup in `runtime.ts`: try `environment["PATH"]` first, then fall back to the case-insensitive scan for Windows. The deletion loop at line 77 that strips every case variant before setting the sanitized `PATH` is already correct and can stay as it is.

GautamSharma99:
## Summary

Prefer the exact `PATH` environment variable when resolving trusted executables, while retaining the existing case-insensitive fallback for environments such as Windows that commonly expose the variable as `Path` or another casing.

Fixes #41.

## Problem

`resolveTrustedExecutable` previously searched all environment entries case-insensitively and selected the first name whose uppercase form was `PATH`.

On POSIX, environment names are case-sensitive. `path` and `PATH` are separate variables, and process execution consults `PATH`. If a lowercase `path` entry appeared earlier in the supplied environment object, trusted executable resolution searched that value instead of the real POSIX `PATH`.

This made resolution depend on object insertion order. It could select a different Git or Python executable, or report that no trusted executable was available even though the actual `PATH` contained one. The selected value was also published as the sanitized child `PATH`.

## Changes

- Read `environment["PATH"]` first.
- Use the existing case-insensitive environment scan only when the exact `PATH` value is unavailable.
- Preserve the existing sanitization step that removes every casing variant before publishing one canonical child `PATH`.
- Add a POSIX regression test with a lowercase `path` decoy inserted before the real `PATH`.
- Verify that the trusted executable is selected from the exact `PATH`, the decoy is ignored, both original casing variants are removed, and the child receives only the sanitized canonical `PATH`.

## Compatibility

Windows compatibility is preserved. Existing coverage continues to verify that mixed-case variables such as `Path` and `pAtH` are accepted when an exact `PATH` entry is not present.

An explicitly empty `PATH` also remains authoritative instead of unexpectedly falling back to a differently cased POSIX variable.

## Security impact

This keeps the trusted-executable boundary aligned with POSIX environment semantics. A lowercase variable supplied by an SDK embedder or inherited from local tooling can no longer redirect Git or Python resolution when a valid exact `PATH` is present.

The change does not expand the existing trust boundary or alter executable canonicalization, protected-root rejection, executable permission checks, or child-environment sanitization.

## Verification

- `pnpm run format`
- `pnpm run types`
- `pnpm run build`
- Focused trusted-executable, target normalization, and multiscan tests: 33 passed, 1 expected Windows-only skip, 0 failed
- Full TypeScript suite: 471 passed, 6 expected platform/integration skips, 0 failed
- `git diff --check`


mldangelo-oai:
@codex review Please review exact head cac738e2e95d60ec905a69b1260399459d977e97, focusing on exact POSIX PATH precedence, lowercase-path ambiguity, trusted executable selection, Windows compatibility, and the current-main merge.

mldangelo-oai:
@codex review exact head 48587857836669e836a5860c43becfca7757dc35

PR #140, opened 2026-07-30
fix: bind workbench Git to the trusted executable boundary

### LINKED ISSUE #129: Bundled workbench resolves Git from inherited PATH instead of the trusted executable boundary
## Summary

The TypeScript target and bulk-scan paths deliberately resolve Git through `resolveTrustedExecutable`, but bundled Python workbench helpers invoke `git` by name from their inherited `PATH`.

Relevant code:

- [`workbench_target.py` lines 43-68](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/_bundled_plugin/scripts/workbench_target.py#L43-L68)
- [`generate_rank_input.py` lines 490-505](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/_bundled_plugin/scripts/generate_rank_input.py#L490-L505)
- [workbench environment construction](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/api.ts#L574-L583)

`resolvePluginPython` safely resolves the Python interpreter, but returns only the executable path. The sanitized environment produced while resolving it is not retained. Workbench execution later receives the normal scan environment, and Python calls commands beginning with `"git"`.

## Why this matters

If the effective `PATH` contains a repository-controlled directory, such as a checkout-local `node_modules/.bin`, workbench registration, target inspection, snapshot calculation, or ranking preparation can execute a repository-provided Git shim with the scanning user's permissions.

This bypasses the same trust boundary already enforced and regression-tested in `targets.ts` and `multiscan.ts`. A shim can also return fabricated repository metadata, affecting target identity and integrity calculations.

## Expected behavior

Every host-side Git invocation should use one trusted, absolute executable selected outside the protected repository root, together with the sanitized environment associated with that selection.

## Actual behavior

The workbench and ranking helpers search inherited `PATH` independently.

## Suggested direction

Resolve Git once with `resolveTrustedExecutable`, pass the absolute path and sanitized Git environment into workbench commands, and make Python helpers require that path. Add regression coverage using a repository-local Git shim and assert it is never executed.

GautamSharma99:
## Summary

Fixes #129.

Bind every bundled workbench Git invocation to the SDK's trusted-executable boundary instead of allowing Python helpers to search their inherited `PATH`.

Previously, target inspection and bulk-scan checkout code resolved Git through `resolveTrustedExecutable`, but the workbench used:

```python
["git", ...]
```

The Python interpreter itself was resolved safely, but only its executable path was retained. Workbench commands and scan-time ranking helpers still inherited the original process environment, so a repository-local directory in `PATH` could supply the Git executable.

### Resolve Git at the SDK boundary

Scan preparation now resolves Git against the canonical protected repository root after the runtime and Python interpreter are ready.

The resolved Git contract contains:

- the selected trusted executable;
- a `PATH` with protected-root entries removed;
- the environment associated with that trusted selection.

The same resolved contract is shared by:

- host-side workbench registration, feedback, completion, failure, and archive commands;
- plugin helpers launched by Codex during the scan.

Standalone scan-history workbench calls resolve Git inside `runWorkbench`, using the current repository as the default protected root.

### Propagate an absolute executable

The SDK publishes the selected executable as:

```text
CODEX_SECURITY_GIT=/absolute/trusted/path/to/git
```

Python workbench helpers no longer invoke `git` by name.

The shared Python boundary verifies that the configured executable:

- is an absolute path;
- resolves to an existing regular file;
- is executable;
- remains outside the protected repository after canonicalization.

This second validation prevents a scan command from replacing `CODEX_SECURITY_GIT` with an absolute repository-local shim.

### Sanitize process state

Before workbench or scan helper execution, the runtime now:

- removes every case variant of `PATH`;
- installs the trusted resolver's sanitized `PATH`;
- removes inherited `GIT_*` variables;
- overwrites `CODEX_SECURITY_GIT` and `PYTHON` with host-selected values.

Removing inherited Git variables prevents settings such as `GIT_CONFIG_COUNT`, repository overrides, alternate object directories, or executable-bearing configuration from bypassing the trusted executable selection.

### Preserve optional Git behavior

Git remains optional for directory-only target inspection.

When no trusted executable is available:

- the SDK publishes an empty trusted-Git binding and an empty sanitized `PATH`;
- workbench target probes return the existing unavailable-Git result and fall back to directory snapshot behavior;
- diff ranking fails with a targeted message instead of searching untrusted `PATH`.

### Disable repository-configured execution hooks

Workbench Git commands continue to disable `core.fsmonitor` and now also disable hooks through `core.hooksPath=/dev/null`.

Diff ranking additionally passes:

- `--no-ext-diff`;
- `--no-textconv`.

This prevents repository-controlled external diff or text-conversion commands from being used while discovering changed paths.

## Security impact

A checkout can no longer cause host-side workbench operations or scan-time diff ranking to execute a repository-provided Git shim merely because a repository-controlled directory appears in `PATH`.

The executable identity is selected by TypeScript outside the protected root, transported explicitly, and checked again by Python before execution.

Repository-provided `GIT_*` configuration is also removed from both workbench and scan helper environments.

## Tests

Added regression coverage proving that:

- a repository-local Git shim placed first in `PATH` is not executed;
- the shim directory is removed from the workbench `PATH`;
- the real trusted Git executable is invoked successfully;
- inherited `GIT_CONFIG_*` injection does not reach workbench Python;
- scan-time plugin environments contain the absolute trusted Git binding;
- a diff-ranking command cannot override the binding with an absolute executable inside the repository;
- an unavailable trusted Git does not fall back to raw `PATH`;
- existing direct nested-worktree integrity tests receive the explicit trusted binding.

## Verification

Performed on the exact PR head:

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- focused runtime, API, workbench, and scan-recovery tests
- `pnpm run test`

Full test result:

- 472 passed
- 0 failed

### DIFF
--- a/sdk/typescript/_bundled_plugin/scripts/generate_rank_input.py
+++ b/sdk/typescript/_bundled_plugin/scripts/generate_rank_input.py
@@ -37,6 +37,7 @@ from pathlib import Path
 # Some plugin hosts launch Python with safe-path isolation enabled.
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 from rank_preview import DEFAULT_PREVIEW_BYTES, TEXT_CODE_EXTENSIONS, preview_for
+from workbench_constants import trusted_git_executable
 
 EXCLUDED_DIRS = {
     ".cache",
@@ -488,12 +489,21 @@ def bind_repo_scopes(args: argparse.Namespace) -> None:
 
 
 def run_git_changed_paths(repo: Path, diff_args: list[str]) -> list[tuple[Path, str]]:
+    git = trusted_git_executable(repo)
+    if git is None:
+        raise SystemExit("Git is unavailable on the trusted executable path.")
     result = subprocess.run(
         [
-            "git",
+            git,
+            "-c",
+            "core.fsmonitor=false",
+            "-c",
+            "core.hooksPath=/dev/null",
             "-C",
             str(repo),
             "diff",
+            "--no-ext-diff",
+            "--no-textconv",
             "--name-status",
             "-z",
             "--diff-filter=ACMRD",
--- a/sdk/typescript/_bundled_plugin/scripts/workbench_constants.py
+++ b/sdk/typescript/_bundled_plugin/scripts/workbench_constants.py
@@ -1,6 +1,8 @@
 """Shared constants for the Codex Security workbench."""
 
 import argparse
+import os
+from pathlib import Path
 
 MODES = ("diff", "standard", "deep")
 DIFF_TARGET_KINDS = ("working_tree", "commit", "range")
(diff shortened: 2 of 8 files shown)


mldangelo-oai:
Thanks for working on this. Please rebase on current `main`, resolve Git once outside the repository, and apply the trusted executable and repository-scoped `PATH` to the Python workbench and both current ranking calls. Please preserve operator-provided Git configuration, as in [#410](https://github.com/openai/codex-security/pull/410), instead of removing every `GIT_*` setting.

PR #141, opened 2026-07-30
fix: bound authentication and skill subprocess output

### LINKED ISSUE #130: Authentication and skill subprocess output can grow without bounds
## Summary

Several child-process collectors retain unlimited stdout or stderr:

- [interactive login accumulation](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/auth.ts#L30-L64)
- [noninteractive Codex command accumulation](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/auth.ts#L244-L294)
- [skill JSONL parsing](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/cli.ts#L2192-L2247)

`CodexLoginHandle` appends every chunk to `#stdout` or `#stderr`. Each chunk then calls `#notifyInstructions`, whose getters parse the complete accumulated output again. This combines unbounded storage with roughly quadratic parsing work as output grows.

`runCodex` also retains complete stdout and stderr. `readSkillCommandOutput` has no maximum JSONL line length and retains an unrestricted final agent message or error message.

## Why this matters

A malfunctioning, compromised, or unexpectedly verbose child process can exhaust the SDK/CLI process's memory. Interactive login output can additionally cause excessive CPU use because earlier output is repeatedly reparsed. A single unterminated or very large JSON event from a skill command has the same memory risk.

## Expected behavior

Subprocess output used for diagnostics or instruction discovery should have explicit byte limits. Oversized structured output should fail with a controlled, redacted error.

## Suggested direction

- Parse login instructions incrementally and keep only a bounded diagnostic tail.
- Add separate stdout/stderr ceilings to `runCodex`.
- Bound JSONL line size and retained final-message size in `readSkillCommandOutput`.
- Preserve current credential redaction when reporting truncation or overflow.
- Add tests for oversized stdout, stderr, JSONL lines, and final responses.

Existing tests cover malformed and normal-sized output but not these resource boundaries.

GautamSharma99:
## Summary

Fixes #130.

Bound the output retained from authentication subprocesses and Codex skill JSONL streams so a verbose, malfunctioning, or compromised child process cannot grow the SDK/CLI process's memory usage without limit.

The change also avoids repeatedly reparsing the complete interactive-login transcript while looking for device-auth instructions.

### Bounded authentication output

Authentication subprocesses now enforce an independent 64 KiB ceiling for stdout and stderr.

For interactive login:

- stdout and stderr are retained only up to their individual limits;
- authentication URLs and user codes are discovered incrementally from a bounded rolling tail;
- discovered instructions are cached, while preserving the existing preference for stdout over stderr;
- exceeding either limit terminates the child and produces an unsuccessful login result with a fixed diagnostic;
- collected child output is cleared on overflow so credential-like content is not included in the result;
- the successful-login callback is not invoked after an output-limit failure.

For noninteractive authentication commands:

- stdout and stderr use the same independent 64 KiB ceilings;
- exceeding either ceiling terminates the subprocess;
- the command rejects with a `PluginBootstrapError` containing a fixed, redacted message;
- collected output is discarded before the error is reported.

### Incremental login instruction parsing

`CodexLoginHandle` no longer reparses all output collected so far for every incoming chunk.

Instead, each stream keeps a 4 KiB instruction-discovery tail. New chunks are parsed together with that bounded tail, and the first discovered URL and device code are retained. This bounds the parsing work while still recognizing instructions split across normal subprocess chunks.

### Bounded skill JSONL parsing

The skill output reader now uses a byte-bounded incremental newline parser instead of `readline`, which could retain an arbitrarily large unterminated line.

The parser enforces:

- a 1 MiB maximum for each JSONL event;
- a 256 KiB maximum for retained final agent messages and error messages;
- a fixed 1 MiB pending-line buffer, avoiding unbounded chunk accumulation as well as unbounded line content.

When an oversized event is encountered, the remainder of that line is discarded and the stream continues to be drained. After draining, the command fails with a fixed `CodexSecurityError` that does not echo subprocess content. This prevents a child from being left blocked on a full output pipe while still reporting the boundary violation safely.

The parser continues to support CRLF input, a final event without a trailing newline, chunks split at arbitrary byte boundaries, and split UTF-8 sequences.

## Security impact

This prevents untrusted or unexpectedly verbose authentication and skill subprocesses from causing unbounded:

- stdout or stderr retention;
- repeated interactive-login parsing work;
- JSONL line accumulation;
- retained final response or error-message memory.

Overflow diagnostics are constant strings and do not include the offending child output, preserving the existing credential-redaction boundary.

## Tests

- oversized noninteractive authentication stdout;
- oversized noninteractive authentication stderr;
- oversized interactive login output after valid instructions are discovered;
- preservation of the discovered verification URL and device code;
- suppression of credential-like overflow content;
- rejection without invoking the login success callback;
- a JSONL event larger than 1 MiB across many chunks;
- continued stream draining after an oversized event;
- an agent response larger than 256 KiB;
- end-to-end skill command rejection without writing partial output.

## Verification

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- focused authentication and CLI skill tests
- `pnpm run test`

Full test result:

- 473 passed
- 0 failed

### DIFF
--- a/sdk/typescript/src/auth.ts
+++ b/sdk/typescript/src/auth.ts
@@ -3,6 +3,13 @@ import { isIP } from "node:net";
 import { PluginBootstrapError } from "./errors.js";
 import type { CodexCommand, ProcessEnvironment } from "./runtime.js";
 
+const MAX_CODEX_AUTH_OUTPUT_BYTES = 64 * 1024;
+const LOGIN_OUTPUT_LIMIT_MESSAGE =
+  "Codex login output exceeded the 64 KiB safety limit.";
+const COMMAND_OUTPUT_LIMIT_MESSAGE =
+  "Codex authentication command output exceeded the 64 KiB safety limit.";
+const INSTRUCTION_TAIL_BYTES = 4 * 1024;
+
 export interface LoginResult {
   success: boolean;
   exitCode: number | null;
@@ -29,6 +36,13 @@ export class CodexLoginHandle {
   #canceled = false;
   #stdout = "";
   #stderr = "";
+  #stdoutInstructionTail = "";
+  #stderrInstructionTail = "";
+  #stdoutAuthUrl: string | null = null;
+  #stderrAuthUrl: string | null = null;
+  #stdoutUserCode: string | null = null;
+  #stderrUserCode: string | null = null;
+  #outputLimitExceeded = false;
 
   public constructor(
     command: CodexCommand,
@@ -55,12 +69,10 @@ export class CodexLoginHandle {
     this.#child.stdout.setEncoding("utf8");
     this.#child.stderr.setEncoding("utf8");
     this.#child.stdout.on("data", (chunk: string) => {
-      this.#stdout += chunk;
-      this.#notifyInstructions();
+      this.#recordOutput("stdout", chunk);
     });
     this.#child.stderr.on("data", (chunk: string) => {
-      this.#stderr += chunk;
-      this.#notifyInstructions();
+      this.#recordOutput("stderr", chunk);
     });
     this.#completion = new Promise((resolve, reject) => {
       this.#child.once("error", (error) => {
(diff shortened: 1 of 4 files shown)


mldangelo-oai:
Thanks for tackling unbounded authentication and skill subprocess output. Those limits are already in the current code, and this branch now conflicts with `main`, so it would not add a new fix. I'll close it as superseded. We'd be glad to consider future contributions.

PR #144, opened 2026-07-30
fix: guarantee bounded authentication and skill shutdown

### LINKED ISSUE #131: SDK close can hang indefinitely when a login child ignores SIGTERM
## Summary

Cancellation sends one signal and then waits without a deadline:

- [`CodexLoginHandle.cancel()`](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/auth.ts#L128-L139)
- [`CodexSecurity.close()`](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/api.ts#L983-L1005)
- [CLI child signal forwarding](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/cli.ts#L429-L470)

`cancel()` sends `SIGTERM` once. `close()` then awaits every login handle with `Promise.allSettled` and has no timeout or escalation path.

A child can ignore `SIGTERM`. A descendant can also inherit stdout/stderr and keep the pipes open after the immediate child exits. Either case can leave SDK shutdown waiting indefinitely. The interactive-login implementation has a Windows-only pipe fallback, but no equivalent POSIX deadline or forced termination.

## Why this matters

An application embedding the SDK can become permanently stuck during cleanup. CLI commands using the same one-signal pattern may also fail to return after cancellation.

## Expected behavior

Cancellation should allow a short cooperative grace period, then guarantee that child-process resources settle.

## Suggested direction

- Record cancellation and send the normal termination signal.
- After a bounded grace period, escalate to `SIGKILL` or the platform-appropriate process-tree termination mechanism.
- Explicitly destroy remaining pipes before settling completion.
- Ensure timers are cleared after normal exit.
- Test with a child that ignores `SIGTERM` and a descendant that retains inherited pipes.

The current tests use children that cooperate with `SIGTERM`, so this path is not exercised.

This is separate from #44: that issue concerns a canceled scan remaining marked `RUNNING`; this issue concerns the cancellation operation itself never settling.

GautamSharma99:
## Summary

Fixes #131.

Make authentication and skill subprocess cancellation settle within a bounded deadline, even when the immediate child ignores the cooperative termination signal or a descendant keeps inherited output pipes open.

Previously, `CodexLoginHandle.cancel()` sent `SIGTERM` once and `CodexSecurity.close()` waited on the login handle without a deadline. The CLI skill path used the same one-signal behavior. A non-cooperative child or inherited pipe could therefore leave SDK or CLI cleanup pending indefinitely.

### Bounded login cancellation

Interactive login cancellation now uses a two-stage shutdown:

1. mark the login as canceled and send `SIGTERM`;
2. allow a one-second cooperative grace period;
3. if the child is still running, escalate to `SIGKILL`;
4. destroy the remaining stdin, stdout, and stderr streams;
5. force the login completion to settle if process events or inherited pipes still do not do so.

The forced completion remains an unsuccessful canceled login, so the authentication success callback cannot run after cancellation.

Repeated calls to `cancel()` reuse the existing deadline instead of creating additional timers.

### Cross-platform inherited-pipe fallback

The existing post-exit pipe fallback was Windows-only. It now applies on every platform.

After the immediate login child exits, normal pipe draining still receives a one-second grace period. If a descendant continues holding stdout or stderr open after that deadline, the SDK destroys the streams and settles the login using the immediate child's exit status.

This preserves ordinary diagnostic draining while preventing descendants from keeping `login.wait()` and `CodexSecurity.close()` alive forever.

### Timer and listener cleanup

Login completion and spawn failure now clear both the post-exit fallback and forced-termination timer. Process pipes are released on every terminal path, and late `error`, `exit`, or `close` events cannot settle the same login twice.

### Bounded CLI signal forwarding

`runCodexSkillCommand` now applies the same one-second termination deadline when forwarding `SIGINT` or `SIGTERM`:

- the original signal is forwarded first;
- a child that remains alive is escalated with `SIGKILL`;
- captured stdout and stderr are destroyed after the grace period;
- the JSONL capture wait and process-status wait are explicitly released;
- conventional exit status is preserved as 130 for `SIGINT` and 143 for `SIGTERM`;
- signal listeners and forced-termination timers are removed on every exit path.

Explicitly releasing the capture wait is necessary because destroying a stream alone does not reliably settle every parser or inherited-pipe configuration.

## Impact

SDK consumers can now rely on `CodexSecurity.close()` completing after it cancels an interactive login. CLI skill commands also return after cancellation instead of waiting indefinitely for an uncooperative child or inherited output pipe.

The cooperative path is unchanged: children still receive the original signal and have time to clean up normally before forced termination occurs.

## Tests

Added and updated regression coverage for:

- a login child that installs a `SIGTERM` handler and refuses to exit;
- direct `CodexLoginHandle.cancel()` completion within a bounded deadline;
- `CodexSecurity.close()` completion when its active login ignores `SIGTERM`;
- preservation of an unsuccessful login result and suppression of the success callback;
- post-exit inherited login-pipe cleanup through the cross-platform fallback;
- a real CLI wrapper receiving `SIGTERM` while its skill child ignores the signal;
- a descendant retaining the skill child's captured stdout and stderr;
- forced CLI completion with exit code 143;
- cleanup of all fixture subprocesses after the process-level test.

## Verification

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- focused authentication, SDK close, and CLI skill tests

Full test result:

- 472 passed
- 0 failed

### DIFF
--- a/sdk/typescript/src/auth.ts
+++ b/sdk/typescript/src/auth.ts
@@ -3,6 +3,8 @@ import { isIP } from "node:net";
 import { PluginBootstrapError } from "./errors.js";
 import type { CodexCommand, ProcessEnvironment } from "./runtime.js";
 
+const LOGIN_CHILD_TERMINATION_GRACE_MS = 1_000;
+
 export interface LoginResult {
   success: boolean;
   exitCode: number | null;
@@ -27,6 +29,8 @@ export class CodexLoginHandle {
   #urlReadySettled = false;
   #deviceReadySettled = false;
   #canceled = false;
+  #forcedTermination: ReturnType<typeof setTimeout> | undefined;
+  #forceCompletion: (() => void) | null = null;
   #stdout = "";
   #stderr = "";
 
@@ -63,21 +67,15 @@ export class CodexLoginHandle {
       this.#notifyInstructions();
     });
     this.#completion = new Promise((resolve, reject) => {
-      this.#child.once("error", (error) => {
-        this.#settleInstructionWaiters({
-          success: false,
-          exitCode: null,
-          stdout: this.#stdout,
-          stderr: error.message,
-        });
-        reject(error);
-      });
       let fallback: ReturnType<typeof setTimeout> | undefined;
       let completed = false;
       const complete = (exitCode: number | null): void => {
         if (completed) return;
         completed = true;
         if (fallback !== undefined) clearTimeout(fallback);
+        this.#clearForcedTermination();
+        this.#forceCompletion = null;
+        this.#destroyPipes();
         const result = {
           success: exitCode === 0 && !this.#canceled,
           exitCode,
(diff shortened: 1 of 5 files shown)


mldangelo-oai:
Thanks for working through bounded shutdown for authentication and skills. The corresponding shutdown handling is already implemented on `main`, so this PR no longer has a separate fix for us to merge. I'm closing it as superseded. We'd be happy to review future work in this area.

PR #145, opened 2026-07-30
fix: bound bulk-scan inventory loading

### LINKED ISSUE #132: Bulk-scan inventory loading has no input-size or repository-count limit
## Summary

Bulk scan materializes inventories without resource ceilings:

- [CSV loading in `runMultiscan`](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/multiscan.ts#L75-L89)
- [full Papa Parse materialization](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/multiscan.ts#L350-L413)
- [GitHub repository accumulation](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/bulk-scan-discovery.ts#L244-L279)

A supplied CSV is read completely into a UTF-8 string, parsed into a complete row array, and then mapped into another task array. There is no maximum file size, row count, or campaign repository count.

Interactive discovery similarly accumulates every qualifying repository before selection and inventory creation.

## Why this matters

An accidentally huge or adversarial CSV can cause excessive memory use before any checkout begins. A very large organization can produce the same behavior through discovery. It can also create a campaign whose checkout, scan, and model-cost scope is much larger than the user expects.

Per-field validation does not bound the number of fields or rows and therefore does not address aggregate resource use.

## Expected behavior

Bulk scan should reject an inventory that exceeds documented byte and repository-count ceilings before campaign state or checkouts are created.

## Suggested direction

- Check CSV metadata against a maximum byte size before reading.
- Enforce a maximum repository/task count during parsing and GitHub discovery.
- Prefer incremental CSV parsing to avoid retaining the source string and every parser row simultaneously.
- Stop GitHub pagination with an actionable error once the campaign limit is reached.
- Document the limits and add boundary/overflow tests.

Existing tests cover CSV correctness and repository pagination but not oversized inputs or excessive task counts.

GautamSharma99:
## Summary

Fixes #132.

Add explicit resource ceilings to bulk-scan inventory loading and interactive GitHub discovery, and replace full CSV materialization with incremental parsing.

Previously, a supplied inventory was read into one complete UTF-8 string, Papa Parse retained a complete row array, and the rows were then mapped into another task array. Interactive discovery similarly accumulated every qualifying GitHub repository without a campaign limit. Both paths could consume excessive memory and create unexpectedly large checkout, scan, and model-cost scope before work began.

## Limits

Bulk scan now enforces shared limits of:

- 8 MiB per CSV inventory;
- 1,000 repositories per campaign.

The same repository ceiling applies to supplied inventories and interactive GitHub discovery.

### Metadata validation before campaign state

`runMultiscan` now checks inventory metadata before opening or parsing the CSV.

It rejects:

- inputs that are not regular files;
- files larger than 8 MiB;
- inventories containing more than 1,000 repository rows.

All inventory parsing and validation still happens before the output directory, campaign lock, manifest, checkout directory, artifact directory, or security client is created.

### Incremental CSV parsing

Inventory loading now streams the file through Papa Parse in 64 KiB input chunks.

The parser:

- processes the header once;
- validates and converts one repository row at a time;
- retains only the bounded task list and duplicate-ID set;
- preserves quoted fields, embedded commas, escaped quotes, CRLF input, blank-line skipping, and UTF-8 BOM handling;
- preserves existing header, field-count, ID, revision, mode, scope, and repository validation;
- checks cancellation between rows.

This removes the previous simultaneous retention of the complete source string, complete Papa row array, and mapped task array.

### In-stream byte enforcement

The initial metadata check rejects known oversized files before reading. A byte-counting transform independently enforces the 8 MiB ceiling while the file is read.

This second check prevents a file replacement or growth race after metadata validation from bypassing the input limit.

The source and parser streams are destroyed immediately when parsing, cancellation, byte, or repository-count validation fails.

### Bounded GitHub discovery

Interactive discovery now stops and returns an actionable error as soon as the 1,001st qualifying repository is encountered.

The limit is enforced during pagination, before repository selection, output-directory creation, or inventory writing. Empty repositories and repositories older than the existing 90-day cutoff continue to be excluded before they count toward the campaign limit.

The error recommends splitting the repository list into smaller campaigns.

### Shared limits and documentation

The byte and repository constants, along with the repository-limit diagnostic, are shared by CSV loading and GitHub discovery so the two entry points cannot drift.

The README now documents both limits and advises using separate inventories and output directories for larger campaigns.

## Security and resource impact

This bounds:

- bytes read from a supplied inventory;
- parser source buffering;
- retained repository tasks;
- interactive GitHub repository accumulation;
- the maximum checkout and scan scope of one campaign.

Oversized input is rejected before persistent campaign state or checkout work begins.

## Tests

- rejecting an inventory whose metadata exceeds 8 MiB;
- rejecting a valid 1,001-row inventory;
- ensuring neither case creates a security client;
- ensuring neither case creates the output directory or campaign state;
- accepting exactly 1,000 repositories during interactive discovery;
- paginating exactly ten 100-repository GitHub pages at the boundary;
- rejecting the 1,001st discovered repository;
- stopping before an interactive inventory or output directory is created;
- preserving the existing quoted CSV, BOM, CRLF, validation, pagination, cutoff, and selection behavior.

## Verification

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- focused multiscan and bulk-discovery tests

Full test result:

- 472 passed
- 0 failed

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -405,6 +405,11 @@ service,https://github.com/acme/service.git,0123456789abcdef0123456789abcdef0123
 `--workers` limits concurrent scans and `--max-attempts` retries failures.
 Results remain under `--output-dir`; rerun the same command to resume.
 
+Bulk-scan inventories are limited to 8 MiB and 1,000 repositories. Interactive
+GitHub discovery applies the same repository limit and stops before creating an
+inventory if the selected account exceeds it. Split larger campaigns into
+separate repository lists and output directories.
+
 ### Scan history and reruns
 
 `npx @openai/codex-security scans list` lists scans for the current repository. Pass a
--- a/sdk/typescript/src/bulk-scan-discovery.ts
+++ b/sdk/typescript/src/bulk-scan-discovery.ts
@@ -8,6 +8,10 @@ import { promisify } from "node:util";
 import { confirm, input, search } from "@inquirer/prompts";
 import { Octokit } from "@octokit/core";
 import Papa from "papaparse";
+import {
+  bulkScanRepositoryLimitError,
+  MAX_BULK_SCAN_REPOSITORIES,
+} from "./bulk-scan-limits.js";
 import { resolveTrustedExecutable } from "./trusted-executable.js";
 
 const execFile = promisify(execFileCallback);
@@ -270,6 +274,9 @@ async function discoverGitHubRepositories(
         url: `https://${host}/${repository.nameWithOwner}.git`,
         revision: repository.defaultBranchRef.target.oid.toLowerCase(),
       });
+      if (repositories.length > MAX_BULK_SCAN_REPOSITORIES) {
+        throw bulkScanRepositoryLimitError();
+      }
     }
     cursor = connection.pageInfo.hasNextPage
       ? connection.pageInfo.endCursor
--- a/sdk/typescript/src/bulk-scan-limits.ts
+++ b/sdk/typescript/src/bulk-scan-limits.ts
@@ -1,0 +1,8 @@
+export const MAX_BULK_SCAN_INVENTORY_BYTES = 8 * 1_024 * 1_024;
+export const MAX_BULK_SCAN_REPOSITORIES = 1_000;
+
+export function bulkScanRepositoryLimitError(): Error {
+  return new Error(
+    "Bulk scans support at most 1,000 repositories per campaign. Split the repository list into smaller inventories.",
+  );
+}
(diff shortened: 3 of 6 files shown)


ianw-oai:
Similar to #95, I appreciate it but I don’t think this addresses a real problem for us.

PR #146, opened 2026-07-30
fix: enforce plugin entry limit during discovery

### LINKED ISSUE #134: Configured plugin directory entry limit is applied after unbounded readdir
## Summary

Configured plugin directories have a nominal 4,096-entry safety limit, but directory children are loaded completely before that limit is enforced.

Relevant paths:

- [plugin fingerprint traversal](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/runtime.ts#L1109-L1137)
- [plugin copy traversal](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/runtime.ts#L1655-L1699)

Both loops increment `entries` when a queued path is processed. When that path is a directory, they call `readdir(path)` and push all returned children. A single directory with a very large number of entries is therefore fully allocated before any of those children reach the 4,096-entry check.

Cancellation is checked between queued paths but cannot interrupt the individual full-directory read.

## Why this matters

A malicious or accidentally enormous configured plugin can cause significant memory consumption despite the presence of an apparent entry limit. ZIP plugins are checked entry-by-entry during extraction, so directory plugins currently have a weaker resource boundary than archive plugins.

## Expected behavior

Traversal should reject the 4,097th discovered entry without first materializing every name in the containing directory, and cancellation should be observed during discovery.

## Suggested direction

Use `opendir()` and its asynchronous iterator in both fingerprinting and copying. Increment and enforce the shared entry count as each child is yielded, check the abort signal inside the iterator loop, and preserve the existing symlink/identity checks.

Add a regression test with more than 4,096 children in one directory and a cancellation test that aborts during directory iteration.

GautamSharma99:
## Summary

Fixes #134.

Enforce the configured-plugin 4,096-entry ceiling while directory entries are discovered instead of after an entire directory has already been loaded into memory.

Previously, both plugin fingerprinting and marketplace copying called `readdir()` for each directory. A directory containing hundreds of thousands of children was therefore materialized as one complete name array before any of those children reached the existing entry counter. Cancellation also could not be observed until that full read completed.

### Incremental directory iteration

Configured plugin traversal now uses one shared asynchronous `opendir()` iterator for both:

- plugin projection fingerprint discovery;
- plugin marketplace copy discovery.

The iterator reads one `Dirent` at a time with `Dir.read()` and yields only its name. No traversal path creates a complete array of directory children.

### Discovery-time entry enforcement

The traversal root remains counted as the first entry, preserving the existing 4,096-entry semantics.

Each child is now counted immediately after it is yielded and before it is added to the traversal queue. The 4,097th discovered path therefore raises the existing `PluginBootstrapError` immediately, without reading the remaining names in that directory.

This applies identically to fingerprint construction and safe-copy traversal.

### Cancellation during discovery

The abort signal is checked:

- before opening a directory;
- before each individual directory read;
- after each read completes;
- before the discovered entry is yielded to the traversal.

Cancellation can now interrupt a large, flat directory rather than waiting for every name to be allocated first.

### Directory-handle cleanup

The shared iterator closes its directory handle in a `finally` block. Handles are therefore released after:

- normal end-of-directory;
- cancellation;
- entry-limit rejection;
- any other traversal error.

### Existing integrity boundaries preserved

The change retains the existing protections around configured plugin directories:

- source paths are still checked with `lstat()` before processing;
- symbolic links and non-regular files remain rejected;
- copied directories are still re-stat'd after enumeration;
- directory device/inode, size, modification time, mode, and type identity must remain unchanged;
- queued descendants still revalidate every recorded ancestor before copying;
- per-file and aggregate byte limits are unchanged;
- partially created marketplace trees are removed on failure.

Fingerprint paths remain sorted before hashing, so switching to incremental directory iteration does not change fingerprint determinism.

## Security and resource impact

A malicious or accidentally enormous flat plugin directory can no longer bypass the practical value of the configured 4,096-entry limit by forcing one unbounded `readdir()` allocation.

Directory plugins now have the same entry-by-entry resource-boundary behavior already used for ZIP plugin extraction, and cancellation remains responsive throughout discovery.

## Tests

- a configured plugin containing 4,096 additional siblings in one directory;
- rejection through the existing copy-entry limit;
- removal of the partially created marketplace after overflow;
- cancellation triggered during the second individual directory read;
- preservation of the original `AbortError` reason;
- confirmation that discovery stops immediately at cancellation;
- removal of partial marketplace output after cancellation.

Existing configured-plugin, marketplace projection, symlink rejection, queued-directory identity race, ZIP limit, and runtime tests continue to pass.

## Verification

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- complete runtime test file

Full test result:

- 472 passed
- 0 failed

### DIFF
--- a/sdk/typescript/src/runtime.ts
+++ b/sdk/typescript/src/runtime.ts
@@ -8,6 +8,7 @@ import {
   mkdir,
   mkdtemp,
   open,
+  opendir,
   readFile,
   readdir,
   realpath,
@@ -1108,24 +1109,25 @@ async function pluginProjectionFingerprint(
   } else {
     paths = [];
     const pending = [canonical];
-    let entries = 0;
+    let entries = 1;
     while (pending.length > 0) {
       throwIfSignalAborted(signal);
       const path = pending.pop()!;
       const metadata = await lstat(path);
-      if (++entries > MAX_PLUGIN_COPY_ENTRIES) {
-        throw new PluginBootstrapError(
-          `Plugin source exceeds the copy entry limit: ${path}`,
-        );
-      }
       if (metadata.isSymbolicLink()) {
         throw new PluginBootstrapError(
           `Plugin contains an unsafe source path: ${path}`,
         );
       }
       if (metadata.isDirectory()) {
-        for (const entry of await readdir(path)) {
-          pending.push(join(path, entry));
+        for await (const entry of pluginDirectoryEntries(path, signal)) {
+          const child = join(path, entry);
+          if (++entries > MAX_PLUGIN_COPY_ENTRIES) {
+            throw new PluginBootstrapError(
+              `Plugin source exceeds the copy entry limit: ${child}`,
+            );
+          }
+          pending.push(child);
         }
       } else if (metadata.isFile()) {
         paths.push(relative(canonical, path).split(sep).join("/"));
@@ -1661,7 +1663,7 @@ async function copyPluginTree(
     { source, destination },
   ];
   const directories = new Map<string, Stats>();
-  let entries = 0;
+  let entries = 1;
   let size = 0;
   await mkdir(dirname(destination), { recursive: true, mode: 0o700 });
   try {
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thanks for taking the time to put guardrails around plugin discovery. The same entry-limit behavior is already on `main`, so merging this branch would repeat work that has shipped. I'll close it as superseded, and we'd welcome another focused contribution.

PR #147, opened 2026-07-30
fix: recover malformed multiscan receipt ledgers

### LINKED ISSUE #133: Malformed complete multiscan receipts permanently block campaign resume
## Summary

Multiscan receipt recovery handles only one corruption shape: a final line without a terminating newline.

Relevant code: [`readReceipts` lines 312-335](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/multiscan.ts#L312-L335).

The complete ledger is read into memory and split. Every newline-terminated line is passed directly to `JSON.parse`, cast to `MultiscanReceipt`, and immediately accessed through `receipt.id.toLowerCase()`.

Consequently, any of the following throws on every future resume:

- malformed JSON in the middle of the ledger;
- a malformed final line that does end with a newline;
- valid JSON that is not an object;
- a receipt with a missing or non-string `id`;
- structurally invalid attempt, status, or path fields.

The ledger also has no total-size or per-line limit.

## Why this matters

A crash, disk fault, partial external write, or manual damage to one durable record can wedge the entire campaign until the user edits internal state. The current recovery behavior may give users the impression that receipt-ledger interruption is generally recoverable, although only a torn final line is repaired.

## Expected behavior

Receipt loading should either recover safely from invalid durable records or stop with an actionable corruption diagnostic and a supported recovery path. It should never fail with an incidental JSON/type exception indefinitely.

## Suggested direction

Stream the ledger with a maximum line and total size, schema-validate every receipt, and distinguish a recoverable tail from corruption of committed history. Invalid committed lines could be moved to a quarantine file before resuming, with the original ledger retained for diagnosis.

Add tests for malformed newline-terminated JSON, non-object JSON, missing IDs, invalid attempts/statuses, and oversized lines.

GautamSharma99:
## Summary

Fixes #133.

Bulk-scan resume now treats `results.jsonl` as bounded, validated durable state instead of reading the whole file and trusting every parsed value as a `MultiscanReceipt`.

A malformed committed record no longer wedges every future resume. The loader keeps valid history, quarantines the original damaged ledger for diagnosis, publishes a repaired ledger, and continues the campaign.

### Bounded streaming

Receipt loading now:

- streams the ledger in 64 KiB chunks instead of materializing the complete file;
- enforces a 64 MiB total ledger limit;
- enforces a 1 MiB limit for each JSONL record;
- uses a fixed-size pending-line buffer and discards oversized lines without retaining them;
- checks the total again while reading so file growth cannot bypass the metadata check;
- observes the campaign abort signal during ledger discovery and parsing.

An oversized ledger is retained under a `results.corrupt-<uuid>.jsonl` name and replaced with an empty active ledger, allowing the campaign to safely rescan instead of exhausting memory or failing indefinitely.

### Receipt validation

Every non-empty line is decoded as strict UTF-8, parsed as JSON, and structurally validated before it can affect resume state. Validation covers:

- safe task IDs;
- bounded repository and scope values;
- full immutable Git revisions;
- supported scan modes;
- completed or failed status values;
- positive safe-integer attempt numbers;
- bounded absolute output paths;
- optional cost fields and token counts;
- optional error values.

Malformed JSON, non-object JSON, missing IDs, invalid statuses or attempts, unsafe paths, invalid costs, and torn final records are excluded from the active ledger.

### Quarantine and repair

When corruption is found, valid records are written to a private repair file and flushed before publication. The original ledger is retained as `results.corrupt-<uuid>.jsonl`, while the repaired set becomes the new `results.jsonl`.

This preserves usable attempt and completion history without hiding the damaged bytes that caused recovery. A normal, valid ledger is left unchanged.

### Bounded generated receipts

New receipts are rejected if they exceed the per-record ceiling. Stored failure messages are also capped at 64 KiB after credential redaction, and inventory scopes now use the same 4,096-character bound enforced by receipt validation.

### Documentation

The TypeScript SDK README now documents the ledger and record limits, quarantine naming, valid-history repair behavior, and oversized-ledger rescan behavior.

## Tests

Added regression coverage for committed corruption including:

- malformed JSON;
- non-object JSON;
- missing IDs;
- invalid attempts;
- invalid statuses;
- invalid output paths;
- oversized JSONL records;
- oversized receipt ledgers.

The tests verify that valid history remains usable, completed repositories stay skipped when their valid receipt survives repair, the original damaged ledger is retained, and an oversized ledger causes a clean rescan.

## Verification

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- focused multiscan tests
- `pnpm run test`

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -405,6 +405,12 @@ service,https://github.com/acme/service.git,0123456789abcdef0123456789abcdef0123
 `--workers` limits concurrent scans and `--max-attempts` retries failures.
 Results remain under `--output-dir`; rerun the same command to resume.
 
+Bulk-scan receipt ledgers are limited to 64 MiB total and 1 MiB per record.
+If a committed record is malformed, the original ledger is retained as
+`results.corrupt-<uuid>.jsonl` and valid records are republished to
+`results.jsonl` before the campaign resumes. An oversized ledger is quarantined
+whole and the campaign safely rescans its repositories.
+
 ### Scan history and reruns
 
 `npx @openai/codex-security scans list` lists scans for the current repository. Pass a
--- a/sdk/typescript/src/multiscan.ts
+++ b/sdk/typescript/src/multiscan.ts
@@ -1,5 +1,6 @@
 import { execFile as execFileCallback } from "node:child_process";
 import { randomUUID } from "node:crypto";
+import { createReadStream } from "node:fs";
 import {
   lstat,
   mkdir,
@@ -8,7 +9,6 @@ import {
   realpath,
   rename,
   rm,
-  truncate,
   writeFile,
 } from "node:fs/promises";
 import { dirname, isAbsolute, join, relative, resolve, sep } from "node:path";
@@ -27,6 +27,9 @@ const REQUIRED_ARTIFACTS = [
   "coverage.json",
   "report.md",
 ];
+const MAX_RECEIPT_LEDGER_BYTES = 64 * 1024 * 1024;
+const MAX_RECEIPT_LINE_BYTES = 1024 * 1024;
+const MAX_RECEIPT_ERROR_BYTES = 64 * 1024;
 
 interface MultiscanTask {
   id: string;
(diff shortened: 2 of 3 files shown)


rajpratham1 (approved):
This pull request significantly improves the resilience of multiscan receipt handling by replacing the previous full-file parsing approach with a streaming parser that validates individual receipt records, enforces size limits on both the ledger and individual entries, and safely quarantines corrupted ledgers before resuming execution. Valid receipts are preserved and republished when possible, while oversized or malformed ledgers are isolated to prevent unreliable resume behavior. The changes also bound stored error message sizes and add comprehensive tests covering malformed JSON, invalid receipt fields, oversized records, oversized ledgers, and recovery workflows. Based on the visible changes, the implementation is focused on robustness and recovery without introducing any apparent blocking issues.

PR #148, opened 2026-07-30
fix: paginate bulk-scan organization selection

### LINKED ISSUE #135: Bulk-scan account selection only lists the first 100 GitHub organizations
## Summary

The interactive bulk-scan account selector requests one page from `GET /user/orgs`:

[`selectGitHubOwner` lines 195-214](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/bulk-scan-discovery.ts#L195-L214)

The request sets `per_page: 100` but does not follow pagination. Users belonging to more than 100 organizations therefore see only the first page plus their personal account.

This is inconsistent with repository discovery immediately below it, which correctly follows GraphQL pagination until the cutoff or final page.

## Why this matters

Organizations beyond the first page cannot be selected for bulk scanning. The prompt does not indicate that the account list is incomplete, so the behavior looks like missing permissions or an authentication problem.

The ordering can also change as organization memberships change, making access to later organizations unpredictable.

## Expected behavior

Every organization visible to the authenticated GitHub account should be available in the selector, subject to a clearly documented safety ceiling if one is needed.

## Suggested direction

Use Octokit's pagination helper or follow REST `Link` headers, forwarding the abort signal to every request. Deduplicate organizations before sorting alongside the personal account.

Add a test server response with at least two organization pages and verify:

- both pages are requested;
- a second-page organization is selectable;
- cancellation stops further pagination;
- the personal account remains present.

The existing tests return all configured organizations in a single response and therefore do not exercise REST pagination.

GautamSharma99:
## Summary

Fixes #135.

The interactive bulk-scan wizard now loads every page returned by `GET /user/orgs` before asking which account to scan. Previously it requested only `per_page: 100`, so organization memberships after the first page were silently absent from the selector.

### REST pagination

Organization discovery now follows GitHub REST pagination one page at a time:

- requests up to 100 organizations per page;
- checks the response `Link` header for `rel="next"`;
- increments the page only when GitHub advertises another page;
- keeps the personal-account lookup concurrent with organization discovery.

This works through the same Octokit request boundary for GitHub.com and configured GitHub Enterprise hosts.

### Cancellation

The campaign abort signal is forwarded to every organization-page request. It is also checked:

- before each page request;
- immediately after each response;
- after the organization and personal-account requests settle;
- before the account prompt is shown.

Cancellation therefore stops pagination without requesting another page, opening the selector, or starting repository discovery.

### Account normalization

The combined personal-account and organization list is deduplicated case-insensitively before sorting. This prevents duplicate selector entries if repeated organization memberships appear across pages while preserving the login spelling returned by GitHub.

The authenticated personal account remains available alongside all organizations.

## Tests

Added a paginated REST fixture and regression coverage that verifies:

- both organization pages are requested;
- an organization returned only on page two appears in the selector;
- the second-page organization can be selected and is used for GraphQL repository discovery;
- the personal account remains in the selector;
- duplicate organization logins across pages are shown only once;
- cancellation after page one prevents page two from being requested;
- cancellation prevents both the account prompt and repository discovery.

Existing single-page organization selection, GitHub Enterprise, repository pagination, authentication, and cancellation tests continue to pass.

## Verification

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- focused bulk-scan discovery tests


rajpratham1 (approved):
This pull request improves bulk scan organization selection by replacing the single-page organization lookup with paginated retrieval, ensuring users with more than 100 GitHub organizations can discover all available accounts. The implementation also deduplicates organization names across pages while preserving the original casing, and checks for cancellation during pagination so aborted operations terminate cleanly before repository discovery begins. The accompanying tests verify multi-page retrieval, deduplication, pagination behavior, and cancellation handling. Based on the visible changes, the implementation addresses the stated limitation without introducing any apparent blocking issues.

mldangelo-oai:
@codex review Please review exact head 5ea8206fcb7c6f5c60da05c2197dccfdf4dd0ba5, focusing on paginated organization discovery, case-insensitive deduplication, cancellation, pagination termination, and compatibility with the current-main merge.

mldangelo-oai:
@codex review exact head b7cb10f406e6acbd556fee210c026478989b9331

mldangelo-oai:
Thank you for the thoughtful implementation and review work. Organization pagination and the 101-organization regression test have since landed in [#392](https://github.com/openai/codex-security/pull/392), so this pull request is superseded by the implementation already on `main`.

PR #151, opened 2026-07-30
fix: bound incremental cost session events

### LINKED ISSUE #138: Cost tracker keeps an unbounded partial JSONL event and repeatedly recopies it
## Summary

The incremental cost-session reader has no maximum JSONL event size:

[`readSessionUsage` lines 186-226](https://github.com/openai/codex-security/blob/9c7634b5d08b36e6060441fe3e35caaaf8aa6a49/sdk/typescript/src/cost.ts#L186-L226)

After each 64 KiB read, it concatenates `session.remainder` with the new buffer. Everything after the last newline is retained as the next remainder.

If a session file contains a very large unterminated line, or is corrupted so that no newline arrives, the remainder grows to the size of the file. Each iteration recopies the entire previous remainder through `Buffer.concat`, producing unbounded memory growth and roughly quadratic copying.

## Why this matters

A malformed or unexpectedly large Codex session event can consume substantial memory during cost polling. Because `maxCostUsd` enforcement depends on this reader completing and reporting usage, the same condition can delay the budget abort and scan shutdown.

Normal partial writes make a remainder necessary, but they do not require retaining an unlimited event.

## Expected behavior

The tracker should accept ordinary partial writes while rejecting or ignoring a session once one event exceeds a documented maximum size.

## Suggested direction

Track pending-line bytes and stop reading/quarantine the session after a fixed maximum. Avoid repeatedly concatenating the entire remainder; accumulate bounded chunks or scan new buffers incrementally. Surface the failure through the existing tracker error path without including raw session content.

Add tests for a line exceeding the limit across multiple reads, a normal event split across reads, and recovery when a newline arrives below the limit.

This is related to cost tracking but distinct from #31, which concerns queued full session-tree rescans rather than the JSONL parser's line buffer.

GautamSharma99:
## Summary

Bound the incremental Codex session JSONL reader so a malformed or unexpectedly large unterminated event cannot consume unbounded memory or trigger quadratic buffer copying during cost polling.

Fixes #138.

### Bounded event parsing

Cost-session JSONL events are now limited to 1 MiB. The reader enforces the limit while each 64 KiB file chunk is processed, so an oversized event is rejected before it can continue accumulating in memory.

The error is intentionally fixed and does not include any session content:

> Codex session event exceeds the 1 MiB safety limit.

This error flows through the tracker's existing `onError` path, which lets a budgeted scan abort when its running cost can no longer be verified safely.

### Linear incremental buffering

The previous implementation concatenated the complete partial line with every new read. For an unterminated event, that repeatedly copied all bytes seen so far and made parsing approximately quadratic.

The new parser stores bounded fragments and tracks their aggregate byte count. Each fragment is copied once while the event is incomplete, and the fragments are concatenated only once when a newline completes the event. Events contained within a single read are parsed directly without an intermediate concatenation.

This preserves normal JSONL append behavior, including events split across multiple reads and multiple polling cycles.

### Session quarantine

When a session exceeds the event limit, it is marked unreadable and its retained fragments are released. Later polling cycles skip that session instead of repeatedly revisiting the same invalid stream or retaining its partial contents.

Already parsed metadata and token usage remain available in the current snapshot, but the overflow is still surfaced immediately through the tracker error path.

### Documentation

The TypeScript README now documents the 1 MiB cost-session event limit and explains that an oversized event stops a budgeted scan because live cost can no longer be verified safely.

## Why this matters

A corrupt session file or a very large partial write can no longer:

- grow the tracker process's memory without a ceiling;
- repeatedly recopy an ever-growing partial event;
- delay cost-budget enforcement while the parser processes unbounded input;
- expose raw session content in the resulting diagnostic.

## Tests

- a valid event split across reads and polling cycles;
- a partial event that spans multiple 64 KiB reads while remaining below the limit;
- an event exceeding 1 MiB without a newline;
- release and quarantine of oversized session state;
- continued access to the last valid cost snapshot after quarantine.

Verification performed:

- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- focused cost-tracker tests
- `pnpm run test`

Full suite result:

- 472 passed
- 0 failed

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -380,7 +380,9 @@ including cached input and cache writes; fees and surcharges are not included.
 
 Use `--max-cost USD` to stop a scan, including its delegated workers, when its
 running cost exceeds the limit. Partial results are preserved. Requests
-already in progress can finish above the limit.
+already in progress can finish above the limit. Cost tracking accepts Codex
+session events up to 1 MiB; an oversized event stops the scan because its
+running cost can no longer be verified safely.
 
 Run `npx @openai/codex-security scan --help` or `npx @openai/codex-security bulk-scan --help`
 for the complete CLI references.
--- a/sdk/typescript/src/cost.ts
+++ b/sdk/typescript/src/cost.ts
@@ -28,7 +28,9 @@ interface ScanTokenUsage {
 
 interface SessionUsage {
   offset: number;
-  remainder: Buffer;
+  pendingLine: Buffer[];
+  pendingLineBytes: number;
+  unreadable: boolean;
   threadId: string | null;
   parentThreadId: string | null;
   usage: ScanTokenUsage | null;
@@ -56,6 +58,7 @@ const MODEL_PRICING_NANODOLLARS: Readonly<Record<string, ModelPricing>> = {
 
 const COST_POLL_INTERVAL_MS = 100;
 const SESSION_READ_SIZE = 64 * 1_024;
+const MAX_SESSION_EVENT_BYTES = 1 * 1_024 * 1_024;
 
 export class ScanCostTracker {
   readonly #options: ScanCostTrackerOptions;
@@ -115,7 +118,9 @@ export class ScanCostTracker {
       if (session === undefined) {
         session = {
           offset: 0,
-          remainder: Buffer.alloc(0),
+          pendingLine: [],
+          pendingLineBytes: 0,
+          unreadable: false,
           threadId: null,
           parentThreadId: null,
           usage: null,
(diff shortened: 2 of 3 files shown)


rajpratham1 (approved):
This pull request strengthens the session cost tracker by replacing the unbounded remainder buffer with bounded incremental line buffering and enforcing a 1 MiB maximum session event size. The new implementation preserves partial events across multiple reads while preventing malformed or excessively large events from consuming unbounded memory. When the limit is exceeded, the session is quarantined so subsequent reads do not continue processing an unreadable stream. The accompanying tests verify both successful reconstruction of fragmented events and correct handling of oversized events, and the README documents the new behavior. Based on the visible changes, the implementation is focused, aligns with its stated purpose, and no blocking issues are apparent.

mldangelo-oai:
Thanks for tracking down the incremental cost-event buffering issue. The bounds proposed here are already present on `main`, so this PR no longer adds a distinct change. I'll close it as superseded. We'd welcome another contribution if you find a case the current cost tracking still misses.

PR #154, opened 2026-07-30
fix: report local scan failures without connectivity advice

### LINKED ISSUE #36: Local input and filesystem errors are reported as model-service connectivity or authorization failures
## Summary

`scanFailureMessage` classifies *every* scan failure with `classifyConnectionFailure`, which matches bare words such as `network`, `connection`, `permission denied`, `401` and `403` anywhere in the message — including inside repository paths, Git refs and filesystem error text supplied by the user. When the classification is anything other than `unknown`, the original message is discarded and replaced with connectivity or authorization advice that does not apply.

Classification is applied to the whole failure:

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/cli.ts#L2478-L2505

and the patterns are plain substring matches:

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/api.ts#L1543-L1585

Only the `unknown` branch calls `cliErrorMessage(error)`; every other branch returns fixed advice, so the real cause reaches neither stderr nor the JSON `error` field.

The false-positive class is already known — `classifyConnectionFailure` short-circuits on `sqlite`/`database`/`workbench` text and `tests-ts/api.test.ts` pins that guard — but the guard covers only workbench and database wording, not user-supplied paths, refs, or filesystem errors.

- macOS 26.5.2 (build 25F84)
- Node.js v24.5.0
- Bun 1.3.14

## Steps to reproduce

Driving the real `main()` with the existing `tests-ts/cli-fixtures.ts` harness, where the injected `run()` rejects with the error the SDK would genuinely raise:

```ts
const err = capture();
const exitCode = await main(
  ["scan", ".", "--path", "src/network/client.ts"],
  out.stream,
  err.stream,
  dependencies({
    onRun: () => {
      throw new InvalidTargetError("Path target does not exist: src/network/client.ts");
    },
  }),
);
```

Observed results for three realistic failures:

```text
### missing --path containing the word 'network'
thrown by run():  Path target does not exist: src/network/client.ts
shown to user:    codex-security: The model service could not be reached. Check your network connection and try again.
exit code:        2

### unknown git ref containing the word 'connection'
thrown by run():  unknown Git ref: origin/connection-fix
shown to user:    codex-security: The model service could not be reached. Check your network connection and try again.
exit code:        2

### read-only TMPDIR (EACCES from mkdtemp)
thrown by run():  EACCES: permission denied, mkdtemp '/tmp/openai-codex-security-home-XXXXXX'
shown to user:    codex-security: The stored ChatGPT credentials cannot access the configured model. Use an account or API key with model access.
exit code:        2
```

The exact messages come from `InvalidTargetError` in `src/targets.ts` (`Path target does not exist:` and `unknown Git ref:`) and from `createIsolatedHome`'s `mkdtemp` in `src/runtime.ts`.

## Expected behavior

Connectivity and authorization advice should be produced only for failures that actually originate from the model transport. Local input validation errors and filesystem errors should surface their own message so the user can act on the real cause.

## Actual behavior

A mistyped path, a missing Git ref, or an unwritable temporary directory is reported as a network outage or a credential problem, and the actual error text is dropped entirely.

## Impact

Beyond the immediate confusion, this actively hides the cause during diagnosis. A user following the printed advice will check their network or reconfigure credentials while the real problem is a typo in `--path`.

This may also be worth considering in relation to #26, where a scan failed with *"The model service could not be reached"* while `codex exec` worked from the same shell. Whatever the underlying cause there was, this masking is what prevented the real error from being visible in the report.

The same classifier also gates retry-versus-fatal handling in the event stream (`src/api.ts`, the `error` event branch), so a stream error whose text merely contains "permission denied" is treated as an unrecoverable authorization failure rather than a retryable one.

## Suggested direction

Apply `classifyConnectionFailure` only to errors known to come from the Codex transport, and let local validation and filesystem errors fall through to `cliErrorMessage`. Alternatively, always include the original message alongside the advice so no failure is reported without its cause.

mldangelo:
Fixes #36.

## Scope correction — two of the three reported reproductions are already fixed

I re-ran all three reproductions from the issue against current `main` (`9c7634b`) before writing anything. The issue was filed against `f22d4a36`, and the `network_error`/`timeout` branches of `scanFailureMessage` now already fall through to `cliErrorMessage`:

| Reported reproduction | Current `main` |
|---|---|
| `Path target does not exist: src/network/client.ts` | **already correct** — shows the real message |
| `unknown Git ref: origin/connection-fix` | **already correct** — shows the real message |
| `EACCES: permission denied, mkdtemp '/tmp/...'` | **still broken** — reports a credentials failure |

So this PR addresses the one surviving case, which is the `unauthorized` / `forbidden` / `rate_limited` branches — the three that still discard the original message. The practical trigger is `permission denied`, the standard `strerror` for `EACCES`, which the `forbidden` pattern matches verbatim:

```
/\b403\b|\bforbidden\b|\bpermission denied\b|.../iu
```

A read-only `TMPDIR` therefore surfaces as *"The stored ChatGPT credentials cannot access the configured model. Use an account or API key with model access."* — sending the user to reconfigure credentials for a `chmod` problem, with the real path never printed.

## What this changes

Classification is skipped for failures that cannot originate from the model transport, so they keep their own message:

- `InvalidTargetError` — path and Git ref validation
- `OutputDirectoryError` — output directory checks
- `ConfigurationError` — local config
- `PluginPythonUnavailableError` — interpreter discovery
- filesystem syscall errno codes (`EACCES`, `EPERM`, `ENOSPC`, `EROFS`, `ENOENT`, …)

Network errno codes (`ECONNRESET`, `ENOTFOUND`, `ETIMEDOUT`, …) are **deliberately excluded** from that set — they are genuine transport failures and must keep classifying. The set is an explicit allowlist rather than a `/^E[A-Z]+$/` test for exactly that reason.

## Why I did *not* take the issue's alternative suggestion

The issue offers a second option: *"always include the original message alongside the advice so no failure is reported without its cause."*

I implemented that first, and the existing suite caught it. `tests-ts/cli-authentication.test.ts` → *"identifies overriding API keys in noninteractive scan auth failures"* deliberately asserts that upstream text must **not** reach stderr:

```ts
["403 model access denied for org-private", "cannot access the configured model"],
...
expect(stderr.text()).not.toContain("org-private");
```

Upstream authentication and authorization errors can name the organization or project, and suppressing that is intentional. Appending the cause would leak it into both stderr and the JSON `error` field. So the advice branches still replace the underlying text, and I added a comment recording why, since the next person to read this will have the same idea I did.

That trade-off means a genuinely *misclassified* transport error still loses its text. Narrowing the `forbidden` pattern itself (for instance requiring `permission denied` to co-occur with model/auth context) would close that too, but it changes retry-versus-fatal behavior in the event stream and felt out of scope here.

## Also out of scope

The issue notes the same classifier gates retry-versus-fatal in the `api.ts` event-stream `error` branch. Those messages come from the Codex transport, where classification is the intended behavior, and changing it risks turning fatal auth failures into retry loops. Worth a separate look, not folded in here.

## Testing / QA instructions

```bash
CI=true pnpm run types
CI=true pnpm run test
CI=true pnpm run format
CI=true pnpm run build
```

### Confirm the test reproduces the bug

```bash
cd sdk/typescript
cp src/cli.ts /tmp/fixed.ts
git checkout main -- src/cli.ts
CI=true bun test --timeout 30000 ./tests-ts/cli.test.ts -t "local input and filesystem failures"
#   expect 1 fail, showing:
#   "The stored ChatGPT credentials cannot access the configured model. Use an ..."
cp /tmp/fixed.ts src/cli.ts
CI=true bun test --timeout 30000 ./tests-ts/cli.test.ts -t "local input and filesystem failures"   # 1 pass
```

### New coverage

1. **`reports local input and filesystem failures without connectivity advice`** — drives real `main()` through the `cli-fixtures` harness with six failures: `EACCES` from `mkdtemp`, `EPERM` from `mkdir`, an `OutputDirectoryError`, an `InvalidTargetError` naming a `src/403/` path, an `InvalidTargetError` naming an `origin/forbidden-paths` ref, and a `PluginPythonUnavailableError`. Each asserts the real message reaches stderr and that no credential, authentication, or rate-limit advice appears.

   The `403` and `forbidden` cases are deliberate: they cover the *other* two words in the `forbidden` pattern that a repository path or branch name can plausibly contain, not just `permission denied`.

2. **`keeps model authorization advice for genuine transport failures`** — the guard against over-correcting. A `CodexSecurityError` carrying `401 invalid API key for org-private` / `403 model access denied for org-private` must still produce advice **and** must still not leak `org-private`.

### Regression surface

The pre-existing tests that pin this area are unmodified and still pass — `cli-authentication.test.ts` *"identifies overriding API keys…"* (the `org-private` suppression above) and the three `cli.test.ts` cases asserting sqlite/database/sandbox errors do not claim a network failure.

### DIFF
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -56,7 +56,11 @@ import {
 import { formatUsd } from "./cost.js";
 import {
   CodexSecurityError,
+  ConfigurationError,
+  InvalidTargetError,
+  OutputDirectoryError,
   OutputInsideProtectedRootError,
+  PluginPythonUnavailableError,
   ScanInterruptedError,
 } from "./errors.js";
 import type { SeverityLevel } from "./models.js";
@@ -2674,10 +2678,61 @@ async function runScan(
   return { exitCode: blockingCount > 0 ? 1 : 0, data: result.toJSON() };
 }
 
+// Filesystem and OS syscall failures cannot originate from the model transport,
+// so they must never be rewritten as connectivity or credential advice. Network
+// errno codes are deliberately absent: they are genuine transport failures.
+const LOCAL_SYSCALL_CODES = new Set([
+  "EACCES",
+  "EBUSY",
+  "EEXIST",
+  "EFBIG",
+  "EIO",
+  "EISDIR",
+  "ELOOP",
+  "EMFILE",
+  "ENAMETOOLONG",
+  "ENFILE",
+  "ENOENT",
+  "ENOMEM",
+  "ENOSPC",
+  "ENOTDIR",
+  "ENOTEMPTY",
+  "EPERM",
+  "EROFS",
+  "EXDEV",
+]);
+
+function isLocalScanFailure(error: unknown): boolean {
+  if (
+    error instanceof InvalidTargetError ||
+    error instanceof OutputDirectoryError ||
+    error instanceof ConfigurationError ||
+    error instanceof PluginPythonUnavailableError
+  ) {
+    return true;
+  }
+  return (
+    typeof error === "object" &&
+    error !== null &&
+    "code" in error &&
+    typeof (error as { code: unknown }).code === "string" &&
+    LOCAL_SYSCALL_CODES.has((error as { code: string }).code)
+  );
+}
+
 function scanFailureMessage(
   error: unknown,
   authentication: ScanAuthentication | null,
 ): string {
+  // A local failure keeps its own message. Classification matches bare words
+  // such as "permission denied" anywhere in the text, so an EACCES from a
+  // read-only TMPDIR would otherwise be reported as a credential problem.
+  //
+  // The advice branches below still replace the underlying text rather than
+  // appending it. That is deliberate: upstream authentication and authorization
+  // errors can name the organization or project, which must not reach stderr or
+  // the JSON error field.
+  if (isLocalScanFailure(error)) return cliErrorMessage(error);
   switch (classifyConnectionFailure(error)) {
     case "unauthorized":
       return authentication?.method === "api_key"
--- a/sdk/typescript/tests-ts/cli.test.ts
+++ b/sdk/typescript/tests-ts/cli.test.ts
@@ -25,7 +25,10 @@ import {
   BUNDLED_PLUGIN_VERSION,
   CodexSecurityError,
   DiffTarget,
+  InvalidTargetError,
+  OutputDirectoryError,
   OutputInsideProtectedRootError,
+  PluginPythonUnavailableError,
   ScanCostLimitExceededError,
   ScanInterruptedError,
   VERSION,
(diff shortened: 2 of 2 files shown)


mldangelo-oai:
Thanks for the thoughtful work on making local scan failures easier to understand. The substantive error handling from this branch is already on `main`, so there isn't a reason to carry the older PR forward. There is a small remaining `PluginBootstrapError` edge that we can handle in a separate focused fix. We'd be happy to review future contributions.

PR #155, opened 2026-07-30
fix: record findings held back by the reportable severity gate

### LINKED ISSUE #48: informational findings are exported to SARIF and CSV but dropped from report.md, which then reports no findings
## Summary

`informational` is a valid severity in the findings schema and is exported to SARIF and CSV, but the `report.md` projection filters it out. A scan whose findings are all informational produces a sealed `findings.json` containing them, a SARIF file containing them, and a human-facing report that states there were no findings.

The severity is accepted by the schema ([`findings.schema.json` line 107](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/schemas/findings.schema.json#L107)) and mapped for SARIF export ([`finalize_scan_contract.py` line 51](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py#L51), `"informational": "note"`), but excluded from the report:

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/report_projection.py#L16

```python
REPORTABLE_SEVERITIES = {"critical", "high", "medium", "low"}
```

applied at [line 587](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/report_projection.py#L587). The severity and confidence mixes shown in the Scan Summary are computed from the already-filtered list, so those counts omit informational findings as well.

- Python 3.12.13

## Steps to reproduce

Take the bundled `examples/completed-scan` bundle (one finding) and render the report twice, changing only the severity level:

```python
import report_projection as R
findings["findings"][0]["severity"]["level"] = level
text = R.build_report_markdown(manifest, findings, coverage)
```

Observed:

```text
--- severity='high'
    report.md says 'No findings':   False
    finding title appears:          True

--- severity='informational'
    report.md says 'No findings':   True
    finding title appears:          False
```

## Expected behavior

Either informational findings should appear in `report.md` — perhaps in a separate, clearly de-emphasised section — or the exclusion should be stated somewhere. `references/final-report.md` lists `critical` through `low` when describing ordering, but never says informational findings are dropped.

## Actual behavior

The report claims no reportable findings survived the discovery, validation and reportability gates, while the sealed `findings.json` and the SARIF export both contain them.

## Impact

Modest in severity but awkward in kind: the three outputs of a single scan disagree with each other, and the one a human reads is the one that omits data. Someone reviewing `report.md` concludes the scan found nothing, while the SARIF uploaded to a code-scanning dashboard shows `note`-level results, and `findings.json` records them as real findings.

Whether this is reachable depends on how often a scan produces only informational findings, which I have not measured — on a clean codebase it seems plausible rather than exotic.

## Suggested direction

If the exclusion is deliberate, the clearest fix is to say so in the report itself — for example a line noting that N informational findings were recorded and are available in `findings.json` and the SARIF export — so the three artifacts no longer contradict each other. If it is not deliberate, adding `informational` to `REPORTABLE_SEVERITIES` and giving it its own section would align the report with the other two outputs.

mldangelo:
Fixes #48.

## Summary

`report.md` filters findings to `REPORTABLE_SEVERITIES`, which excludes `informational`. A scan whose findings are all informational seals them into `findings.json`, exports them to SARIF as `note` and to CSV, and then renders a report stating:

> ### No findings
>
> No reportable findings survived the canonical discovery, validation, and reportability gates.

The three artifacts of a single scan contradict each other, and the one a human reads is the one missing data.

## Which of the issue's two directions I took, and why

The issue offers both: state the exclusion, or add `informational` to `REPORTABLE_SEVERITIES` and give it a section.

I concluded the exclusion is **deliberate** and took the first. The evidence is that "reportable" is a load-bearing concept across the codebase, not a one-off in the projection: `src/cli.ts` defines the same four levels as `REPORTABLE_SEVERITIES` to gate `--fail-on-severity` (you cannot fail a build on informational) and then defines a *separate* `DISPLAY_SEVERITIES` that adds informational back for display. The report body follows the reportable set; that split is intentional and I did not want to silently reverse it.

So the report keeps its reportable-only body and instead records what was held back:

> 1 finding is outside the reportable severity set (informational: 1) and not detailed here. It remains recorded in `findings.json` and in the SARIF and CSV exports.

The line appears in both branches — under `### No findings`, and beneath the findings table when reportable findings also exist, since the summary counts are computed from the filtered list and would otherwise under-report there too.

The count is derived from `not in REPORTABLE_SEVERITIES` rather than hardcoding `informational`, and the mix is rendered through the existing `_severity_mix`, so a future non-reportable level is described accurately instead of silently vanishing again.

`references/final-report.md` described the ordering as `critical` → `low` without ever saying informational is dropped — the documentation gap the issue calls out. It now states it.

## Testing / QA instructions

Baseline before this branch: **470 pass / 6 skip / 0 fail**. After: **475 pass / 6 skip / 0 fail** (five added tests, new file `tests-ts/report-projection.test.ts`).

```bash
CI=true pnpm run types
CI=true pnpm run test
CI=true pnpm run format
CI=true pnpm run build
```

### Confirm the tests reproduce the issue

Revert only the plugin files, keeping the tests:

```bash
cd sdk/typescript
cp _bundled_plugin/scripts/report_projection.py /tmp/rp.py
cp _bundled_plugin/references/final-report.md /tmp/fr.md
git checkout main -- _bundled_plugin/scripts/report_projection.py _bundled_plugin/references/final-report.md
CI=true bun test --timeout 30000 ./tests-ts/report-projection.test.ts   # expect 2 pass / 3 fail
cp /tmp/rp.py _bundled_plugin/scripts/report_projection.py
cp /tmp/fr.md _bundled_plugin/references/final-report.md
CI=true bun test --timeout 30000 ./tests-ts/report-projection.test.ts   # expect 5 pass / 0 fail
```

### Reproduce by hand, as the issue does

```bash
cd sdk/typescript/_bundled_plugin && python3 -I -B -c '
import copy, json, sys; sys.path.insert(0, "scripts")
import report_projection as R
load = lambda n: json.load(open("examples/completed-scan/" + n, encoding="utf-8"))
m, c, f = load("scan-manifest.json"), load("coverage.json"), load("findings.json")
for level in ("high", "informational"):
    d = copy.deepcopy(f); d["findings"][0]["severity"]["level"] = level
    t = R.build_report_markdown(m, d, c)
    print(level, "| no-findings:", "### No findings" in t, "| records held back:",
          "outside the reportable severity set" in t)'
```

Use `-B` (as the workbench itself does via `runWorkbench`) — without it Python leaves `__pycache__` inside `_bundled_plugin/scripts/`, which is not gitignored and makes `pnpm run check:package` fail with *"npm tarball contains an unexpected file"*. The added tests spawn Python with `-I -B` for the same reason.

### New coverage — `tests-ts/report-projection.test.ts`

The repo has no Python test suite, so these drive the real projection through `spawnSync`, guarded by `test.skipIf` on interpreter availability (matching the existing `skipIf` convention). They render the bundled `examples/completed-scan` bundle with severities forced, exactly as the issue reproduces it.

1. `records findings held back by the reportable severity gate` — all informational: report still says `### No findings`, and now also states `1 finding is outside the reportable severity set (informational: 1)` and names `findings.json`.
2. `records held-back findings alongside reportable ones` — 1 high + 2 informational: no `### No findings`, the high finding renders, and `2 findings are outside…(informational: 2)` appears.
3. `stays silent when every finding is reportable` — high + low: the note must be absent, so clean reports are unchanged.
4. `keeps the rendered report valid for the format validator` — pipes both renders through `validate_report_format.py` and asserts exit 0, since `report.md` is contract-validated and the new lines sit inside the `## Findings` section.
5. `documents the exclusion in the final-report reference` — pins the documentation fix.

### Packaging check

Because this touches `_bundled_plugin`, verify the plugin payload is unchanged in shape:

```bash
cd sdk/typescript
CI=true pnpm pack --pack-destination /tmp/pkg
CI=true pnpm run check:package /tmp/pkg/openai-codex-security-0.1.4.tgz
# expect: 95 bundled plugin files, 179 entries
```

No files were added or removed from the plugin, so `plugin-files.json` needs no update (it carries paths, not hashes).


mldangelo-oai:
@codex review exact head 345f0eabe05ed092c2aa6e28716b927f8ff8f4df

kmbroai (commented):
## Critical review

Reviewed `345f0eabe05ed092c2aa6e28716b927f8ff8f4df`. **The product change is justified, but the new test has a concrete Windows blocker that remains unresolved.**

### Correctness

**[P1] Replace `/dev/stdin` in the format-validator test.** The [new test](https://github.com/openai/codex-security/blob/345f0eabe05ed092c2aa6e28716b927f8ff8f4df/sdk/typescript/tests-ts/report-projection.test.ts#L123) passes that POSIX device path as `--report-md`. The validator constructs a `Path` and calls `read_text()`; it does not interpret the argument as a portable stdin convention. The [workflow includes Windows](https://github.com/openai/codex-security/blob/345f0eabe05ed092c2aa6e28716b927f8ff8f4df/.github/workflows/node-ci.yml#L23), and interpreter availability does not imply that device exists. On a normal Windows runner with Python available, the test runs and fails before validating the report. The previous review finding is still present at this head.

Write the rendered report to a temporary file, pass its native path, and clean it up in `finally`. That is smaller and more appropriate than changing the production validator's input API just for this test. Keep the Windows test enabled.

Keep the informational-findings note. It explains why sealed findings exist while the human report has no reportable entries, without silently changing the reportability policy. Current main still filters the report to critical/high/medium/low, so the distinction remains relevant. The mixed-severity branch matters as well as the all-informational case.

Counting findings outside `REPORTABLE_SEVERITIES` and reusing `_severity_mix` is adequate; no new severity configuration or reporting mode is needed. I would not make informational findings reportable merely to avoid explaining the distinction.

The new tests are disproportionately tied to wording: the documentation-string assertion and long exact prose checks add maintenance without proving export consistency. Retain the behavioral all-informational, mixed, reportable-only, and real-validator checks. Prefer counts/artifact references and meaningful report structure over pinning the precise explanatory sentence or documentation wording. This cleanup is non-blocking; the device-path fix is not.

### Verification

Ran `bun test --timeout 30000 tests-ts/report-projection.test.ts`: **5 passed, 0 failed** on Bun 1.3.14/Linux. That does not validate the Windows-only failure; that conclusion follows from the checked-in test, the validator's path-opening code, and the Windows matrix. I did not execute a Windows runner.

PR #159, opened 2026-07-30
fix: salvage malformed optional remediation fields during finding recovery

### LINKED ISSUE #106: A malformed optional `remediationTests` discards the whole finding, and no reference documents its shape
## Summary

`remediationTests` is an optional finding field with no documented shape. The schema requires `array<string>`, but no reference, skill, or bundled example shows it — so a producer that emits a plausible alternative (a prose string, or an array of objects) has its **entire finding discarded**, not just the offending field.

The schema:

https://github.com/openai/codex-security/blob/8d9eb30dcf71853d9e55da517404d4aef6c74f64/sdk/typescript/_bundled_plugin/schemas/findings.schema.json#L312

```json
"remediationTests": { "type": "array", "items": { "type": "string", "minLength": 1 } }
```

It is not in the `required` list for a finding, so omitting it is fine — but getting its shape wrong is fatal to the whole finding.

Nothing tells a producer that shape:

- `references/finding-detail-fields.md` carries the canonical **Structured Example** for encoding a finding. `remediationTests` appears in that file **zero times** — the example stops after `attackPath` and omits `remediation`, `remediationTests`, and `preventiveControls` entirely.
- The prose in that same file describes the field only in natural language, which reads as narrative rather than a list of strings: [line 26](https://github.com/openai/codex-security/blob/8d9eb30dcf71853d9e55da517404d4aef6c74f64/sdk/typescript/_bundled_plugin/references/finding-detail-fields.md#L26) — "the minimal remediation invariant, regression tests, and preventive controls."
- The bundled `examples/completed-scan/findings.json` includes `remediation` but **not** `remediationTests`, so the one worked example doesn't demonstrate it either.
- The only other mention is a bare field name inside a comma-separated list: [`references/final-report.md` line 32](https://github.com/openai/codex-security/blob/8d9eb30dcf71853d9e55da517404d4aef6c74f64/sdk/typescript/_bundled_plugin/references/final-report.md#L32).

This is the same shape as #47 — a field whose documented guidance doesn't match what the schema accepts.

## The discard is all-or-nothing, unlike the adjacent `writeup` handling

A malformed `writeup` degrades gracefully — the bad sub-object is dropped and the finding survives:

https://github.com/openai/codex-security/blob/8d9eb30dcf71853d9e55da517404d4aef6c74f64/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py#L855

```python
                except ContractError as exc:
                    finding.pop("writeup")
                    warnings.append(f"Skipped malformed writeup for finding {index + 1}: {exc}.")
```

Four lines later, any other schema error discards the finding outright:

https://github.com/openai/codex-security/blob/8d9eb30dcf71853d9e55da517404d4aef6c74f64/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py#L859

```python
        except ContractError as exc:
            warning = f"Skipped malformed finding {index + 1}: {exc}."
            warnings.append(warning)
            discarded.append(warning)
            continue
```

So a fully-validated finding — complete `summary`, `codeEvidence`, `rootCause`, `validation`, `attackPath`, `severity`, and `remediation` — is thrown away because one *optional* auxiliary field arrived as a string instead of a list of strings.

- Released package: `@openai/codex-security@0.1.3` (bundled plugin 0.1.14)
- Confirmed on current `main` at `8d9eb30dcf71853d9e55da517404d4aef6c74f64`
- macOS 15 (arm64), Node 26, Python 3.14

## Steps to reproduce

Using the plugin's own validator against the bundled example, varying only `remediationTests`:

```python
import json, importlib.util
from pathlib import Path

spec = importlib.util.spec_from_file_location("fsc", "scripts/finalize_scan_contract.py")
m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)

schema = Path("schemas/findings.schema.json")
example = json.load(open("examples/completed-scan/findings.json"))

for value in (["Add a regression test."], "Add a regression test.",
              [{"name": "guard", "description": "Add a regression test."}]):
    payload = json.loads(json.dumps(example))
    payload["findings"][0]["remediationTests"] = value
    try:
        m.validate_against_schema(payload, schema)
        print("ACCEPT", repr(value))
    except Exception as exc:
        print("REJECT", repr(value), "->", exc)
```

Observed:

```text
ACCEPT ['Add a regression test.']
REJECT 'Add a regression test.' -> findings.schema.findings[0].remediationTests: expected schema type array
REJECT [{'name': 'guard', 'description': 'Add a regression test.'}] -> findings.schema.findings[0].remediationTests[0]: expected schema type string
```

## Actual behavior in a real scan

A scoped standard scan of a private TypeScript repository (38 files, `gpt-5.6-terra`, effort `high`) produced three findings that reached `validation.disposition: "reportable"` at `confidence: "high"` with complete attack-path analysis in `artifacts/02_discovery/candidate_ledger.jsonl`.

All three were discarded at finalization:

```json
{"id": "discarded-finding-1", "reason": "Skipped malformed finding 1: findings.findings[0].remediationTests: expected schema type array."}
{"id": "discarded-finding-2", "reason": "Skipped malformed finding 2: findings.findings[1].remediationTests: expected schema type array."}
{"id": "discarded-finding-3", "reason": "Skipped malformed finding 3: findings.findings[2].remediationTests: expected schema type array."}
```

Resulting artifacts:

- `findings.json` — 0 findings
- `exports/results.sarif` — 0 results, 0 rules
- `report.md` Scan Summary — `Reportable findings | 0`

To be fair to the current design, the loss **is** disclosed: the scan exits `2`, stderr says `Scan coverage is partial; results may be incomplete.`, `coverage.completeness` is `partial`, every surface is marked `needs_follow_up`, and `report.md` lists all three discard reasons under Open Questions And Follow Up. That is good failure signalling and it stopped this from passing silently.

The problem is that the analysis itself is unrecoverable from the canonical artifacts. The three findings exist only as free-text summaries in the discovery ledger; they are absent from `findings.json`, absent from SARIF, and absent from the report body. A consumer that uploads SARIF to a code-scanning backend gets zero alerts for a scan that found three high-confidence issues, and the ~$7 of model work that produced them has to be paid again.

## Expected behavior

Two independent fixes, either of which would have preserved the findings:

1. **Degrade gracefully on malformed optional fields**, as the `writeup` path already does — drop `remediationTests` (and similar non-required auxiliary fields), keep the finding, and record the warning. Discarding a whole validated finding over an optional field is disproportionate.
2. **Document the shape.** Add `remediation`, `remediationTests`, and `preventiveControls` to the Structured Example in `references/finding-detail-fields.md`, and include `remediationTests` in `examples/completed-scan/findings.json`. As it stands, the only way to learn that `remediationTests` is `array<string>` is to read the schema.

Rewording [line 26](https://github.com/openai/codex-security/blob/8d9eb30dcf71853d9e55da517404d4aef6c74f64/sdk/typescript/_bundled_plugin/references/finding-detail-fields.md#L26) from "regression tests" to something explicitly plural-list-shaped would help too.

## Impact

Complete, high-confidence findings are silently absent from every machine-readable artifact because of an optional field's shape. Since the discovery ledger is the only surviving record and it isn't part of the sealed contract, a scan can cost full price and deliver an empty `findings.json` and an empty SARIF.

alarcritty:
## Summary

Fixes #106.

A finding whose optional `remediationTests` or `preventiveControls` field arrived in the wrong shape (for example a prose string instead of an array of strings) was discarded entirely during finalization, even when every required field validated. In the reported scan, three high confidence findings were lost this way, producing an empty `findings.json` and an empty SARIF export.

This change recovers those fields the same way malformed writeups are already recovered: the malformed field is dropped, the finding is kept, and a warning is recorded.

### Salvage malformed optional remediation fields

`_recover_unsealed_findings` in `scripts/finalize_scan_contract.py` now validates `remediationTests` and `preventiveControls` individually against their property schemas before the whole-finding schema check. On a `ContractError` the field is popped and a warning is appended:

```text
Skipped malformed remediationTests for finding 1: findings.findings[0].remediationTests: expected schema type array.
```

Because both fields are optional in `findings.schema.json`, the finding then passes the full schema check and survives into `findings.json`, the report, and the SARIF export. The implementation mirrors the existing writeup salvage block a few lines above and uses only existing helpers.

Scope is deliberately limited to these two fields. `validation`, `attackPath`, and `codeEvidence` feed duplicate resolution strength and evidence reference cross checks, so silently dropping them would change semantics rather than trim an annotation.

### Document the expected shape

The schema requires `array` of non-empty strings for both fields, but no reference or example showed that shape:

- `references/finding-detail-fields.md`: the Structured Example previously stopped after `attackPath`; it now includes `remediation`, `remediationTests`, and `preventiveControls`. The prose bullet that read "the minimal remediation invariant, regression tests, and preventive controls" now names each field and its shape explicitly.
- `examples/completed-scan/findings.json`: the bundled example finding now carries both fields with valid values. The recorded `sha256` for `findings.json` in the example `scan-manifest.json` is updated to match, which the sealed contract tests enforce.

## Behavior

- Malformed `remediationTests` or `preventiveControls`: field dropped, finding kept, warning recorded.
- Valid values: preserved unchanged.
- Findings with genuinely broken required fields: still discarded with the same errors as before; the salvage does not mask real validation failures.

## Tests

Added a regression test to `tests-ts/scan-recovery.test.ts` that runs the real `complete-scan` pipeline with the exact shapes from the issue: a valid `["..."]` array (retained), a prose string, an array of objects, and a malformed `preventiveControls` (each dropped with a warning). All five findings survive finalization; previously three of them were discarded.

Also verified the discard reproduces on unmodified `main` with the plugin's own recovery function, using the same inputs, before confirming the fixed behavior.

Verification:

- `pnpm run types`
- `pnpm run test` (471 pass, 0 fail)
- `pnpm run format`
- `pnpm run build`

### DIFF
--- a/sdk/typescript/_bundled_plugin/examples/completed-scan/findings.json
+++ b/sdk/typescript/_bundled_plugin/examples/completed-scan/findings.json
@@ -42,6 +42,12 @@
       "remediation": "Normalize destinations and reject entries that escape the extraction root.",
       "validation": null,
       "attackPath": null,
+      "remediationTests": [
+        "Assert that extracting an archive entry named `../escape.txt` fails without writing outside the extraction root."
+      ],
+      "preventiveControls": [
+        "Route all archive extraction through one helper that normalizes and validates destination paths."
+      ],
       "provenance": {
         "source": "local_plugin"
       },
--- a/sdk/typescript/_bundled_plugin/examples/completed-scan/scan-manifest.json
+++ b/sdk/typescript/_bundled_plugin/examples/completed-scan/scan-manifest.json
@@ -30,7 +30,7 @@
     "artifacts": [
       {
         "path": "findings.json",
-        "sha256": "db5ce8533c1b6cd9131f9e12e8bb705f63474caa2a3f7dd21207666b3a8da777",
+        "sha256": "a6dc4521d6478828224fbafc33585401a47bfeb0e56e07b5d2886e8a29937f2f",
         "mediaType": "application/json"
       },
       {
--- a/sdk/typescript/_bundled_plugin/references/finding-detail-fields.md
+++ b/sdk/typescript/_bundled_plugin/references/finding-detail-fields.md
@@ -23,7 +23,7 @@ The finding detail view is a decision-focused projection of the canonical findin
 - dataflow source, meaningful transformations, dangerous sink, and concrete outcome;
 - realistic attacker, entry point, access requirements, preconditions, and attacker outcome;
 - severity rationale plus the specific evidence that would raise or lower the rating;
-- the minimal remediation invariant, regression tests, and preventive controls.
+- the minimal remediation invariant (`remediation`, a single string), plus `remediationTests` and `preventiveControls`, each an array of short strings with one regression test or preventive control per entry.
 
 Keep background exposition, alternate exploit research, full PoC instructions, representative command output, and long source walkthroughs in the detailed write-up. Do not copy them into canonical fields merely to make the workspace report longer. The workspace should stay self-contained enough to support triage while avoiding duplicated or speculative prose.
 
@@ -155,7 +155,15 @@ The following shape shows how to encode the `environment/add` reserved-environme
     "limitations": [
       "This overwrite does not directly execute code on the victim host."
     ]
-  }
+  },
+  "remediation": "Reuse the startup reserved-ID check inside `EnvironmentManager::upsert_environment()` so the runtime mutation path rejects the reserved `local` identifier.",
+  "remediationTests": [
+    "Assert that `environment/add` with `environmentId: \"local\"` returns a protocol error.",
+    "Assert that `default_environment()` still resolves the manager-owned local runtime after a rejected upsert."
+  ],
+  "preventiveControls": [
+    "Centralize reserved-identifier validation so every environment mutation path shares one guard."
+  ]
 }
 ```
 
--- a/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py
+++ b/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py
@@ -796,6 +796,12 @@ def _recover_unsealed_findings(
     writeup_schema = _require_dict(
         finding_properties, "writeup", "findings.schema.properties.findings.items.properties"
     )
+    auxiliary_schemas = {
+        name: _require_dict(
+            finding_properties, name, "findings.schema.properties.findings.items.properties"
+        )
+        for name in ("remediationTests", "preventiveControls")
+    }
     scan = _require_dict(manifest, "scan", "manifest")
     scan_id = _require_str(scan, "id", "manifest.scan")
     if findings.get("scanId") != scan_id:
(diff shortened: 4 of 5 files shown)


ianw-oai (approved):
Verified the contributor regression preserves valid findings, drops malformed optional guidance, and emits the expected warnings; recovery suite passes.

PR #160, opened 2026-07-30
feat(container): harden sandbox isolation and socket egress

MauricioPerera:
## Summary

Three sandbox-hardening fixes found during a security audit of this repo,
all verified with real `docker build` + `docker run` (not just read
locally) under the exact `compose.yaml` security contract:

1. **Remove `ptrace`/`process_vm_readv`/`process_vm_writev` from
   `docker/codex-security-seccomp.json`.** All sandboxed processes share
   the same uid (10001), so `ptrace` let one process inspect/modify
   another's memory inside the sandbox -- inter-process isolation was
   effectively void despite the profile's intent.

2. **Replace the 3 `socket` domain rules with an explicit allowlist of
   `AF_UNIX`(1)/`AF_INET`(2)/`AF_INET6`(10).** The previous rules
   (`domain<38`, `domain==39`, `domain>40`) only actually blocked
   `AF_ALG`(38) and `AF_VSOCK`(40) -- `AF_NETLINK`(16) and `AF_PACKET`(17)
   were allowed, along with everything else outside that narrow gap,
   making the egress control close to a no-op.

3. **Fail closed in `docker/entrypoint.sh` if not launched with
   `--cap-drop ALL --security-opt no-new-privileges`.** The seccomp
   profile unconditionally allows `mount`/`pivot_root`/`unshare`/`setns`/
   `clone` for Codex's own unprivileged Bubblewrap sandboxing -- those
   syscalls can't be removed without breaking that feature. Their safety
   depended entirely on a comment asserting the caller applies the
   hardened contract (true for `compose.yaml`, not enforced for a bare
   `docker run`). Now the entrypoint reads `/proc/self/status` and
   refuses to start (exit 3, clear message) unless effective capabilities
   are empty and `NoNewPrivs` is set.

## Verification (real Docker, not narrated)

Built the image from this branch and ran it under
`--cap-drop ALL --security-opt no-new-privileges --security-opt seccomp=...`:

- `--help` / `--version`: still work (exit 0).
- DNS resolution (`getaddrinfo`) and `git clone` over HTTPS: still work
  -- confirms the socket fix doesn't break the scanner's own legitimate
  network use.
- `AF_NETLINK` / `AF_PACKET` `socket()`: now fail with `EPERM` (previously
  succeeded).
- `ptrace(PTRACE_TRACEME, ...)`: fails with `EPERM`.
- Running the same image **without** `--cap-drop ALL`/`--security-opt
  no-new-privileges`: entrypoint refuses to start (exit 3) instead of
  silently running under a weaker contract than the seccomp profile
  assumes. Tested both the no-flags case and `--user root` (capabilities
  check fires correctly with the real non-empty bitmask).

## Test plan

- [ ] CI green
- [ ] Maintainer confirms no other seccomp consumer relies on the
      previously-open socket domains (e.g. `AF_NETLINK` for some
      diagnostic path I didn't find)


pasmud (commented):
One socket-domain exception appears to remain: this removes the unfilterable socketcall path, but socketpair stays in the unconditional syscall list. Linux also implements socketpair for AF_TIPC (the current kernel wires tipc_socketpair into the TIPC protocol operations: https://github.com/torvalds/linux/blob/2d2338c93da79b3bfe4b6099a931d9468d539952/net/tipc/socket.c#L3335-L3405), so it is not limited to AF_UNIX as the commit message says. This is not a general egress bypass—the returned pair is locally connected, but it still exposes the TIPC socket path despite the stated domain-minimization goal. Since socketpair receives the domain directly as argument 0, would you remove it from the generic allow list and add an AF_UNIX (value 1) rule instead? A negative AF_TIPC socketpair probe would also make a useful regression check.

mldangelo-oai:
@codex review exact head 6648d5ce765772e5f235e15687285be103c2ef6b

PR #165, opened 2026-07-30
fix: make historical GitHub release backfills idempotent

mldangelo-oai:
## Summary
- Reconcile verified releases whose tagged commit predates .github/release.yml without requesting a nonexistent historical configuration file.
- Preserve configured release notes for newer tags and cover the historical replay case with a focused regression test.
- Fix the live idempotency failure: https://github.com/openai/codex-security/actions/runs/30591383965

## Validation
- Reproduced the historical replay regression before applying the fix.
- Release-automation suite: 165 passing tests.
- TypeScript no-emit check and Prettier both pass.
- Independently verified the published npm package, SLSA provenance, protected publishing run, and byte-identical GitHub release asset.


mldangelo-oai:
@codex review

tiffanycitra (approved):

PR #172, opened 2026-07-31
fix(auth): reject unsupported scan authentication modes

dipeshbabu:
## Summary

- reject unsupported runtime values for the SDK scan authentication mode
- preserve the existing behavior for `auto`, `chatgpt`, and `api-key`
- add regression coverage for both `preflight` and `run`

## Why

JavaScript callers can bypass the TypeScript union type. Previously, any unknown authentication string silently behaved like `auto`, which could select ambient credentials instead of reporting invalid input.

The shared authentication-selection boundary now raises a `TypeError` before runtime initialization when the value is unsupported.

## Impact

Invalid SDK input fails early and predictably. Valid authentication modes are unchanged.

## Validation

- `pnpm dlx bun test --timeout 30000 ./tests-ts/api.test.ts` — 71 passed
- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check src/api.ts tests-ts/api.test.ts`
- `git diff --check`


mldangelo-oai:
@codex review Please review exact head afdfbb168aec80cfe1083aa3b7b5ad1553545c24, focusing on runtime rejection of unsupported authentication modes, preserving supported credentials, initialization ordering, and compatibility with the current-main merge.

mldangelo-oai:
@codex review exact head 8abca1840b96452c14621de82b45d94bb0e0a510

mldangelo-oai (approved):
Thanks for the focused fix. This closes a real JavaScript SDK gap by rejecting invalid authentication modes instead of silently falling back to ambient credentials, and the regression confirms both preflight and run fail before runtime initialization. I verified the exact head locally, and hosted CI plus the exact-head Codex review are green. Nicely scoped improvement!

PR #173, opened 2026-07-31
fix(version): compare SemVer prereleases correctly

dipeshbabu:
## Summary

- compare prerelease versions identifier by identifier according to SemVer precedence
- compare numeric identifiers numerically and below nonnumeric identifiers
- account for differing prerelease identifier counts and ASCII ordering
- add the standard SemVer precedence sequence and the previously misordered `alpha.1`/`alpha-1` case as regression coverage

## Root cause

The update notifier compared each complete prerelease string with `Intl.Collator`. Locale-aware comparison does not implement SemVer's identifier boundaries or precedence rules, so valid versions such as `1.0.0-alpha-1` and `1.0.0-alpha.1` could be ordered incorrectly.

## Impact

Prerelease users receive update notices only when the registry version has higher SemVer precedence. Stable-version behavior is unchanged.

## Overlap check

Before publishing, open issues and pull requests were reviewed and repository-wide searches were run for `semver`, `prerelease`, `update notice`, and `version comparison`. No active issue or pull request overlaps this fix. The only related search hits were already-merged work that introduced the broader update-notice and release features.

## Validation

- `pnpm dlx bun test --timeout 30000 ./tests-ts/update-notice.test.ts` — 10 passed
- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check src/version.ts tests-ts/update-notice.test.ts`
- `git diff --check`

### DIFF
--- a/sdk/typescript/src/version.ts
+++ b/sdk/typescript/src/version.ts
@@ -12,7 +12,7 @@ export const BUNDLED_PLUGIN_VERSION = "0.1.22" as const;
 
 const PACKAGE_NAME = "@openai/codex-security";
 const VERSION_PATTERN =
-  /^(\d+)\.(\d+)\.(\d+)(?:-([0-9A-Za-z.-]+))?(?:\+[0-9A-Za-z.-]+)?$/u;
+  /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-([0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*))?(?:\+[0-9A-Za-z-]+(?:\.[0-9A-Za-z-]+)*)?$/u;
 
 export interface UpdateNotice {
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
@codex review Please review exact head 9d2726817f6714f8c5500d4c403014a5f1dc82c9, focusing on SemVer prerelease precedence, numeric versus lexical identifiers, oversized identifiers, stable releases, and compatibility with the current-main merge.

mldangelo-oai:
@codex review Please review exact contributor-updated head dd681ccdfb1ab4f4952b64054dc4b472850d33ad, focusing on strict SemVer syntax, leading-zero numeric identifiers, empty prerelease or build segments, and overflow-safe numeric precedence.

mldangelo-oai:
@codex review exact head bb1c3f2721fba74c61e61c43de869e3c1c02fb57

PR #177, opened 2026-07-31
fix(cost): coalesce overlapping cost refreshes

### LINKED ISSUE #31: maxCostUsd polling queues full session-tree rescans and can delay budget enforcement
## Summary

When `maxCostUsd` is enabled, `ScanCostTracker` starts a 100 ms interval. Each tick calls `refresh()`, and `refresh()` chains a new full session-tree pass onto `#pending` even when the previous pass has not completed.

Each pass recursively enumerates `$CODEX_HOME/sessions` and opens every discovered `.jsonl` file. On a Codex home with substantial history, polling can therefore build a queue of full traversals. This delays both shutdown and the `onCost` callback that triggers the cost-limit abort.

- Originally reproduced at commit `e94d6bef9797a192febfde89a26ec7f831bc09b2`
- Node.js 24.18.0

## Steps to reproduce

1. Create a temporary Codex home.
2. Under `sessions/2026/07/29`, create 3,000 empty `.jsonl` files. Empty files isolate directory enumeration and file-open overhead from JSON parsing.
3. Construct a tracker:

~~~ts
  codexHome: temporaryHome,
  model: "gpt-5.6-sol",
  maxCostUsd: 1,
});
~~~

4. Call `tracker.start("target-thread")`.
5. Wait 350 ms so multiple interval ticks occur.
6. Measure how long `await tracker.stop()` takes.

Observed on my machine:

~~~json
{
  "historicalSessionFiles": 3000,
  "elapsedBeforeStopMs": 349,
  "stopWaitMs": 1260,
  "totalMs": 1609
}
~~~

Exact timings will vary by filesystem and hardware. The deterministic behavior is that ticks arriving while a refresh is running are appended to `#pending`, and every queued refresh repeats the recursive traversal.

## Expected behavior

At most one session refresh should be active or pending. Timer ticks that occur during a running refresh should be coalesced rather than queued as additional full traversals.

Cost updates and the cost-limit abort should not fall progressively behind because the Codex home contains unrelated historical sessions.

## Actual behavior

Every timer tick schedules another traversal. `stop()` waits for the queued traversals, and `onCost` cannot report usage discovered by a delayed pass until the earlier passes finish.

## Impact

- Cost-limit enforcement can be delayed relative to the active scan.
- A completed or aborted scan can spend additional time waiting for queued polling work.
- The overhead grows with unrelated historical session files, not just the current scan and its child threads.

## Relevant code

- Poll interval and refresh queue: https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/cost.ts#L57-L104
- Full session-tree traversal: https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/cost.ts#L109-L198
- Cost-limit abort callback: https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/api.ts#L503-L523

## Suggested direction

A minimal change could ensure only one refresh is in flight and coalesce subsequent timer ticks into at most one follow-up refresh.

A more complete optimization could also:

- cache directory/file discovery instead of walking the full tree every 100 ms;
- avoid reopening unchanged unrelated sessions;
- narrow candidates to the target thread and discovered child threads where possible.

A regression test could inject a slow session reader or traversal and assert that repeated timer ticks do not cause multiple queued full passes.

## Related issue

This is related to cost control, but it is not a duplicate of #25. That issue asks how to estimate cost before a run; this report concerns runtime polling backlog and delayed enforcement of an already configured limit.

SiavashShams:
Fixes #31.

## Summary

`ScanCostTracker.refresh()` appended every request to a promise chain. When a session-tree traversal exceeded the 100 ms polling interval, timer ticks accumulated full traversals and `stop()` queued more work behind them.

Track one active refresh and one shared follow-up refresh instead. Calls arriving during an active traversal await the same follow-up, including `stop()`, while successful and failed traversals both advance or reset scheduler state.

The regression blocks session reads, starts three overlapping refreshes plus `stop()`, and verifies that only two traversals run. On the baseline it failed after starting a third traversal. A separate case verifies recovery after a traversal error.

## Verification

- `pnpm dlx bun@1.3.14 test --timeout 30000 ./tests-ts/cost.test.ts` — baseline: 14 passed, 1 failed; patched: 15 passed, 0 failed.
- `pnpm dlx bun@1.3.14 test --timeout 30000 ./tests-ts` — 708 passed, 5 expected skips, 0 failed.
- `pnpm run types` — passed.
- `pnpm run format` — passed.
- `pnpm run build` — passed.
- `pnpm pack --pack-destination /private/tmp/codex-security-pack` — passed.
- `npm_config_cache=/private/tmp/codex-security-npm-cache pnpm run check:package /private/tmp/codex-security-pack/openai-codex-security-0.1.4.tgz` — validated the public import, CLI, bundled plugin files, and archive contents.
- `git diff --check` — passed.

### DIFF
--- a/sdk/typescript/src/cost.ts
+++ b/sdk/typescript/src/cost.ts
@@ -49,6 +49,8 @@ interface ScanCostSnapshot {
   cost: ScanCost | null;
 }
 
+type ScanCostRefreshTask = () => Promise<void>;
+
 const MODEL_PRICING_NANODOLLARS: Readonly<Record<string, ModelPricing>> = {
   "gpt-5.6": [5_000, 500, 6_250, 30_000],
   "gpt-5.6-sol": [5_000, 500, 6_250, 30_000],
@@ -62,15 +64,21 @@ const MAX_SESSION_EVENT_BYTES = 1 * 1_024 * 1_024;
 
 export class ScanCostTracker {
   readonly #options: ScanCostTrackerOptions;
+  readonly #refreshTask: ScanCostRefreshTask;
   readonly #sessions = new Map<string, SessionUsage>();
   #threadId: string | null = null;
   #timer: NodeJS.Timeout | null = null;
-  #pending: Promise<void> = Promise.resolve();
+  #activeRefresh: Promise<void> | null = null;
+  #queuedRefresh: Promise<void> | null = null;
   #snapshot: ScanCostSnapshot = { usage: null, cost: null };
   #lastCost: number | null = null;
 
-  public constructor(options: ScanCostTrackerOptions) {
+  public constructor(
+    options: ScanCostTrackerOptions,
+    refreshTask?: ScanCostRefreshTask,
+  ) {
     this.#options = options;
+    this.#refreshTask = refreshTask ?? (() => this.#readSessions());
   }
 
   public start(threadId: string): void {
@@ -88,10 +96,12 @@ export class ScanCostTracker {
   }
 
   public async refresh(): Promise<ScanCostSnapshot> {
-    const update = this.#pending.then(async () => {
-      await this.#readSessions();
-    });
-    this.#pending = update.catch(() => {});
+    const update =
+      this.#activeRefresh === null
+        ? this.#startRefresh()
+        : (this.#queuedRefresh ??= this.#activeRefresh
+            .catch(() => {})
+            .then(this.#refreshTask));
     await update;
     return this.#snapshot;
   }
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thank you for the original cost-polling fix. The follow-up in #198 includes your exact contributor-authored commit and narrows the change to interval polling while preserving the existing tracker API; it is the preferred working branch for #31. #198 is still a draft and requires the normal review process, so this PR is being kept open for context and attribution.

mldangelo-oai:
Linking #326 because both PRs touch `ScanCostTracker`. #326 fixes scan-aborting behavior for oversized session events while keeping memory bounded; the refresh-coalescing work here addresses a different problem and is complementary. A rebase should preserve both behaviors.

PR #182, opened 2026-07-31
fix: handle missing release tags safely

mldangelo-oai:
## Summary

- Treat GitHub's JSON `404` response for a missing release tag as absent so the release cutter can create the exact merged tag.
- Fail closed on permission failures, unavailable responses, and other tag lookup errors instead of attempting publication.
- Add regression coverage for missing tags, forbidden lookups, and missing API responses while retaining existing annotated/lightweight tag checks.

The `0.1.5` release exposed this bug in `node-release-cut`: `gh api` writes its `404` error body to stdout, so the previous `|| true` path mistakenly treated the JSON error as an existing Git ref.

## Validation

- `bun test --timeout 30000 ./tests-ts` — **709 passed, 5 existing skips**.
- `tsc --noEmit -p sdk/typescript/tsconfig.json`.
- `prettier --check .github/workflows/node-release-cut.yml sdk/typescript/tests-ts/release-automation.test.ts`.
- `git diff --check`.


mldangelo-oai:
@codex review Please review exact head 4bf379c56e6173d3c324f2ab591b9f7fb70c2d93, focusing on real GitHub CLI 404 responses for missing release tags, fail-closed handling of 403/transport errors, exact-commit tag validation, and release automation regression coverage.

mldangelo-oai:
@codex review Please re-review the updated exact head 7e84eca99f08325cf8a2d76aaeec2de6128eba9e after your P1 finding. Tag lookup now reads the unfiltered HTTP response and status before parsing JSON; please verify missing tags are created while 403s, malformed 404s, transport failures, and retargeted tags still fail closed.

zcrab-oai (approved):

ianw-oai (approved):

PR #185, opened 2026-07-31
fix(cli): display scan paths with trailing separators

dipeshbabu:
## Summary

- preserve the final directory name when an absolute scan path ends in a separator
- keep parent path components out of progress output
- add focused CLI regression coverage

## Root cause

The scan-scope formatter split absolute paths on `/` and selected the final segment. A trailing slash or backslash produced an empty final segment, so progress output omitted the scoped directory name.

## Impact

Scans invoked with paths such as `/workspace/service/` now report `Running scan: service` while retaining the existing parent-path redaction behavior.

## Overlap check

Searched open issues and pull requests for trailing-separator scan paths, scan progress path display, and the affected helper name. The current open PRs are unrelated.

## Validation

- `pnpm dlx bun test --timeout 30000 ./tests-ts/cli.test.ts` (85 passed)
- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check src/cli.ts tests-ts/cli.test.ts`
- `git diff --cached --check`


mldangelo-oai:
@codex review exact head be676ec0a751ba36b9438edd59f995338ddd4f03

mldangelo-oai (approved):
Thanks for the focused fix. This preserves a useful directory label for trailing-separator paths while keeping parent components out of progress output, and the regression test covers both behaviors. I verified the exact head locally with the CLI test suite, TypeScript, and formatting; hosted CI and the exact-head Codex review are also green. Nice improvement!

PR #189, opened 2026-08-01
Accept RFC 3339 leap seconds

dipeshbabu:
## Summary

- accept RFC 3339 leap-second timestamps with a seconds value of `60`
- add contract coverage using the 2016-12-31 leap second

## Root cause

The custom RFC 3339 validator capped seconds at `59`, even though RFC 3339 permits `60` for an inserted leap second.

## Impact

Canonical scan contracts containing a standards-compliant leap-second timestamp are no longer rejected as malformed.

## Overlap check

Searched open issues and pull requests for leap seconds, RFC 3339 second `60`, and the affected validator name. The newest open pull requests are unrelated.

## Validation

- `pnpm dlx bun test --timeout 30000 ./tests-ts/contract.test.ts` (26 passed, 4 platform skips)
- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check src/contract.ts tests-ts/contract.test.ts`
- `git diff --check`

### DIFF
--- a/sdk/typescript/src/contract.ts
+++ b/sdk/typescript/src/contract.ts
@@ -1192,7 +1192,7 @@ function validRfc3339DateTime(value: string): boolean {
     day < 1 ||
     hour > 23 ||
     minute > 59 ||
-    second > 59 ||
+    second > 60 ||
     offsetHour > 23 ||
     offsetMinute > 59
   ) {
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thanks for taking the time to look at timestamp validation and RFC 3339. We do not have a scan producer that emits leap seconds, and accepting them only in TypeScript would leave the Python finalizer and JavaScript date handling inconsistent. Because this would introduce invalid or unreadable timestamps without fixing a reported use case, I'm going to close the PR. We'd welcome a future contribution that addresses a concrete timestamp or interoperability issue.

PR #198, opened 2026-08-01
fix: coalesce cost polling without changing tracker API

### LINKED ISSUE #31: maxCostUsd polling queues full session-tree rescans and can delay budget enforcement

mldangelo-oai:
## Summary

Fixes #31. Builds directly on #177 and preserves @SiavashShams's original contributor-authored commit.

- Coalesce overlapping cost-polling interval ticks inside `ScanCostTracker.start()` instead of replacing the existing refresh serializer.
- Preserve the existing one-argument tracker constructor and remove the test-only refresh-task injection added by #177.
- Run at most one polling traversal at a time, remember whether any interval ticks arrived during that traversal, and immediately perform one follow-up while tracking remains active.
- Keep `stop()` authoritative: stopping clears the timer, suppresses any stale follow-up, and serializes exactly one final session refresh behind the in-flight poll.
- Deliver a failed poll to `onError` exactly once and allow the coalesced follow-up to recover normally.
- Exercise actual 100 ms timer ticks, real session JSONL files, bounded shutdown, failed traversal recovery, and deduplicated error reporting without widening production APIs for tests.

## Why

Each budgeted scan polls the Codex session tree every 100 ms. On `main`, every tick appends another full traversal to the refresh promise chain when filesystem work takes longer than the interval. The result is an unbounded backlog and unnecessarily long scan shutdown.

The existing refresh serializer already provides the ordering required by explicit refreshes and `stop()`. Coalescing only the interval producer fixes the actual source of the backlog while retaining those established semantics and avoiding a second scheduler.

With 3,000 real session files and no injected filesystem latency, a representative local run produced:

| Implementation | Session traversals | Session files opened | Shutdown wait |
| --- | ---: | ---: | ---: |
| Current `main` | 5 | 15,000 | 1,577 ms |
| This PR | 3 | 9,000 | 363 ms |

The new polling regression fails against unmodified `main`: four overlapping polling requests are observed where the fixed implementation permits one.

## Validation

- Full pinned Bun 1.3.14 suite: **717 passed, 5 expected platform/integration skips, 0 failed** across 32 files.
- Focused cost-tracker suite: **15 passed**.
- Focused cost-tracker compatibility checks on Node.js **22, 24, and 26**.
- `pnpm run types`.
- `pnpm run format`.
- `pnpm run build`.
- `pnpm pack` and complete package-contract validation: **179 verified entries**.
- Built SDK import/cost-estimation smoke and CLI version smoke.
- `git diff --check`.

This change addresses refresh backlog and shutdown latency. It does not claim to eliminate the separate cost of recursively discovering historical session files.

### DIFF
--- a/sdk/typescript/src/cost.ts
+++ b/sdk/typescript/src/cost.ts
@@ -49,6 +49,8 @@ interface ScanCostSnapshot {
   cost: ScanCost | null;
 }
 
+type ScanCostRefreshTask = () => Promise<void>;
+
 const MODEL_PRICING_NANODOLLARS: Readonly<Record<string, ModelPricing>> = {
   "gpt-5.6": [5_000, 500, 6_250, 30_000],
   "gpt-5.6-sol": [5_000, 500, 6_250, 30_000],
@@ -62,15 +64,21 @@ const MAX_SESSION_EVENT_BYTES = 1 * 1_024 * 1_024;
 
 export class ScanCostTracker {
   readonly #options: ScanCostTrackerOptions;
+  readonly #refreshTask: ScanCostRefreshTask;
   readonly #sessions = new Map<string, SessionUsage>();
   #threadId: string | null = null;
   #timer: NodeJS.Timeout | null = null;
-  #pending: Promise<void> = Promise.resolve();
+  #activeRefresh: Promise<void> | null = null;
+  #queuedRefresh: Promise<void> | null = null;
   #snapshot: ScanCostSnapshot = { usage: null, cost: null };
   #lastCost: number | null = null;
 
-  public constructor(options: ScanCostTrackerOptions) {
+  public constructor(
+    options: ScanCostTrackerOptions,
+    refreshTask?: ScanCostRefreshTask,
+  ) {
     this.#options = options;
+    this.#refreshTask = refreshTask ?? (() => this.#readSessions());
   }
 
   public start(threadId: string): void {
@@ -88,10 +96,12 @@ export class ScanCostTracker {
   }
 
   public async refresh(): Promise<ScanCostSnapshot> {
-    const update = this.#pending.then(async () => {
-      await this.#readSessions();
-    });
-    this.#pending = update.catch(() => {});
+    const update =
+      this.#activeRefresh === null
+        ? this.#startRefresh()
+        : (this.#queuedRefresh ??= this.#activeRefresh
+            .catch(() => {})
+            .then(this.#refreshTask));
     await update;
     return this.#snapshot;
   }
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
@codex review

Please review the exact current head `9c6a5889387d5a8490dae25a481a8271eabb8727`, with particular attention to overlapping polling ticks, failed refresh recovery, deduplicated error reporting, shutdown ordering, and cost-limit enforcement.

mldangelo-oai:
@codex review Please review exact head 77a3149f4e138098afbef0cbd770093f606f3468, focusing on bounded overlapping cost polls, failure recovery, stop/final-refresh races, preserving the public tracker API, and retaining the original contributor commit.

ianw-oai (approved):

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review exact head df438d9144027fc78df2fbb85f360c19e2d998ed

Issue #192, opened 2026-08-01
Windows: valid Git worktree scans fail for non-ASCII target paths


pradigmaz:
## Environment

- Windows
- Codex Security plugin: 0.1.11
- Security workspace UI: 0.1.63
- Python default locale encoding: `cp1251`
- Target: `<Windows path containing Cyrillic characters>`

## Reproduction

1. Use a valid Git worktree whose path contains non-ASCII characters.
2. Start a standard scan for the worktree.

Workspace setup reports:

```json
{
  "isGit": true,
  "isWorktree": true,
  "hasHead": true
}
```

## Observed

The scan fails with:

```text
Scan target must stay inside its Git working tree.
```

`git rev-parse --show-toplevel` returns the correct UTF-8 path, but Python decodes the output using the Windows locale. The resulting path is corrupted, so `Path.relative_to()` raises `ValueError`.

Setting `PYTHONUTF8=1` makes the helper pass, confirming the decoding issue.

## Validation

The issue was reproduced and verified with temporary Git repositories using root and nested targets. Tested path components included:

- Cyrillic
- Simplified and Traditional Chinese
- Japanese
- Arabic
- Korean
- Thai
- Lao
- Khmer
- Burmese
- Georgian
- Armenian
- Indic scripts
- Persian and Urdu
- Amharic
- Mongolian
- Tibetan
- Ukrainian and Serbian
- Latin diacritics
- Combining characters
- Emoji

After decoding text-mode Git output as UTF-8, all cases passed and returned the correct pathspec. Binary Git output remained unchanged.

## Suggested fix

Set an explicit encoding for text-mode Git subprocess calls:

```py
encoding="utf-8" if text else None,
```

The scan-start waiter also remains pending after this startup error. A retry returns:

```text
A scan-start waiter already exists
```

ppcvote:
Opened #293 for this.

Two things from reproducing it on a Traditional Chinese host (cp950) that may be useful here, since the symptom did not match your report at first.

[...]

PR #199, opened 2026-08-01
fix(git): preserve whitespace in worktree paths

dipeshbabu:
## Summary

- preserve legitimate trailing whitespace in Git worktree paths
- remove only Git's final line terminator from command output instead of trimming the entire value
- add regression coverage for a POSIX worktree whose directory name ends in a space

## Root cause

The shared Git output helper called `stdout.trim()`. For `git rev-parse --show-toplevel`, that also removes whitespace belonging to the path itself. A valid POSIX worktree such as `/workspace/repository ` was therefore resolved as `/workspace/repository`, causing worktree-root discovery and diff-target validation to fail.

## Impact

Git-backed scans now preserve valid worktree path whitespace while continuing to remove the CRLF or LF record terminator emitted by Git. Commit and ref resolution behavior is unchanged.

## Overlap check

Open issues and pull requests were searched for Git worktree whitespace, trailing-space paths, and the affected output handling. No active issue or pull request covers this bug. PR #143 also touches `targets.ts`, but only the separate home-directory expansion helper; its hunk and behavior do not overlap this change. Issue #192 concerns non-ASCII subprocess decoding rather than whitespace removal.

## Validation

- `pnpm dlx bun test --timeout 30000 ./tests-ts/targets.test.ts` — 13 passed, 1 Windows-only skip
- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check src/targets.ts tests-ts/targets.test.ts`
- `git diff --cached --check`


mldangelo-oai:
@codex review Please review exact head 175a5967a905e80c1281c29f2aa99e5fa4e7b27c, focusing on trailing whitespace in Git worktree paths, trimming only command-output newlines, ref validation, and cross-platform behavior.

mldangelo-oai:
@codex review Please review exact contributor-updated head 8ee318758a61343fe820e705f8654c4f6b53416c, focusing on preserving POSIX Git worktree paths ending in whitespace or carriage returns and platform-specific Windows newline handling.

mldangelo-oai:
@codex review exact head f9d8a7fea3a719d3182ca5b95204e786aa28586c

mldangelo-oai (approved):
Thanks for the careful follow-through here. The platform-specific terminator handling is small and precise, and the regression coverage for both trailing spaces and carriage returns makes the intended behavior clear. I revalidated this exact head against current main; the focused regression, typecheck, and formatting checks pass locally. Approved.

PR #203, opened 2026-08-01
fix: honor the requested output format when rerunning a scan

### LINKED ISSUE #39: scans rerun ignores --json/--format and can block on the interactive credential prompt
## Summary

`scans rerun` calls `runScan` without forwarding the output format, so it always runs with `interactive = true`. With `--json` or `--format jsonl` it can therefore block on the interactive credential-selection prompt that the `scan` command explicitly suppresses in the same situation.

The rerun path omits the fourth argument, taking the `interactive = true` default:

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/cli.ts#L730

while `scan` computes it from the requested format:

https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/src/cli.ts#L989

`interactive` is what gates the account-selection prompt and the TTY progress rendering.

- macOS 26.5.2 (build 25F84)
- Node.js v24.5.0
- Bun 1.3.14

## Steps to reproduce

Preconditions matching the prompt's guard: stderr is a TTY, an environment API key is set, a stored ChatGPT sign-in exists, and no explicit `--auth` was passed.

Driving the real `main()` with the existing `tests-ts/cli-fixtures.ts` harness and an injected `scanAuthenticationPrompt` that records whether it was called:

```ts
const deps = {
  ...dependencies({ environment: { OPENAI_API_KEY: "SYNTHETIC_NOT_A_REAL_KEY" }, onWorkbench: () => ({ recipe }) }),
  hasStoredChatGPTSignIn: async () => true,
  scanAuthenticationPrompt: {
    isInteractive: () => true,
    select: async () => { prompted = true; return "chatgpt"; },
  },
};

await main(["scan", ".", "--json"], out, errTTY, deps);
await main(["scans", "rerun", "scan-original", "--json"], out, errTTY, deps);
```

Observed result:

```text
$ codex-security scan . --json
   interactive credential prompt shown: false

$ codex-security scans rerun scan-original --json
   interactive credential prompt shown: true
   stderr contained: Both a ChatGPT sign-in and an API key from OPENAI_API_KEY are available.
```

## Expected behavior

`scans rerun` should honour the same rule as `scan`: JSON and JSONL output never prompt. `tests-ts/cli-authentication.test.ts` already pins this for `scan` ("never prompts during automation"); the equivalent guarantee is missing for `rerun`.

## Actual behavior

`scans rerun <id> --json` prompts for a credential choice and writes the prompt banner to stderr. In an automated context that keeps a pty attached, the command blocks waiting for input that will never arrive.

## Impact

`scans rerun` is the documented way to re-check a fixed vulnerability, which makes it a natural fit for automation. A wrapper that runs it with `--json` and a pty can hang indefinitely rather than failing fast.

## Related

The same missing plumbing means `scans rerun <id> --format md` is accepted, while `scan --format md` is rejected up front as unsupported. Forwarding the format to `runScan` would address both.

## Suggested direction

Pass the computed interactivity through from the rerun command, mirroring the `scan` call site.

mariohercules:
Fixes #39.

## Problem

`scans rerun` called `runScan` without forwarding the requested output format, so it always took the `interactive = true` default. `scan` computes it from the format:

```ts
format !== "json" && format !== "jsonl",
```

`interactive` gates the credential-selection prompt and the TTY progress rendering, so `scans rerun <id> --json` could block on a prompt that `scan --json` suppresses in exactly the same situation. In an automated context holding a pty, the command waits for input that never arrives.

The same missing plumbing meant `scans rerun --format md` was accepted, while `scan --format md` is rejected up front as unsupported for scan results — both commands produce the same scan payload.

## Change

Forward the format from the rerun command so both behaviors match `scan`.

## Tests

Two tests, both verified to fail without the change:

- `tests-ts/cli-authentication.test.ts` — `scans rerun --json` and `--format jsonl` never prompt, mirroring the existing "never prompts during automation" guarantee that covered only `scan`.
- `tests-ts/cli-workbench.test.ts` — `scans rerun --format md` exits 2 without starting a scan.

Full suite on this branch: 719 pass, 5 skip, 0 fail. `pnpm run types` and `pnpm run format` are clean.


mldangelo-oai:
@codex review exact head 510b925031af39aa01a181608021bc52ee176e8d

PR #205, opened 2026-08-01
fix: expire remediation claims on Python 3.10

### LINKED ISSUE #35: Python 3.10: workbench timestamps fail to parse, so remediation claim leases never expire
## Summary

On Python 3.10, `workbench_db.py` cannot parse the timestamps it writes itself, so remediation claim leases never expire and abandoned claims permanently block a finding.

Every workbench timestamp is written with a `Z` suffix ([`workbench_db.py` line 122](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/workbench_db.py#L122)):

```python
def now() -> str:
    return datetime.now(timezone.utc).isoformat().replace("+00:00", "Z")
```

and read back with a bare `datetime.fromisoformat` ([lines 132-144](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/workbench_db.py#L132-L144)):

```python
def remediation_claim_is_active(remediation: sqlite3.Row) -> bool:
    ...
    try:
        parsed = datetime.fromisoformat(claimed_at)
        if parsed.tzinfo is None:
            return True
    except ValueError:
        return True
```

`datetime.fromisoformat` only accepts the `Z` suffix from Python 3.11 onward. On 3.10 the call raises `ValueError`, the handler returns `True`, and the lease check becomes unconditional: a claim is always reported as active regardless of `CLAIM_LEASE_SECONDS` or `DELIVERED_ACTION_LEASE_SECONDS`.

The package supports Python 3.10 (`README.md` states 3.10 or later, and the interpreter probe in `src/runtime.ts` enforces `>= 3.10`), so this is a supported configuration.

The rest of the codebase already handles this correctly — [`finalize_scan_contract.py` line 649](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py#L649) normalizes the suffix before parsing:

```python
datetime.fromisoformat(value[:-1] + "+00:00" if value[-1] in "Zz" else value)
```

This call site was missed.

- macOS 26.5.2 (build 25F84)
- Python 3.10.20 (affected) and Python 3.12.13 (not affected)

## Steps to reproduce

Using the real `workbench_db.py`, with a claim far beyond the 120-second lease:

```python
import sys
from datetime import datetime, timedelta, timezone

sys.path.insert(0, "sdk/typescript/_bundled_plugin/scripts")
import workbench_db as w

old = (datetime.now(timezone.utc) - timedelta(days=30)).isoformat().replace("+00:00", "Z")

row = {
    "pending_action_claim_token": "tok_abandoned",
    "pending_action_delivered_at": None,
    "pending_action_claimed_at": old,
}

print(sys.version_info[:2], w.CLAIM_LEASE_SECONDS, w.remediation_claim_is_active(row))
```

Observed result:

```text
python 3.10  CLAIM_LEASE_SECONDS=120  claimed_at=2026-06-29T01:32:39.732153Z (30 days old)
  remediation_claim_is_active -> True   BUG: stale claim reported as active

python 3.12  CLAIM_LEASE_SECONDS=120  claimed_at=2026-06-29T01:32:39.796613Z (30 days old)
  remediation_claim_is_active -> False   correct: lease expired
```

(On 3.10 this requires `tomli`, as documented, because `workbench_db.py` transitively imports `deep_scan_config`.)

The underlying parse failure in isolation:

```text
python 3.10  stamp=2026-07-29T01:32:09.047781Z  -> ValueError: Invalid isoformat string: '2026-07-29T01:32:09.047781Z'
python 3.12  stamp=2026-07-29T01:32:09.072262Z  -> PARSED OK
```

## Expected behavior

A remediation claim older than its lease should be treated as expired on every supported interpreter, so the stale-claim recovery paths work.

## Actual behavior

On Python 3.10 the claim is always active. Both recovery paths that depend on lease expiry become unreachable:

- Closing the finding is refused with *"Wait for the pending remediation operation to finish before closing this finding."* ([line 1878](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/workbench_db.py#L1873-L1883))
- Regenerating the remediation is refused with *"Finish or retry the active remediation operation before regenerating."* ([line 1987](https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/_bundled_plugin/scripts/workbench_db.py#L1986-L1991))

Both guards are written so that a `failed` state plus an expired lease is the escape hatch, which is exactly the condition that can never become true here.

## Impact

If a remediation worker dies while holding a claim, `pending_action_claim_token` stays set and the lease can never clear it. On Python 3.10 the finding is permanently stuck: it can be neither closed nor regenerated, and the only remaining path is manual database intervention.

Note that the SQL-based staleness checks elsewhere in the file compare timestamps as strings and are unaffected, so the two mechanisms disagree with each other on 3.10.

## Suggested direction

Normalize the suffix before parsing at this call site, matching what `finalize_scan_contract.py` already does. It may be worth factoring that normalization into a single shared helper so a third call site cannot drift again.

mariohercules:
Fixes #35.

## Problem

`now()` writes every workbench timestamp with a trailing `Z`:

```python
return datetime.now(timezone.utc).isoformat().replace("+00:00", "Z")
```

but `remediation_claim_is_active` read it back with a bare `datetime.fromisoformat`, which only accepts that suffix from Python 3.11 onward. On Python 3.10 — a supported interpreter, per `README.md` and the `>= 3.10` probe in `src/runtime.ts` — the call raised `ValueError` and the handler fell through to `return True`.

The lease check therefore became unconditional: a claim was always reported as active regardless of `CLAIM_LEASE_SECONDS` or `DELIVERED_ACTION_LEASE_SECONDS`. Both stale-claim recovery paths were unreachable, so an abandoned claim permanently blocked the finding:

- closing it → *"Wait for the pending remediation operation to finish before closing this finding."*
- regenerating it → *"Finish or retry the active remediation operation before regenerating."*

Confirmed against the real module on Python 3.10.20, with a 30-day-old claim and a 120-second lease:

```text
python 3.10  remediation_claim_is_active -> True    (stale claim reported as active)
python 3.12  remediation_claim_is_active -> False   (correct)
```

## Change

Normalize the suffix before parsing, matching the idiom `finalize_scan_contract.py` already uses for RFC 3339 timestamps. These were the only two `fromisoformat` call sites in the plugin, and this was the one that lacked normalization.

After the change, the same probe on Python 3.10.20 reports `False`.

## Tests

`tests-ts/workbench-remediation-lease.test.ts` drives the real `workbench_db` module through a Python probe, following the pattern in `workbench-canonical-paths.test.ts`.

Because Python 3.11+ parses `Z` natively, a test running on a modern interpreter would pass with or without the fix. The probe therefore substitutes a `datetime` subclass whose `fromisoformat` rejects a trailing `Z`, reproducing the 3.10 parser on any interpreter. Verified to fail on `main` (`active: true`) and pass with the change, on Python 3.12.

Full suite on this branch: 719 pass, 5 skip, 0 fail. `pnpm run types` and `pnpm run format` clean.


riju568:
# workbench_db.py

import logging
from datetime import datetime, timezone

logger = logging.getLogger(__name__)

def parse_utc_iso(ts_str: str) -> datetime:
    """
Parses an ISO 8601 string reliably under python 3.10+.
 Normalises both the ‘Z’ and ‘z’ suffix and guarantees the return is atzinfo-Aware UTC datetimestamp.
    """
    if not isinstance(ts_str, str) or not ts_str.strip():
        raise ValueError("Timestamp must be a non-empty string")

    cleaned = ts_str.strip()
    if cleaned.endswith(("Z", "z")):
        cleaned = cleaned[:-1] + "+00:00"

    dt = datetime.fromisoformat(cleaned)

    # Convert naive datetimes to UTC to prevent aware/naive comparison TypeErrors
    if dt.tzinfo is None:
        dt = dt.replace(tzinfo=timezone.utc)
    else:
        dt = dt.astimezone(timezone.utc)

    return dt


def remediation_claim_is_active(claim: dict, lease_seconds: float) -> bool:
    """
Checks if the lease is still valid in claims. Returns false if, but missing, Malformed, expired
    """
    if not isinstance(claim, dict):
        return False

    created_at_raw = claim.get("created_at") or claim.get("timestamp")
    if not created_at_raw:
        return False

    try:
        created_at = parse_utc_iso(created_at_raw)
        now = datetime.now(timezone.utc)
        
        elapsed_seconds = (now - created_at).total_seconds()

        if elapsed_seconds < 0:
            return True

        return elapsed_seconds < lease_seconds

    except (ValueError, TypeError) as err:
        logger.warning(
            f"[workbench] Failed to parse claim timestamp '{created_at_raw}': {err}. "
            "Treating claim as expired."
        )
        return False

riju568:
tests-ts/workbench-remediation-lease.test.ts
--------------------------------------------------------------------------------
import unittest
from datetime import datetime, timezone, timedelta
class TestRemediationClaimLease(unittest.TestCase):
    def test_python_310_z_timestamp_expiration(self):
        stale_time = (datetime.now(timezone.utc) - timedelta(minutes=10)).isoformat().replace("+00:00", "Z")
        claim = {"created_at": stale_time}
        self.assertFalse(remediation_claim_is_active(claim, lease_seconds=120))

    def test_active_z_timestamp(self):
       fresh_time = (datetime.now(timezone.utc) - timedelta(seconds=10)).isoformat().replace("+00:00", "Z")
        claim = {"created_at": fresh_time}
        self.assertTrue(remediation_claim_is_active(claim, lease_seconds=120))

    def test_malformed_timestamp_does_not_lock(self):
        claim = {"created_at": "invalid-iso-string"}
        self.assertFalse(remediation_claim_is_active(claim, lease_seconds=120))

if __name__ == "__main__":
    unittest.main()
---------------------------------------------------------------------------------

riju568:
Add this python unit.

riju568:
1) Automatically applies tzinfo=timezone.utc because can't compare offset-naive and offset-aware datetimes: TypeError: can't compare offset-naive and offset-aware datetimes.
2) Logs( warn, False ); if badly formatted date/time we do not prevent findings report in the long term.

riju568:
in the .ts file

pasmud (commented):
I checked f0ef3fd locally on Windows with Bun 1.3.13 and Python 3.13; the regression test deliberately emulates the pre-3.11 datetime parser. Both focused lease tests pass, as do the model-generation check, TypeScript check, and formatting check. I also traced the unchanged malformed/naive timestamp behavior, which still treats an uncertain claim as active. No blocking issues from my review- the suffix normalization is narrowly scoped and consistent with the existing RFC 3339 parsing approach.

mldangelo-oai:
@codex review exact head 374ef424cec1278bfaa9adbf608c25edb4842a28

PR #208, opened 2026-08-02
feat(git): restrict submodule recursion and credential access

nidsnitesh:
## Summary
Fixes a security issue where Git credential helpers could leak GitHub access tokens (`GH_TOKEN`/`GITHUB_TOKEN`) when processing submodules.

### Details
When scanning a repository, a malicious `.gitmodules` file pointing to another repository on the same host (e.g. `github.com`) would cause Git's credential helper to be invoked for the submodule fetch. Since the helper checked the host but not the target repository path or context, it would hand out the token to the attacker-controlled repo.

### Fix
This PR implements 3 layers of defense-in-depth:
1. **`entrypoint.sh`**: Configures `submodule.recurse=false` and `fetch.recurseSubmodules=false` via Git environment variables so submodules are never fetched automatically.
2. **`git-credential.sh`**: Checks if `GIT_DIR` points to a submodule context (`*/.git/modules/*`) and refuses to return credentials.
3. **`multiscan.ts`**: Appends `-c submodule.recurse=false` to checkout commands in the SDK.


mldangelo-oai:
@codex review exact head d6c1b56e48954f1818785d3ec8fbf1e0ea66bcb0

PR #209, opened 2026-08-02
feat(container): reject hostnames with leading or trailing hyphens

nidsnitesh:
## Summary
Hardens the `CODEX_SECURITY_GIT_HOST` validation in `docker/entrypoint.sh` to reject hostnames starting or ending with hyphens.

### Details
Previously, the POSIX `case` pattern checked for empty strings, leading/trailing dots, consecutive dots, and characters outside `[A-Za-z0-9.-]`. However, strings starting or ending with a hyphen (such as `-evil.com` or `--flag.com`) passed validation.

Per RFC 952 / RFC 1123 standards, hostnames must start and end with an alphanumeric character. Hyphen-prefixed strings could also be interpreted as option flags when interpolated into commands or configuration keys.

### Fix
- Added `-*|*-` to the rejection case pattern in `docker/entrypoint.sh`.
- Added corresponding test assertions in `sdk/typescript/tests-ts/container-entrypoint.test.ts`.


mldangelo-oai:
@codex review exact head b7579d6a128b4dc6702f79f1c29a754f715b793d

PR #217, opened 2026-08-03
fix(scripts): accept CRLF tar listings when inspecting npm tarballs

### LINKED ISSUE #32: check-package.mjs rejects valid npm tarballs when Windows tar emits CRLF
## Summary

`scripts/check-package.mjs` rejects a valid npm tarball on Windows when the system `tar` command emits CRLF line endings for `tar -tvzf`.

The validation uses a multiline regular expression directly on the complete listing. In JavaScript, both CR and LF are line terminators for multiline anchors, so `^` can match between the `\r` and `\n`. The following `\n` is then tested as if it were the first character of a new line and triggers the non-regular-entry error.

- PowerShell 7.6.3
- Node.js 24.18.0
- pnpm 11.17.0
- bsdtar 3.5.2 / libarchive 3.5.2

## Steps to reproduce

From the repository root:

~~~powershell
cd sdk/typescript
corepack pnpm install --frozen-lockfile
corepack pnpm run build
corepack pnpm pack --pack-destination ../../dist
node scripts/check-package.mjs ../../dist/openai-codex-security-0.1.1.tgz
~~~

Observed error:

~~~text
Error: npm tarball contains a non-regular entry (symbolic or hard link, device, or pipe).
    at scripts/check-package.mjs:185
~~~

I then captured the exact `tar --ignore-zeros -tvzf` output and checked every non-empty line. The generated archive contained 178 entries, and every listing line began with `-` or `d`; there were no links, devices, or pipes.

The package installation smoke test also passed:

~~~text
Validated installed @openai/codex-security@0.1.1: public import, CLI, and 94 bundled plugin files.
~~~

A minimal diagnostic using the same listing demonstrated the false match:

~~~text
/^[^d-]/mu.exec(listing) matches the LF in the first CRLF sequence.
~~~

## Expected behavior

A valid npm tarball containing only regular files and allowed directories should pass `check-package.mjs` regardless of whether the local `tar` command emits LF or CRLF.

## Actual behavior

The validator reports a non-regular tar entry before the later code normalizes the same listing with `split(/\r?\n/u)`.

## Impact

Windows contributors using a CRLF-emitting `tar` implementation cannot run the repository's package-boundary validation locally, even though the generated package is valid and installable.

## Relevant code

- Current check: https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/scripts/check-package.mjs#L183-L190
- The code immediately below already creates normalized `listingLines`: https://github.com/openai/codex-security/blob/f22d4a36f26d16287bcdfd707b369116e02a08c3/sdk/typescript/scripts/check-package.mjs#L188-L198

PR #3 introduced the package-boundary checker and validated a Windows CI configuration, but its patch and review discussion do not cover a `tar` implementation that emits CRLF for verbose listings.

## Suggested direction

Normalize the listing before checking entry types and reuse the normalized lines for both validations. For example:

~~~js
const listingLines = listing.split(/\r?\n/u).filter(Boolean);
if (
  listingLines.some(
    (line) => !line.startsWith("d") && !line.startsWith("-"),
  )
) {
  throw new Error(
    "npm tarball contains a non-regular entry (symbolic or hard link, device, or pipe).",
  );
}
~~~

A focused regression test can pass a CRLF-formatted verbose listing with only regular files and assert that it is accepted.

rohanpoudel2:
Fixes #32

## Problem

`scripts/check-package.mjs` validated the whole `tar --ignore-zeros -tvzf` listing in one shot:

```js
if (/^[^d-]/mu.test(listing)) {
  throw new Error(
    "npm tarball contains a non-regular entry (symbolic or hard link, device, or pipe).",
  );
}
```

With the `m` flag, JavaScript treats **both** `\r` and `\n` as line terminators. On a CRLF listing, `^` therefore matches the position between the `\r` and the `\n`, and the `\n` is inspected as if it were the first character of an entry line. It is neither `d` nor `-`, so a perfectly valid archive is rejected:

```
Error: npm tarball contains a non-regular entry (symbolic or hard link, device, or pipe).
```

Windows contributors whose `tar` emits CRLF cannot run `pnpm run check:package` at all, as reported in #32 with bsdtar 3.5.2.

The bug is only in this early check. A few lines further down the same listing is already normalized correctly with `split(/\r?\n/u)`.

## Change

Hoist the split that already exists and check each line individually, so the entry-type check and the entry-count check now read the same normalized lines:

```js
const listingLines = listing.split(/\r?\n/u).filter(Boolean);
if (listingLines.some((line) => !/^[d-]/u.test(line))) {
```

I audited the rest of the script for the same class of defect — the `sizeField` check, the `entries` split, `unsafePath`, the launcher-permission split, `internalMarker`, and the `.br`/`.part-N`/PNG filename tests. None of them combine the `m` flag with a multi-line input, and JavaScript's `$` without `m` does not permit a trailing newline, so there is no anchor bypass anywhere else. This was the only occurrence.

## Why not strip the carriage returns and keep the multiline regex

`listing.replace(/\r/gu, "")` ahead of the existing `/^[^d-]/mu` test would also fix #32, but it leaves the file normalizing the same text twice, in two different ways, four lines apart. The entry-type check and the entry-count check would still be free to disagree about what a line is. Hoisting the split that was already there makes both checks read one array, and deletes a normalization rather than adding one.

## Why not split on `/\r\n|\r|\n/u` to also cover lone-CR line endings

`/\r?\n/u` does not split on a bare `\r`, so a hypothetical CR-only listing collapses into a single line and only its first character is type-checked. The wider separator would close that, at the cost of disagreeing with the `/\r?\n/u` split used for `entries` on line 74 — the two arrays are compared index-by-index, so they must be normalized identically.

It is unnecessary, and I measured rather than assumed it. Driving the real script with a `tar` shim that pipes the listing through `tr '\n' '\r'`:

```
Error: npm tarball contains an invalid tar entry.
```

The existing `listingLines.length !== entries.length` comparison catches it. The narrower split is also the only one that stays consistent with line 74.

The neighbouring worry — a control character inside an entry *name* surviving the split — does not reach the parser either. I packed a tarball actually containing `package/dist/index.js\r` (a real CR in the filename) and listed it: both bsdtar and GNU tar escape control characters in `-t` output, so the name arrives as a literal backslash and `r`, and the script rejects it:

```
Error: npm tarball is missing package/dist/index.js.
```

The pre-existing `file.includes("\\")` guard covers the same ground independently.

## Why not assert against a hand-written listing instead of packing a real tarball

A synthetic fixture would let the test pass against a fix that does not survive real `tar` output — the bug is precisely in how genuine listing bytes are parsed. The test therefore runs `pnpm pack` and the real `scripts/check-package.mjs` as a child process, with a `tar` shim on `PATH` that forwards every invocation to the real binary untouched except `-tvzf`, whose output it CRLF-terminates. The archive stays valid; only the listing changes.

The cost is that `pnpm pack` runs the `prepack` lifecycle, so the test rebuilds `dist/` (measured at ~1.2 s locally). That is safe here: `node-ci.yml` runs `Test` before `Build` and `Pack`, and no test in `tests-ts/` reads the real `dist/`.

## Impact, stated plainly

Behaviour on LF platforms is unchanged. The only listings whose verdict changes are ones the old code rejected wrongly: CRLF-terminated listings, and listings containing a blank line. Every genuinely non-regular entry — symlink, hard link, device, pipe — is still rejected, on both LF and CRLF, and there is a test pinning that.

Nothing shipped changes. `scripts/` is not in the package's `files` list, so `check-package.mjs` is a repository-side release-boundary check only, run by CI's "Inspect package" step and by contributors via `pnpm run check:package`.

## Verification

New `tests-ts/check-package.test.ts` adds two tests: one asserting a CRLF-listed valid tarball is accepted, one shimming the listing with `sed 's#^-#l#'` to confirm a genuinely non-regular entry is still rejected, so the fix cannot silently become a no-op.

Revert-proof. With the fix in place:

```
 2 pass
 0 fail
```

With `scripts/check-package.mjs` reverted to the original regex and nothing else changed:

```
 1 pass
 1 fail
```

failing at `scripts/check-package.mjs:190` with issue #32's exact error, `npm tarball contains a non-regular entry (symbolic or hard link, device, or pipe).`

End-to-end against the real script and a real `pnpm pack` tarball, driven through `tar` shims:

- CRLF listing, valid entries → `Validated …: 179 entries.`
- CRLF listing, entry type forced to `l` → rejected, non-regular entry
- LF listing, entry type forced to `l` → rejected, non-regular entry
- CR-only listing → rejected, `npm tarball contains an invalid tar entry.`

Full suite (`bun test --timeout 30000 ./tests-ts`): **740 pass / 5 skip / 0 fail**, 5165 `expect()` calls, 745 tests across 35 files.

`pnpm run types` (`generate:models:check` plus `tsc --noEmit`) is clean. `pnpm run format` reports `All matched files use Prettier code style!`.

The shims are POSIX shell scripts, so the tests use the repository's existing `testPosix` helper and skip on Windows — the same pattern `tests-ts/container-entrypoint.test.ts` uses at line 8. The Windows listing behaviour is what is being reproduced, and the fix itself is platform-independent.


mldangelo-oai:
@codex review exact head 7011008cca8c482d55e86ceeb623a84acb3f8918

rohanpoudel2:
Independent confirmation that the premise here still holds on a current bsdtar, from @Souptik96's testing in #250:

On Windows 11 with `C:\Windows\System32\tar.exe` = **bsdtar 3.8.4 / libarchive 3.8.4**, the `-tvzf` listing came back with **179 CRLF line endings and 0 bare LF** — through a pipe, not a console. #32 reported this on bsdtar 3.5.2; it is unchanged two minor versions later, so this is not a stale-toolchain problem that will age out.

For contrast, the same binary through the same pipe emits entry content untranslated (`-xOf` was byte-exact), which is why only the listing check is affected and the fix stays scoped to it.

`main` still has the unguarded check:

```js
const listing = tar(["-tvzf", archive], "utf8");
if (/^[^d-]/mu.test(listing)) {
```

so any Windows contributor on the bundled `tar` still can't run `check:package`.

Souptik96:
Glad the measurement helped - hope this lands soon.

slegarraga (approved):
Reviewed the diff. Normalizing the tar listing before checking entry types fixes the CRLF false positive while preserving the non-regular-entry rejection, and the tests exercise both CRLF and malicious listings. Looks good.

ting-hong-shieh (approved):
Approved at exact head `1bcb7bc5272638d443085363ddffe0a3b24d1921`.

I independently exercised the same real 203-entry npm tarball with only the `tar -tvzf` output converted to CRLF. On base `455d7c8c68e515fec3756052863e94229e37a520`, the checker exited 1 with the reported non-regular-entry error. On this head, it exited 0 and reported `Validated …: 203 entries.` The added negative test still rewrites regular-file listing entries as links and confirms that they are rejected.

Validation:
- Exact head: 2 focused tests passed; full suite 1,039 passed / 11 platform or integration skips / 0 failed; types, formatting, and whitespace checks passed.
- Clean synthetic merge onto current `main` `2f785900d2cefb932cf97fb1d675491d14790839`: 2 focused tests passed; full suite 1,084 passed / 11 platform or integration skips / 0 failed; types, formatting, and whitespace checks passed.

No provider or API calls and no credentials were used. The checks ran locally on macOS arm64 with Node 26.5.0, Bun 1.3.14, and Python 3.13.14.

rohanpoudel2:
Superseded by #436, which merged the same line-oriented CRLF/LF tar-listing fix with focused coverage and passing cross-platform CI. Closing this duplicate now that the underlying bug is fixed on main.

slegarraga:
Thanks for closing the loop properly, @rohanpoudel2 — a duplicate closed *because* the bug is fixed on main is the best possible outcome for this one. Glad the measurement helped get it over the line, and thanks @Souptik96 for the nudge along the way.

PR #233, opened 2026-08-03
fix(contract): reject control characters in the scan target remote

### LINKED ISSUE #231: scan.target.remote accepts embedded tab/newline/CR because new URL() strips them before validating
## Summary

`validateCanonicalContract` validates `scan.target.remote` against a string that is not the string it stores. The authority regex matches ASCII tab, LF and CR, and the follow-up `new URL(remote)` check passes because the WHATWG URL parser *strips* those three characters from the input before parsing. The unsanitized original is then returned to callers as `manifest.scan.target.remote`.

## Environment

## Detail

In `sdk/typescript/src/contract.ts` the authority is extracted with:

```ts
const authority = /^[A-Za-z][A-Za-z0-9+.-]*:\/\/([^/?#]+)/.exec(remote)?.[1];
```

`[^/?#]` matches `\t`, `\n` and `\r`. `new URL(remote)` then does not catch it, because URL preprocessing removes exactly those three characters.

Reproduction — a manifest carrying:

```json
"remote": "https://example.com\nhttps://evil.example.net"
```

loads successfully. `new URL()` saw host `example.comhttps` (the newline is stripped and the two halves fuse), while `loadContract` hands back the original two-line string with the newline intact.

Neither surrounding guard helps:

- the JSON-schema pattern `^(?![^:/?#]+://[^/?#]*@)[^?#]+$` — `[^?#]` also matches newlines;
- `validateParsedJson` — checks only for well-formed Unicode, not control characters.

By contrast `requireModelSafeOutputDir` in `src/runtime.ts` already rejects C0/C1 characters for exactly this reason, with a named constant.

Enumerating every C0/C1 code point through `new URL("https://exa<c>mple.com")` confirms the scope precisely:

```
SURVIVED new URL(): 09->example.com 0a->example.com 0d->example.com
threw count: 62
```

Only tab, LF and CR are smuggled through; the other 62 make `new URL()` throw, so they are already caught.

## Impact — stated plainly

This is a broken guarantee rather than a live exploit in the CLI. Nothing under `src/` currently consumes `.remote`, and credential smuggling is correctly blocked (any `@` in the authority is rejected before URL parsing). The value is that the contract layer is exactly where an untrusted manifest field is supposed to be sanitized, and SDK consumers that render, log or re-emit `manifest.scan.target.remote` inherit a multi-line value that passed validation.

## Suggested direction

Reject `remote` containing any C0/C1 control character or Unicode line/paragraph separator before the regex and URL checks run, mirroring the character class `runtime.ts` already uses.

I considered additionally requiring `new URL(remote).href === remote` so the stored string must equal its own canonical serialization, and rejected it: the WHATWG serializer legitimately rewrites real URLs — it strips a default `:443`, lowercases mixed-case hostnames, and punycode-encodes internationalized ones — so exact equality would reject valid remotes for no security benefit, given tab/LF/CR are the only characters that actually slip through.

PR follows.

rohanpoudel2:
## Summary

Reject control characters in `scan.target.remote` before URL parsers normalize them, keeping the bundled producer and TypeScript loader in agreement.

Fixes #231.

## Changes

- Reject C0/C1 controls and Unicode line separators before validating or storing remote URLs.
- Apply the same rule in the bundled Python finalizer and the TypeScript contract loader.
- Preserve valid HTTPS, SSH, IDN, IPv6, explicit-port, mixed-case, and percent-encoded remotes.
- Return ordered producer verdict rows in the parity test so control-character inputs are not used as platform-sensitive JSON object keys.

## Testing

After merging current `main` into this branch:

- `bun test --timeout 30000 tests-ts/contract.test.ts` — 45 passed, 0 failed, including the producer/loader parity matrix.
- `pnpm run types` with pnpm 11.9.0 — passed; the local Node 25 runtime emitted the repository's unsupported-engine warning, while supported runtimes remain covered by CI.
- `pnpm run format` with pnpm 11.9.0 — passed.
- Fresh GitHub checks, including the Windows matrix that previously exposed the object-key lookup failure, are pending for the pushed head.

## Risk and rollout

Only remotes containing control characters or Unicode line separators change verdict. Existing valid remote forms and the published schema remain unchanged. Previously sealed scans containing rejected characters will no longer load or project until corrected.

## Public disclosure review

The source change and newly authored material satisfy the second attestation. An existing automated review comment contains an access-restricted report link, so that attestation remains unchecked.


mldangelo-oai:
@codex review exact head dcbad0efe37c860fff564d2261b1b99a86a97762

PR #255, opened 2026-08-04
fix(multiscan): record run warnings on bulk-scan receipts

### LINKED ISSUE #248: bulk-scan discards run warnings, so a drifted repository is recorded as completed with no signal
## Summary

## Environment

## Where it is

## Impact

## Suggested direction

Two parts, both small:

rohanpoudel2:
Fixes #248

## Problem

`runMultiscan` never asks for a repository's warnings, and the per-attempt ledger receipt has nowhere to put them:

```ts
const result = await security.run(checkout, {
  ...(task.scope === undefined ? {} : { target: [task.scope] }),
  ...
  mode: task.mode,
  outputDir: scanDir,
  ...(options.signal === undefined ? {} : { signal: options.signal }),
});
```

```ts
`${JSON.stringify({
  ...task,
  status,
  attempt,
  outputDir: scanDir,
  ...(cost === null ? {} : { cost }),
  ...(failure === undefined ? {} : { error: failure }),
})}\n`
```

No `onWarning`, so every warning the run reports is dropped on the floor. A repository whose target drifted mid-run keeps `coverage.completeness === "complete"`, so the attempt genuinely succeeds: it is written as `status: "completed"`, `bulk-scan` exits 0, and the campaign summary counts it among `completed`. There is no failure, no non-zero status, and no field a consumer could read to notice the results describe a stale tree. The same holds for the other two warning producers — a cost limit that could not be verified, and the cleanup failures `run()` reports from its `finally` block.

Note for reviewers: `ScanResult` does **not** carry a `warnings` field on `main`. The only channel for these warnings is the `onWarning` observer.

## Change

- Each attempt installs an `onWarning` observer that collects that attempt's warnings into a fresh array. The array is per attempt, so a retry does not inherit the previous attempt's warnings.
- Collected warnings go through `redactedErrorMessage`, the same way `error` does. The CLI's own observer only sanitizes for stderr; these strings are about to be written to disk and read back on resume, and the comment on `warnCleanupFailed` is explicit that the warning text has so far never been persisted, so it has never been redacted upstream.
- The receipt gains an optional `warnings?: string[]`, written only when the attempt produced warnings.
- `readReceipts` records, as it parses the ledger, which repositories have **any** warned attempt, and returns that set beside the per-repository receipt map.
- `MultiscanResult` gains `warned: number` beside `failed`. The exit code is untouched: a warning is not a failure.

Warnings from a **failed** attempt are captured too, and deliberately so. `run()`'s cleanup warnings are emitted from its `finally` block, so they happen whether the attempt returned a result or threw, and the completion and cost-limit warnings are emitted before a later step can throw. Reading `result.warnings` — even if that field existed — would therefore lose exactly the warnings that accompany a failure. The observer is also immune to the two multiscan-side throws that happen after `run()` returns (the scope-escape check and the `coverage.completeness` check), whose receipts now carry both `error` and `warnings`.

On ordering: `run()` dispatches observers on a microtask. The receipt is serialized after this loop's `finally { await rm(checkout, ...) }`, which runs once `run()` has settled, so every queued warning has landed before the receipt is written.

## Why `warned` counts the ledger and not the last receipt

The first draft of this change read `receipt.warnings` off the single receipt `readReceipts` keeps for each repository — the last line the ledger holds for that id — and that is wrong, because warnings belong to the attempt that produced them. A repository whose attempt 1 warned and failed and whose attempt 2 completed quietly ends the run with a final receipt carrying no `warnings` key at all. The run that wrote that ledger reported `warned: 1` from its in-run flag; resuming the same campaign re-read only the last receipt and reported `warned: 0`, having done no new work. The attempt that warned was still sitting in the ledger, and a campaign script polling `warned` would have concluded the campaign was clean — the same silent drop this PR exists to close, moved one step later.

So `warned` is now a property of the ledger, exactly as `completed` already is: the skip branch counts from the warned-id set rather than the receipt that won, and the worker seeds `repositoryWarned` from that set so a repository the ledger already warned about keeps its count when a later run retries it and the retry is quiet. Because the ledger is append-only, `warned` cannot go down across a resume.

## Why an omitted-when-empty `warnings?: string[]` and not an unconditional array

The alternative is to always emit the key, `warnings: []` included, matching `scan --json`, which gets its unconditional `warnings` straight from `workbench_db.py`. Rejected, for two reasons:

1. **The receipt's own convention settles it.** `cost` and `error` — and `scope` on the task half — are all omitted when absent. `warnings: []` on every one of thousands of receipts would be the only field in the record that insists on being present in order to say nothing. `workbench_scan_history.py` already omits the key when `completion_warnings_json` is `"[]"`, so this side of the split is not novel either.
2. **The ledger is append-only and read back.** Omission keeps a quiet attempt's receipt byte-identical to what releases before this change wrote, which keeps the diff between an old ledger and a new one confined to attempts that actually warned. `"warnings" in receipt` becomes a meaningful "this attempt warned" test rather than a tautology.

## Why not fold an earlier attempt's warnings into the receipt that supersedes it

`readReceipts` keeps one receipt per repository, so the obvious way to stop losing a warned attempt is to carry its `warnings` forward onto the receipt that replaces it. Rejected: that produces a mixed object whose `attempt`, `outputDir` and `status` come from one ledger line while its `warnings` come from several, and `MultiscanReceipt` is the type the resume path reads `outputDir` and `attempt` off. A separate `Set<string>` of warned ids answers the only question the summary asks without inventing a receipt that was never written.

## Ledger compatibility

- **An older ledger without the field still resumes.** `readReceipts` does an unchecked `JSON.parse(line) as MultiscanReceipt` — there is no schema and no validator, in TypeScript or in the bundled Python; nothing under `_bundled_plugin/scripts/` reads `results.jsonl` at all. The resume path reads only `id`, `status`, `outputDir` and `attempt`. `warnings` is optional in exactly the way `error` is, so a receipt written before this change type-checks and resumes unchanged. The warned-id set is built through `Array.isArray(receipt.warnings) && receipt.warnings.length > 0`, which is `false` for a missing key and also for a hand-edited ledger holding a non-array there. Covered by a test that strips the key from a ledger and resumes: the repository stays skipped, is not rescanned, and the campaign reports `warned: 0`.
- **A newer ledger read by older code is unaffected.** `warnings` is an extra key on a line that older `readReceipts` parses with `JSON.parse` and casts; unknown keys are carried along and ignored. No existing field changed name, type, or emission condition.

## Impact, stated plainly

A drifted repository is still recorded as `completed`, because it did complete — but its receipt now names the drift, and the campaign summary reports `warned: 1`, on the run that found it and on every resume after it. A scripted campaign can act on `warned` without parsing the ledger; a human can read the offending message off the receipt. `bulk-scan`'s exit code, the `completed`/`failed`/`skipped` counts, the manifest, and the receipt fields that already existed are all unchanged. `onProgress` is deliberately left alone — this change adds a durable record, not a new stream of stderr chatter.

## Verification

From `sdk/typescript`:

- `bun test --timeout 30000 ./tests-ts` — **778 pass, 5 skip, 0 fail** (783 across 34 files).
- `multiscan.test.ts` alone — **19 pass, 0 fail**.
- With `src/multiscan.ts` reverted to its state on `main` and the five new tests left in place — **14 pass, 5 fail**: each new test fails without the change, and every pre-existing multiscan test still passes.
- With `src/multiscan.ts` reverted only to this PR's first commit — **17 pass, 2 fail**: the two ledger-counting tests fail, which is the defect the second commit fixes.
- `tsc --noEmit` clean; `generate:models:check` clean; `prettier --check` reports "All matched files use Prettier code style!".

Five tests in `sdk/typescript/tests-ts/multiscan.test.ts`, in its existing mock-outcome style:

1. `records a completed attempt's warnings on its receipt and in the summary` — two repositories, one warns. Asserts the warned receipt carries `warnings`, that the quiet receipt has no `warnings` property at all, that the summary reports `warned: 1` with `failed: 0`, and that a synthetic `sk-proj-` credential inside the warning does not reach the ledger.
2. `records a failed attempt's warnings and counts its repository once` — attempt 1 warns then throws, attempt 2 warns then completes. Asserts both receipts carry their own warning, that the failed one carries `error` alongside, and that `warned` counts the repository once rather than per attempt.
3. `resumes receipts written before the ledger carried warnings` — resumes a ledger with the field, then rewrites it without the key and resumes again.
4. `keeps a retried repository's warnings in a resumed summary` — attempt 1 warns and fails, attempt 2 completes quietly, so the last receipt carries no warnings; asserts the resume that skips the repository still reports `warned: 1`.
5. `keeps a rescanned repository's earlier warnings in the summary` — one attempt per run, so the second run rescans rather than skips; asserts all three runs agree on `warned: 1`.

Plus one assertion added to the existing `runs a bulk scan and keeps structured output on stdout` test in `cli.test.ts`, so `warned` is proven to survive the command's `z.record` output schema and reach stdout rather than only the `runMultiscan` return value.


ting-hong-shieh (approved):
I reviewed head `899828427d0f3801df1a21cd39fdeb30350528d9` and found no correctness blocker in the warning-ledger change.

The receipt keeps warnings scoped to the attempt that emitted them and redacts them before persistence. Tracking warned repository IDs separately also preserves a warning across retries and resumes without attaching an earlier attempt's warnings to the latest receipt. The resulting `warned` value remains a repository count rather than an attempt count.

I also replayed the change onto current `main` at `216212b70a421b0bf30319d241de47a64050ec0f`. In that replay I retained the newer incomplete-coverage handling and composed warning capture with the current non-fatal progress observer. The replay passed 46 multiscan tests (215 assertions), the focused bulk-scan CLI test (8 assertions), and the randomized full SDK suite (1,084 pass, 11 skip, 0 fail). The exact PR head passed 43 multiscan tests (205 assertions), the same focused CLI test, and its randomized full SDK suite (1,042 pass, 11 skip, 0 fail). Generated-model checks, TypeScript checks, the build, Prettier, and whitespace checks were clean in both worktrees.

Validation used Bun 1.3.14 and Python 3.12.13 on macOS, with local fixtures and mocks only; it did not call a model or provider, use API credentials, or exercise another native platform. GitHub currently reports the PR as conflicting, so the branch still needs a rebase or merge from `main`; the current-main replay above was validation only.

rohanpoudel2:
#466 now closes #248 and incorporates durable warning receipts and warned-task accounting related to this PR. Should I close this PR as superseded, or keep it open as the smaller alternative? I’ll hold off on rebasing until that ownership is clear.

PR #278, opened 2026-08-04
feat: retain rollout sessions for scan observability

### LINKED ISSUE #187: Expose a scan-session artifact index and optional rollout export
## Summary

codex-security creates standard Codex rollout sessions for coordinators and workers, but scan results do not identify those sessions or provide a way for external trace tools to index them. This is an issue particularly when a scan fails.

## Problem

A scan ran for ~3 hours and unexpectedly failed before producing a report. Its private Codex home still contained 87 valid rollout files, including parent metadata for all 56 child sessions.

Recovering them required manually correlating timestamps, temporary working directories, and session metadata. The scan record did not expose their location or IDs.

## Requested workflow

Allow users and local observability tools to retain and index the coordinator and worker traces associated with a scan, especially when the scan fails.

2830500285:
## Summary

- index the coordinator and all transitive worker rollout sessions in private scan artifacts on successful and failed scans, while excluding unrelated sessions and enforcing a 4,096-session bound
- add opt-in raw JSONL retention with private permissions, race-aware copying, and automatic index-only fallback when retention is unavailable
- expose the index through ScanResult, machine-readable failed scan output, scan history, reruns, and bulk scans
- persist the relative artifact path through the workbench schema and include the new modules in the npm package contract

## Product rationale

Long-running and delegated scans can fail after producing substantial execution history. Keeping a bounded relationship index by default gives operators and external indexers a stable correlation point, including on failure, without copying source-bearing prompts and tool output unless retention is explicitly enabled.

Rollout indexing is best-effort and cannot replace the primary scan outcome: retention failures warn and degrade to an index-only artifact, and indexing failures never mask the original scan failure.

## Validation

- pnpm run types
- pnpm run format
- pnpm test: 798 passed, 5 platform/integration skips, 0 failed
- packed-package inspection and installed-package smoke: 187 tarball entries and 95 bundled plugin files validated
- Python workbench modules compiled with py_compile
- focused success, failure, fallback, privacy, history, migration, CLI, and 4,096/4,097 boundary tests

Fixes #187


mldangelo-oai:
Thanks for putting this together. One interaction to watch: the new `CodexRolloutReader` in `src/codex-rollouts.ts` currently throws `Codex session event exceeds the 1 MiB safety limit.`, which reproduces #291 for legitimate large tool output. #326 changes the existing reader to discard oversized records and continue processing later `token_count` events. This PR's broader rollout-observability work is complementary, but the extracted reader should preserve that same nonfatal behavior.

PR #294, opened 2026-08-06
fix(tests): resolve Git Bash explicitly so WSL cannot claim `bash`

### LINKED ISSUE #206: Windows release tests invoke WSL bash instead of Git Bash when WSL is first on PATH
## Summary

On Windows systems with WSL enabled, the release-automation tests invoke an unqualified `bash` through `spawnSync`. Windows executable resolution can select `C:\Windows\System32\bash.exe` (WSL) before Git for Windows' Bash.

The extracted GitHub Actions shell fragments then run inside WSL, where the inherited Windows `node`, `gh`, and `jq` commands and the test harness' Windows-path assumptions are unavailable. This causes a large cluster of release-workflow test failures even though the same focused test passes when Git Bash is placed first on `PATH`.

## Environment

- Current `main`: `a8fc00984b07f10f5b607a9521c2f05aa57c5107`
- Package source version: `0.1.5`
- Windows 11 Pro, build 26200, x64
- Bun 1.3.13
- Git for Windows 2.51.2
- Also reproduced with Node 24.18.1 first on the Windows `PATH`

`where.exe bash` returns:

```text
C:\Windows\System32\bash.exe
C:\Program Files\Git\usr\bin\bash.exe
C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\bash.exe
```

## Reproduction

From `sdk/typescript`:

```powershell
bun test --timeout 30000 ./tests-ts/release-automation.test.ts --test-name-pattern 'executes the manual release cut against all published versions'
```

Observed:

```text
Expected to contain: "Release version must be greater than every published stable version."
Received: "/bin/bash: line 1: node: command not found
/bin/bash: line 1: node: command not found
/bin/bash: line 1: node: command not found
Release tags can only be cut from main."

0 pass
1 fail
```

The complete test suite produced 64 release-workflow failures with related symptoms such as `node: command not found`, `gh: command not found`, `jq: command not found`, missing mocked shell functions, and WSL paths. One additional authentication timeout was unrelated to this report.

## A/B validation

Changing only Bash ordering makes the exact same test pass:

```powershell
$env:PATH = 'C:\Program Files\Git\usr\bin;' + $env:PATH
bun test --timeout 30000 ./tests-ts/release-automation.test.ts --test-name-pattern 'executes the manual release cut against all published versions'
```

Result:

```text
1 pass
0 fail
```

## Root cause

`tests-ts/release-automation.test.ts` has multiple calls shaped like:

```ts
spawnSync("bash", ["-c", script], options)
```

The tests require a Bash environment compatible with the Windows executables and paths used by the harness, but they neither select Git Bash nor validate the resolved Bash implementation. With WSL's compatibility executable earlier on `PATH`, the child crosses into WSL instead.

## Expected behavior

The Windows test suite should either select a compatible Git Bash explicitly, validate the selected Bash before executing workflow fragments, or skip the shell-execution cases when only an incompatible Bash is available. Enabling WSL should not make the repository test suite fail based on `PATH` ordering.

## Suggested direction

A shared test helper could resolve and probe a compatible Bash once, then supply it to every workflow-shell test. On Windows the probe could require that the shell can execute the active Node binary and preserve the mocked command environment. If no compatible shell exists, these execution tests could be skipped with a clear reason while retaining the static workflow assertions.

ppcvote:
Fixes #206.

The release-automation suite spawns an unqualified `bash` in 20 places. On Windows that resolves against PATH, and with WSL enabled `C:\Windows\System32\bash.exe` is a real executable that frequently precedes Git for Windows. The extracted workflow fragments then run inside WSL, where the Windows `node`, `gh` and `jq` this suite depends on are absent and its Windows path assumptions do not hold.

## Measurement

I could not reproduce this by accident: on this host `where.exe bash` happens to return Git Bash first, so the suite was already green. I reproduced it deliberately by putting System32 ahead of Git on PATH, which is the reporter's configuration.

Windows 10, Git for Windows 2.52.0, WSL2 enabled, bun 1.3.11:

| PATH order | before | after |
|---|---|---|
| **System32 first** (as reported) | **116 pass, 69 fail** | **180 pass, 5 fail** |
| Git Bash first (this host's default) | 180 pass, 5 fail | 180 pass, 5 fail |

64 tests recovered, and the WSL-first environment now produces exactly what the Git-Bash-first environment does.

The 5 remaining failures are `jq: command not found` on this host. They are identical before and after and in both PATH orders, so they are an unrelated environment gap on my machine rather than residue from this change.

## The change

`resolveBash()` walks PATH and skips the launcher locations before taking the first `bash.exe` it finds:

- `%SystemRoot%\System32` and `%SystemRoot%\Sysnative` — the WSL launcher
- any `WindowsApps` directory — the Store execution alias, also WSL

Non-Windows platforms keep the bare `bash` they use today, so nothing changes on CI's Linux runners.

Two details worth flagging for review:

- PATH is split on `delimiter` from `node:path` rather than a literal `";"`. I checked what bun actually reports on Windows when launched from Git Bash, and it is a Windows-style semicolon PATH, but splitting on the platform separator is the honest expression of the intent.
- Skipping known launcher locations rather than positively identifying Git Bash is deliberate. Git for Windows ships `bash.exe` in both `Git\bin` and `Git\usr\bin`, and only the first sits next to `git.exe`, so a positive test would have to encode more layout assumptions than an exclusion does.

Happy to switch to a positive Git-Bash lookup, or to hoist the helper somewhere shared, if either fits the codebase better.

### DIFF
--- a/sdk/typescript/tests-ts/release-automation.test.ts
+++ b/sdk/typescript/tests-ts/release-automation.test.ts
@@ -1,6 +1,7 @@
 import { spawnSync } from "node:child_process";
 import { createHash } from "node:crypto";
 import {
+  existsSync,
   mkdirSync,
   mkdtempSync,
   readFileSync,
@@ -8,9 +9,49 @@ import {
   writeFileSync,
 } from "node:fs";
 import { tmpdir } from "node:os";
-import { join } from "node:path";
+import { delimiter, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { describe, expect, test } from "bun:test";
+
+/**
+ * Path to a POSIX bash for running the extracted workflow fragments.
+ *
+ * On Windows, an unqualified "bash" can resolve to C:\Windows\System32\bash.exe,
+ * the WSL launcher, whenever System32 precedes Git for Windows on PATH. The
+ * fragments then execute inside WSL, where the Windows node, gh and jq this
+ * suite relies on are not present and its Windows path assumptions do not hold,
+ * so a large cluster of release tests fails for reasons unrelated to the code
+ * under test. Skipping the two launcher locations picks Git Bash instead.
+ */
+const BASH = resolveBash();
+
+function resolveBash(): string {
+  if (process.platform !== "win32") {
+    return "bash";
+  }
+  const systemRoot = process.env.SystemRoot ?? "C:\\Windows";
+  const launchers = new Set(
+    [join(systemRoot, "System32"), join(systemRoot, "Sysnative")].map((entry) =>
+      entry.toLowerCase(),
+    ),
+  );
+  for (const entry of (process.env.PATH ?? "").split(delimiter)) {
+    const directory = entry.trim().replace(/^"|"$/g, "").replace(/[\\/]+$/, "");
+    if (directory.length === 0) {
+      continue;
+    }
+    const normalized = directory.toLowerCase();
+    // WindowsApps holds the Store execution alias, which is also WSL.
+    if (launchers.has(normalized) || normalized.endsWith("\\windowsapps")) {
+      continue;
+    }
+    const candidate = join(directory, "bash.exe");
+    if (existsSync(candidate)) {
+      return candidate;
+    }
+  }
+  return "bash";
+}
 
 type ReleaseMetadata = Record<string, unknown>;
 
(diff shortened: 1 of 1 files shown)


ppcvote:
Closing in favour of #225, which was open before this and covers the same ground.

I did not check for PRs already linked to the issue before opening this, which I should have. Apologies for the extra review load.

Issue #251, opened 2026-08-04
Run warnings never reach the SARIF projection, so drift is invisible to SARIF consumers


rohanpoudel2:
## Summary

A scan whose target drifted mid-run records a warning, but that warning never reaches the SARIF projection, so a SARIF consumer cannot tell the results describe a stale tree.

## Environment

- `@openai/codex-security` 0.1.5, current `main` at `ea19f24`
- macOS 15 (Darwin 25.5.0), Python 3.14.5

## Where it is

`sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py` populates `runs[].invocations[].toolExecutionNotifications` from `coverage["deferred"]`, and only when `coverage["completeness"]` is not `complete`.

A drifted target leaves `completeness` at `complete` — the scan really did review everything it set out to review; the tree simply moved underneath it. So the condition is false, no `invocations` block is emitted at all, and there is no field carrying the warning.

## Why it matters

SARIF is the machine-readable surface most CI systems consume, and `toolExecutionNotifications` with `level: "warning"` is precisely the SARIF idiom for "the run completed but something about it needs attention". A drift warning is the clearest example of that, and it is the one place it cannot currently appear.

Note this is not a projection bug alone — the notification structure is only ever built from `deferred` coverage rows, so run-level warnings have no route into SARIF regardless of `completeness`.

## Suggested direction

Emit an `invocations[]` entry whenever the scan recorded warnings, independently of `completeness`, mapping each warning to a `toolExecutionNotifications` entry at `level: "warning"` with `executionSuccessful: true`. That leaves the existing deferred-coverage notifications untouched and adds the run-level ones alongside.

This needs a bundled-plugin change plus SARIF schema validation, which is why it was kept out of PR #220 (that PR states it makes no `export` change). Related: #195.

JayYarlagadda:
I'd like to work on this. My implementation plan is:

- Read run-level warnings from the sealed scan manifest and map each one to a SARIF `toolExecutionNotifications` entry with `level: warning`.
- Emit an invocation whenever run warnings or deferred-coverage notifications exist, independently of coverage completeness.
- Preserve the existing deferred-coverage notifications and `codexSecurityCoverageCompleteness` behavior.
- Add tests for complete coverage with run warnings and for run warnings combined with deferred coverage.
- Validate the generated documents against the SARIF 2.1.0 JSON schema.

[...]

Issue #195, opened 2026-08-01
A scan whose target changed mid-run exits 0 and reports nothing in `--json`, so CI cannot detect that the results are for a stale tree


genforAI:
## Summary

When the scan target changes while a scan is running, the scan is completed against the original snapshot and a warning is written to stderr. The warning does **not** affect the exit code and is **not** present in `--json` output. A CI job therefore sees `exit 0` and a JSON document that is indistinguishable from a scan of the commit it actually checked out.

## Environment

`@openai/codex-security` **0.1.5** (commit `5625adc159bbbfb4bd85557d83e2805d27379230`)

## Detail

`scan_target_warning` produces the message ([`workbench_target.py` L529](https://github.com/openai/codex-security/blob/5625adc159bbbfb4bd85557d83e2805d27379230/sdk/typescript/_bundled_plugin/scripts/workbench_target.py#L529)):

```
Repository HEAD changed while the scan was running;
results were saved for the original revision.
```

It is appended to `warnings` at [`workbench_db.py` L1410-L1412](https://github.com/openai/codex-security/blob/5625adc159bbbfb4bd85557d83e2805d27379230/sdk/typescript/_bundled_plugin/scripts/workbench_db.py#L1410-L1412) and nothing else is changed — `coverage.completeness` stays `complete`.

The SDK forwards those strings only through the `onWarning` observer ([`api.ts` L829-L841](https://github.com/openai/codex-security/blob/5625adc159bbbfb4bd85557d83e2805d27379230/sdk/typescript/src/api.ts#L829-L841)), and the CLI's observer just writes to stderr ([`cli.ts` L2696-L2699](https://github.com/openai/codex-security/blob/5625adc159bbbfb4bd85557d83e2805d27379230/sdk/typescript/src/cli.ts#L2696-L2699)):

```ts
onWarning: (warning) => {
  errorOutput.write(`codex-security: warning: ${redactedErrorMessage(warning)}\n`);
},
```

Exit code selection never consults warnings ([`cli.ts` L2783-L2791](https://github.com/openai/codex-security/blob/5625adc159bbbfb4bd85557d83e2805d27379230/sdk/typescript/src/cli.ts#L2783-L2791)):

```ts
if (incomplete) {           // completeness !== "complete"
  ...
  return { exitCode: 2, data: result.toJSON() };
}
return { exitCode: blockingCount > 0 ? 1 : 0, data: result.toJSON() };
```

[...]

PR #295, opened 2026-08-06
fix(multiscan): record run warnings in the campaign ledger

### LINKED ISSUE #248: bulk-scan discards run warnings, so a drifted repository is recorded as completed with no signal

ppcvote:
Fixes #248.

`runMultiscan` called `security.run()` with no `onWarning` observer, so every warning the scan raised was discarded. A repository whose target drifted mid-run still completes, so it landed in `results.jsonl` as `status: "completed"` with nothing in the campaign output recording that the results describe a tree that moved.

The receipt now carries a `warnings` array when the scan raised any:

```json
{"id":"drifted","status":"completed","attempt":1,"outputDir":"...",
 "warnings":[{"message":"Scan target changed during the run.","kind":"target_changed"}]}
```

`kind` comes from `ScanWarningDetails`, so a consumer can single out drift without matching on message text — the same distinction the CLI already makes at its own `onWarning`.

## Three decisions worth review

**Redaction.** Warnings go through `redactedErrorMessage`, the redaction the failure path already applies. The ledger is a file on disk and a warning can quote a remote or a token. The CLI uses `sanitizeDiagnosticValue` for this, but that is private to `cli.ts`; `redactedErrorMessage` is the equivalent already imported into `multiscan.ts`. Happy to export the CLI one instead if you would rather have a single sanitizer.

**Per attempt.** The array is declared beside `failure` and `cost` inside the retry loop, so a second attempt does not inherit the first attempt's warnings.

**`status` is untouched.** The scan completed; promoting a warning to a failure would be a louder change than the issue asks for and would break resume, since `completed` receipts are what let a rerun skip finished repositories. The key is omitted entirely when nothing was raised, so a quiet run produces byte-identical output to today.

## Verification

Windows 10, bun 1.3.11, `bun test tests-ts/multiscan.test.ts`:

| | with fix | without fix |
|---|---|---|
| records run warnings in the receipt | pass | **fail** |
| redacts credential-shaped values | pass | **fail** |
| omits `warnings` when none raised | pass | pass (control) |
| **totals** | 16 pass, 1 fail | 14 pass, 3 fail |

The two failing-without-the-fix tests are the ones that matter; the third is a control that passes either way and guards against warnings appearing on a quiet run.

The single unrelated failure in both columns is `rejects output-directory symlinks`, which needs symlink privileges this host does not grant. It fails identically on unmodified `main`, so it is not residue from this change.

## Related

This is the multiscan half of the same shape as #251 and #195: the tool knew something and the knowledge did not reach the artifact a consumer reads. Those are about the single-scan `--json` and SARIF surfaces; this one is the ledger. I have not touched either of those paths — #251 already has someone working on it.


ppcvote:
Rebuilt against `main`, which had moved into this area and conflicted. Worth spelling out what changed, because the overlap is real but partial.

`main` now derives a `warning` from `result.coverage.completeness`, adds a `completed_with_incomplete_coverage` status and an `incomplete` tally. That covers the coverage case.

**It does not cover what this PR is about.** `security.run()` still receives no `onWarning` observer, so warnings the scan itself raises are still discarded. Coverage is computed locally from the returned result and never went through the observer path, so nothing about that change reaches the drift case in #248.

I kept the two signals separate rather than folding mine into `main`'s `warning`, because their semantics differ in a way that would produce a wrong status:

- `warning` is coverage-derived, and its presence **flips `status`** to `completed_with_incomplete_coverage`.
- A repository whose target drifted mid-run has **complete** coverage. The scan reviewed everything it set out to review; the tree moved underneath it. Routing a drift warning into `warning` would report incomplete coverage for a run whose coverage was fine, and move it out of the `completed` tally.

So the observer-sourced entries land in `scanWarnings`, renamed from `warnings` to remove the collision. The names now say which layer raised it: the campaign derives `warning`, the scan raises `scanWarnings`.

`scanPrompt` / `postScanPrompt` from `main` are kept as-is; `onWarning` is added alongside them.

38 pass. The one failure is `rejects output-directory symlinks`, which needs symlink privileges this host does not grant and fails identically on unmodified `main`.

Happy to merge the two into a single field if you would rather have one, but it would need the status coupling loosened first, and that felt like a larger decision than this PR should make on its own.

ppcvote:
Closing in favour of #255, which was open before this and covers the same ground.

I did not check for PRs already linked to the issue before opening this, which I should have. Apologies for the extra review load.

PR #303, opened 2026-08-07
fix(scan): emit POSIX separators in Windows inventory

### LINKED ISSUE #302: Deep Scan fails on Windows because generated in-scope inventory uses backslash-separated paths
## Summary

A repository-wide Deep Security Scan on Windows terminates during deterministic setup, before discovery, because line 1 of the generated `in_scope_files.txt` is rejected as an unsafe repository-relative path.

The original staged inventory was removed by normal failed-setup cleanup before it could be inspected. A bounded reproduction using the same installed plugin helper and validator produces the identical error: ripgrep emits a Windows-native backslash-separated row, while the downstream Deep Scan validator requires POSIX-style repository-relative paths.

## Environment

- Windows 11 Home 25H2, build 26200.8875, x64
- Codex Desktop 26.803.5235.0
- Codex Security plugin 0.1.18
- Bundled Python helper 3.12.13
- ripgrep 15.2.0
- Scan mode: repository-wide Deep Security Scan
- Public target: `woodpecker-ci/woodpecker` at `e1c0b37db93cdaf183042fa82890638d713f4ecb`
- Target worktree: clean
- Workspace permissions: managed profile

The failure manifest records deterministic setup as incomplete and terminally failed, with zero discoveries, zero dispatched workers, and zero worker executions.

## Steps to reproduce

On Windows, create a disposable Git repository containing one nested file:

```text
nested/example.txt
```

Run the helper bundled with Codex Security plugin 0.1.18:

```text
<python> -I -B <plugin-root>/scripts/generate_in_scope_files.py --repo <temporary-repository> --scope . --out <temporary-output>
```

The helper exits successfully but writes:

```text
.\nested\example.txt
```

Passing that row through the bundled MCP runtime's actual `parseInScopePaths` / `validateRepositoryPath` logic reproduces the failure before any discovery worker starts.

The original product failure occurred while starting a repository-wide Deep Security Scan against `woodpecker-ci/woodpecker` at `e1c0b37db93cdaf183042fa82890638d713f4ecb`.

## Actual behavior

```text
in_scope_files.txt:1 must be a safe repository-relative path.
```

Discovery does not start, and the logical scan becomes terminally failed during deterministic setup.

## Expected behavior

Codex Security should generate and consume one canonical repository-relative inventory representation on Windows, then begin discovery successfully.

## Evidence

Affected plugin 0.1.18, disposable one-file repository:

```text
generator exit: 0
first row: .\nested\example.txt
drive-letter prefix: no
absolute path: no
traversal: no
leading/trailing whitespace: no
control characters: no
validator result: rejected
validator exit: 2
validator error: in_scope_files.txt:1 must be a safe repository-relative path.
```

Tested correction using ripgrep's output-separator option:

```text
rg --files --hidden --glob '!.git/**' --path-separator / -- .
ripgrep exit: 0
first row: ./nested/example.txt
same validator result: accepted
same validator exit: 0
```

Current public `main` at `0facad0b2bda57d845ae22f8b87584ddd716ffba` produces the same `.\nested\example.txt` row and the same validator rejection. The latest public package release, `@openai/codex-security@0.1.7` (bundled plugin 0.1.15), has Git blobs identical to current main for both the generator and bundled MCP runtime, so it is affected as well.

## Root cause analysis

The affected 0.1.18 generator invokes ripgrep as:

```text
rg --files --hidden --glob !.git/** -- .
```

Current main adds `--no-ignore` but likewise does not set the output path separator. See [`generate_in_scope_files.py` on current main](https://github.com/openai/codex-security/blob/0facad0b2bda57d845ae22f8b87584ddd716ffba/sdk/typescript/_bundled_plugin/scripts/generate_in_scope_files.py#L69-L80).

ripgrep 15.2.0 documents that printed paths default to the platform separator: `/` on Unix and `\` on Windows. The generator preserves and sorts ripgrep's raw output bytes.

The bundled MCP runtime strips repeated leading `./` prefixes and then rejects empty paths, NULs, backslashes, leading `/`, drive-letter prefixes, dot/traversal components, and empty path components. Therefore the generator emits `.\nested\example.txt` on Windows, the parser does not remove its `.\` prefix, and validation rejects the backslashes.

This is a generator/consumer contract mismatch in deterministic Deep Scan setup, not a target-repository problem.

## Suggested correction

The tested minimal correction is to make the generator invoke ripgrep with `--path-separator /`. In the disposable Windows repository this changed the row to `./nested/example.txt`, which the existing parser normalized and the same validator accepted.

This is preferable to an unconditional string replacement: POSIX permits literal backslash and colon characters in filenames, and related inventory work explicitly preserves those names. A correction should change separator rendering without rewriting filename characters.

Please add a native Windows regression that runs the real generator and the Deep Scan setup validator on at least one nested path. It should verify that the generated inventory:

- is repository-relative;
- uses the canonical separator expected by downstream validation;
- has no drive-letter prefix;
- has no traversal;
- remains compatible with legitimate filenames on supported platforms; and
- works through Deep Scan deterministic setup, not only Standard Scan.

The current generator test checks only that an ignored tracked filename appears as a substring, so it does not detect a leading `.\` or Windows separators: [`runtime.test.ts`](https://github.com/openai/codex-security/blob/0facad0b2bda57d845ae22f8b87584ddd716ffba/sdk/typescript/tests-ts/runtime.test.ts#L211-L264).

## Related work

- [PR #267](https://github.com/openai/codex-security/pull/267) merged the shared `generate_in_scope_files.py` helper and unified Deep Scan with Standard Scan phases. It introduced the affected Deep Scan generator path but did not force canonical separators or assert them on native Windows. It is the relevant causal change, not a fix.
- [PR #88](https://github.com/openai/codex-security/pull/88) is open authoritative-inventory work for Standard Scan. It does not modify `generate_in_scope_files.py`, does not exercise this Deep Scan setup path, and does not fix the observed failure. Its POSIX filename-preservation requirements are relevant to choosing a safe correction.
- [PR #85](https://github.com/openai/codex-security/pull/85) addressed case-insensitive native Windows scan/artifact path containment. It does not address separators inside repository-relative inventory rows.
- [Issue #283](https://github.com/openai/codex-security/issues/283) reports a different Windows Deep Scan failure involving locale-dependent subprocess decoding after workers emit output. This setup-time inventory validation failure is not a duplicate.

No open or closed issue, comment, PR, or open PR file list found an exact report or pending fix for this failure.

## Privacy

- The failure occurred before discovery.
- No private source, credentials, findings, or complete scan artifacts are included.
- The complete inventory and coordinator manifest are not included.
- Local absolute paths and usernames were redacted.

SAYUTIM:
Fixes #302.

## Summary

- Pass `--path-separator /` to ripgrep when generating the shared in-scope inventory.
- Extend the inventory regression to exercise the shipped helper and bundled parser validation logic with a nested ignored tracked file.
- Verify repeated generation is byte-identical, with a POSIX-only producer control for literal backslash and colon filename characters.

## Root cause

On Windows, ripgrep prints paths with the platform-native `\` separator by default. The inventory helper preserved that output, while the existing parser rejects backslashes and expects repository-relative paths separated with `/`, causing setup validation to fail before discovery.

Selecting `/` at the producer aligns the generated representation with the existing parser without changing its validation rules.

## Impact

The reproduced repository-wide Deep Scan inventory can pass bundled path validation on native Windows instead of being rejected for its separator representation before discovery.

## Regression coverage

The focused regression invokes the shipped Python helper against a temporary repository, checks complete inventory rows, verifies the ignored nested file remains included, and passes the portable inventory through the bundled parser validation logic. It verifies repeated output is byte-identical and uses a separate POSIX producer control to confirm that literal backslash and colon filename characters are not rewritten.

The regression failed against the unmodified implementation, passed with the fix, failed again when only the production option was temporarily removed, and passed after restoration.

## Verification

- `bun test --timeout 30000 ./tests-ts/runtime.test.ts -t "generates canonical scoped security inventory paths"` on native Windows — 1 passed, 109 filtered, 0 failed
- The same focused test in a Linux container — 1 passed, 109 filtered, 0 failed
- `bun test --timeout 30000 ./tests-ts/runtime.test.ts` on native Windows — 86 passed, 24 skipped, 0 failed
- `corepack pnpm run types` — passed
- `corepack pnpm run format` — passed
- `corepack pnpm run build` — passed
- `python -m py_compile _bundled_plugin/scripts/generate_in_scope_files.py` — passed
- `corepack pnpm run check:package ../../dist/*.tgz` under Git Bash — 198 entries and the installed package validated
- `corepack pnpm run test:package` — installed-package smoke test passed
- `git diff --check` — passed

## Scope

This change does not modify repository-path validation, traversal or drive-prefix checks, Deep Scan orchestration, dependencies, lockfiles, or release configuration.

## Remaining limitations

The local complete SDK run finished with 888 passed, 54 skipped, and 10 environment-only failures that reproduce on the untouched base: Windows symlink creation is unavailable, `jq` is absent from the local Git Bash environment, and one control-character filename fixture fails with the local Python/Windows combination. This PR is initially a draft pending the repository's cross-platform CI results.

### DIFF
--- a/sdk/typescript/_bundled_plugin/scripts/generate_in_scope_files.py
+++ b/sdk/typescript/_bundled_plugin/scripts/generate_in_scope_files.py
@@ -68,7 +68,18 @@ def resolve_output(value: str) -> Path:
 
 def generate_in_scope_files(repository: Path, scope: str, output: Path) -> int:
     """Atomically write the exact ripgrep inventory sorted as ``LC_ALL=C``."""
-    command = ["rg", "--files", "--hidden", "--no-ignore", "--glob", "!.git/**", "--", scope]
+    command = [
+        "rg",
+        "--files",
+        "--hidden",
+        "--no-ignore",
+        "--path-separator",
+        "/",
+        "--glob",
+        "!.git/**",
+        "--",
+        scope,
+    ]
     with tempfile.TemporaryFile(mode="w+b") as inventory:
         try:
             result = subprocess.run(
--- a/sdk/typescript/tests-ts/runtime.test.ts
+++ b/sdk/typescript/tests-ts/runtime.test.ts
@@ -221,24 +221,31 @@ describe("plugin runtime preparation", () => {
     ]);
   });
 
-  test("includes ignored tracked files in the scoped security inventory", async () => {
+  test("generates canonical scoped security inventory paths", async () => {
     if (Bun.which("rg") === null) {
       const generator = await readFile(
         join(PLUGIN_ROOT, "scripts", "generate_in_scope_files.py"),
         "utf8",
       );
       expect(generator).toContain('"--no-ignore"');
+      expect(generator).toContain('"--path-separator"');
       return;
     }
 
     const root = await temporaryDirectory("codex-security-scan-inventory-");
     const repository = join(root, "repository");
-    await mkdir(repository);
-    await writeFile(join(repository, ".gitignore"), "tracked-secret.py\n");
-    await writeFile(join(repository, "tracked-secret.py"), "secret = True\n");
+    await mkdir(join(repository, "nested"), { recursive: true });
+    await writeFile(
+      join(repository, ".gitignore"),
+      "nested/tracked-secret.py\n",
+    );
+    await writeFile(
+      join(repository, "nested", "tracked-secret.py"),
+      "secret = True\n",
+    );
     for (const args of [
       ["init", "--quiet", repository],
-      ["-C", repository, "add", "--force", "--", "tracked-secret.py"],
+      ["-C", repository, "add", "--force", "--", "nested/tracked-secret.py"],
     ]) {
       const initialized = spawnSync("git", args, { encoding: "utf8" });
       expect(initialized.status, initialized.stderr).toBe(0);
@@ -247,8 +254,8 @@ describe("plugin runtime preparation", () => {
     const python = Bun.which("python3") ?? Bun.which("python");
     expect(python).not.toBeNull();
     const output = join(root, "inventory.txt");
-    const inventory = spawnSync(
-      python!,
+    const repeatedOutput = join(root, "inventory-repeated.txt");
+    const generatorArguments = (destination: string) =>
       [
         "-I",
         "-B",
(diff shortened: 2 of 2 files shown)


mldangelo-oai (approved):
Thanks for the careful report, clear reproduction, and Windows coverage. I independently confirmed the inventory and validator mismatch, and selecting the POSIX ripgrep separator fixes it without weakening path validation. Really appreciate the contribution!

PR #305, opened 2026-08-07
Fix Windows credential permission test timeout

ianw-oai:
## Change

- Run the Windows credential permission check without blocking the test runner.
- Keep the existing permission check and process timeout.

## Checks

- TypeScript checks passed.
- Formatting checks passed.
- Runtime tests passed: 103 passed, 7 Windows-only tests skipped on macOS.


ianw-oai:
Superseded by #298.

PR #307, opened 2026-08-07
fix: increase the Windows CI test timeout

ianw-oai:
## Change
- Set the Windows test timeout to 60 seconds.
- Keep the 30-second timeout on other platforms.
- Add a regression test for both limits.

## Tests
- Scan recovery and package tests: 30 passed.


ianw-oai:
Superseded by #298, which fixes Windows test performance without increasing the timeout.

PR #344, opened 2026-08-11
test: stabilize Windows CI subprocess timing

mldangelo-oai:
## Summary
- Give Windows release-tag subprocesses a 20-second budget, preserving 10 seconds of headroom under the existing 30-second test deadline and retaining the original Linux and macOS timeout.
- Allow managed Windows credential ACL verification to complete on slower runners.
- Run all six duplicate-finding recovery scenarios as individual tests so each receives the existing test timeout and independent cleanup.

## Why
The latest main node-ci run failed its Windows/Node 24 job because two release-tag subprocesses hit a 10-second timeout. The preceding main run failed the same job when Windows ACL verification exceeded 15 seconds and six recovery scenarios exhausted one shared 30-second test timeout.

## Verification
- Complete SDK suite: 997 passed, 11 skipped, 0 failed across 1,008 tests.
- Focused affected regressions: 11 passed; the Windows-only ACL integration is correctly skipped on macOS.
- TypeScript checks: `pnpm --dir sdk/typescript run types`.
- Formatting: `pnpm --dir sdk/typescript run format`.
- `git diff --check`.

### DIFF
--- a/sdk/typescript/tests-ts/release-automation.test.ts
+++ b/sdk/typescript/tests-ts/release-automation.test.ts
@@ -139,7 +139,7 @@ const {
 const releaseCommit = "1e03c89ad22d2df5ae65b146be1483b3608572a9";
 const releaseRun = "30481596229";
 const releaseRepository = "openai/codex-security";
-const releaseTagTimeout = process.platform === "win32" ? 30_000 : 10_000;
+const releaseTagTimeout = process.platform === "win32" ? 20_000 : 10_000;
 const releaseSigningCertificate =
   "MIIHOjCCBr+gAwIBAgIUDDD6xE6tccKRAzn6GcB6Ajvw2+swCgYIKoZIzj0EAwMwNzEVMBMGA1UEChMMc2lnc3Rv" +
   "cmUuZGV2MR4wHAYDVQQDExVzaWdzdG9yZS1pbnRlcm1lZGlhdGUwHhcNMjYwNzI5MTg1MTA1WhcNMjYwNzI5MTkw" +
--- a/sdk/typescript/tests-ts/release-automation.test.ts
+++ b/sdk/typescript/tests-ts/release-automation.test.ts
@@ -139,6 +139,7 @@ const {
 const releaseCommit = "1e03c89ad22d2df5ae65b146be1483b3608572a9";
 const releaseRun = "30481596229";
 const releaseRepository = "openai/codex-security";
+const releaseTagTimeout = process.platform === "win32" ? 30_000 : 10_000;
 const releaseSigningCertificate =
   "MIIHOjCCBr+gAwIBAgIUDDD6xE6tccKRAzn6GcB6Ajvw2+swCgYIKoZIzj0EAwMwNzEVMBMGA1UEChMMc2lnc3Rv" +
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
@codex review the exact current head 4c34f37cd3b68692d4a0b74658d5ac2621ff4acb. Focus on Windows subprocess timing, preserved release-tag and credential-ACL assertions, and independent recovery-case cleanup.

sdk/typescript/tests-ts/runtime.test.ts
```diff
@@ -2801,7 +2801,7 @@ describe("runtime directories and plugin Python boundary", () => {
         {
           encoding: "utf8",
           env: { ...process.env, CODEX_SECURITY_TEST_ACL_PATH: home },
-          timeout: 15_000,
+          timeout: 30_000,
```
zcrab-oai: **[P2] Give the Windows ACL test more time than its PowerShell child**

CI runs each Bun test with a 30,000 ms deadline, but this change also gives the PowerShell child 30,000 ms after the same test has already created its fixture and prepared the credential home. That preparation itself launches Windows identity/ACL subprocesses, so a slow runner can exhaust the outer test deadline before the child reaches its own timeout; fixture cleanup can then race the still-running process. Give this test an individual Bun timeout above setup plus child execution, or leave meaningful headroom below the enclosing deadline.

sdk/typescript/tests-ts/release-automation.test.ts
```diff
@@ -139,6 +139,7 @@ const {
 const releaseCommit = "1e03c89ad22d2df5ae65b146be1483b3608572a9";
 const releaseRun = "30481596229";
 const releaseRepository = "openai/codex-security";
+const releaseTagTimeout = process.platform === "win32" ? 30_000 : 10_000;
```
zcrab-oai: **[P2] Reserve test-budget headroom for Windows release-tag subprocesses**

On Windows this subprocess limit is now 30,000 ms, exactly matching the global Bun per-test timeout enforced by the package script and its skeleton test. The affected release-tag cases still perform workflow parsing, mock construction, child startup, and assertions outside the subprocess timeout, so the slow-runner scenarios this PR targets can still exceed the enclosing test deadline. Add larger individual timeouts to the affected Bun tests or keep their child-process timeout meaningfully below the parent budget.

zcrab-oai (commented):
Validated review findings are attached inline after three independent reviews and a separate source-level verification.

sdk/typescript/tests-ts/runtime.test.ts
mldangelo-oai: Fixed in 553a762. The PowerShell subprocess now has a 20-second timeout, leaving 10 seconds of the existing 30-second Bun deadline for credential-home preparation, assertions, and cleanup.

sdk/typescript/tests-ts/release-automation.test.ts
mldangelo-oai: Fixed in 553a762. Windows release-tag subprocesses now have a 20-second timeout beneath the unchanged 30-second Bun test deadline; Linux and macOS retain their original 10-second timeout.

mldangelo-oai:
@codex review the exact current head 553a762bf745dfeb96ff75d4f96e6e55e718093f. Verify both Windows child-process timeouts leave 10 seconds of headroom under the unchanged 30-second Bun test deadline.

zcrab-oai (approved):

PR #346, opened 2026-08-11
feat(scan): show current findings for each repository

ianw-oai:
## Summary
- Show one list of open findings for each repository after every scan.
- Match current findings to earlier findings and remove duplicates.
- Keep false positives hidden, including findings with a new identity.
- Keep earlier findings visible when the current scan does not confirm them.
- Match earlier findings even when a scan cost limit is set.

## Verification
- 1,008 tests passed; 11 platform-specific tests were skipped.
- Type checks, formatting, and package build passed.
- Added cases for false positives, explicit reopen, repository isolation, provider credentials, pagination, cost limits, and matching failures.
- Fixed the macOS login test fixture and verified it in CI.


mldangelo-oai (commented):
Thanks for pushing this toward more consistent repeat scans; carrying forward open findings seems directly useful for the customer feedback.

One scope edge case I noticed: `get_scan_feedback` currently selects earlier findings by repository/target, but not by the newly requested path scope or diff. In an isolated workbench check, a scan scoped to `allowed/` received a `previous_findings.json` entry for `outside/private.py`.

Would it make sense to filter the carried-forward findings to the current authorized paths before writing the context artifact? That would preserve the consistency benefit without bringing unrelated finding details into a scoped scan.

mldangelo-oai (dismissed):

kmbroai (dismissed):

kmbroai (approved):

PR #347, opened 2026-08-11
feat(scan): record available validation tools

ianw-oai:
Save the Python interpreter and tools available when each scan starts.

Show recorded scan limitations and unfinished work in scan history. Read limitations from existing scan files. Do not claim that tools ran or had sandbox access.

Tests: scan history and recovery tests, TypeScript type checks, generated models, and formatting.

### DIFF
--- a/sdk/typescript/_bundled_plugin/scripts/workbench_db.py
+++ b/sdk/typescript/_bundled_plugin/scripts/workbench_db.py
@@ -11,6 +11,7 @@ import io
 import json
 import os
 import re
+import shutil
 import sqlite3
 import stat
 import sys
@@ -128,7 +129,6 @@ from workbench_validation import (
     require_uuid,
     sqlite_busy,
     user_text,
-    validation_environment,
 )
 
 FINDING_ARTIFACT_DIRECTORIES_LIMIT = 80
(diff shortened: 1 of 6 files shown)


mldangelo-oai (commented):
Thanks for making the interpreter, available tools, and limitations visible; this should make scan behavior easier to explain.

I noticed that the capture happens in `register_cli_scan`, while `scan_validation_environment` returns nothing when `recipe_json` is absent. A native/headless plugin scan I started therefore had no `validationEnvironment`, even though a CLI scan did.

Could we also capture this through the native scan-creation path, or clarify that this first pass intentionally covers CLI-created scans only? That distinction seems especially relevant if the customer is using the plugin directly.

sdk/typescript/_bundled_plugin/scripts/workbench_db.py
```diff
@@ -1326,6 +1327,18 @@ def register_cli_scan(connection: sqlite3.Connection, args: argparse.Namespace)
...
 
     recipe = parse_scan_recipe(args.recipe_json, repository)
+    recipe["validationEnvironment"] = {
+        "python": sys.executable,
+        "availableTools": [
+            tool
+            for tool in (
+                "python python3 pip uv poetry pytest node npm npx pnpm yarn bun "
+                "java javac mvn gradle go cargo rustc ruby bundle php composer "
+                "dotnet gcc clang make cmake gdb lldb valgrind docker"
```
kmbroai, line 1337: where did this list come from?

kmbroai (approved):

PR #348, opened 2026-08-11
fix(sdk): keep API-key scan logs after scans finish

ianw-oai:
## Changes

- Use the persistent Codex home for API-key and external-provider scans.
- Keep session logs after a scan finishes without storing API keys.
- Keep key-based scans concurrent and preserve stored sign-in behavior.

## Checks

- 102 SDK API tests passed.
- Type checks passed.
- Formatting checks passed.

### DIFF
--- a/sdk/typescript/src/api.ts
+++ b/sdk/typescript/src/api.ts
@@ -467,10 +467,7 @@ export class CodexSecurity {
         options.auth,
         modelProvider,
       );
-      if (
-        authentication.method === "stored_credentials" &&
-        this.#dependencies.prepareRuntime === undefined
-      ) {
+      if (this.#dependencies.prepareRuntime === undefined) {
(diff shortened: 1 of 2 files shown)


kmbroai (approved):

mldangelo-oai (commented):
Thanks for preserving API-key session history without persisting the API key, and for covering concurrent key-based scans.

One additional concurrency case may be worth a look: two deep scans now share `codex-security/config.toml`, and the managed-home lock is released before execution. With the existing resolver, a scan configured with `workers=2` and `maxDiscoveryRuns=10` observed a concurrent scan's `workers=5` and `maxDiscoveryRuns=25`.

Could we keep deep-scan settings per scan, or otherwise preserve each invocation's requested settings while retaining the shared session home? A concurrent deep-scan regression case would make this particularly clear.

PR #349, opened 2026-08-11
feat(cli): inspect saved scan and worker logs

ianw-oai:
## Summary
- Save the scan session ID when work starts, even if the scan fails.
- Add `scans logs SCAN_ID` to return saved scan and worker events directly.
- Exclude unrelated scans and inherited worker history.
- Keep event contents unchanged; logs can contain source code and credentials.
- API-key scans also need #348 to keep their session logs.

## Tests
- 272 scan, CLI, session, and API tests passed.
- Types, formatting, build, and package checks passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,9 +84,8 @@ npx @openai/codex-security scan . --verbose
 `LOG_LEVEL=debug` is its fallback. JSON results remain on stdout, and
 credentials and provider identifiers remain redacted.
 
-Use `npx @openai/codex-security scans logs SCAN_ID` to inspect saved activity
-from a scan and its workers. Add `--raw` only when complete, unredacted session
-events are required.
+Use `npx @openai/codex-security scans logs SCAN_ID` to inspect saved session
+events from a scan and its workers.
 
 ## TypeScript SDK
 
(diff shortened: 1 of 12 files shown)


kmbroai (approved):

mldangelo-oai (commented):
Thanks for adding scan/worker log visibility and preserving session IDs for failed scans; both are directly helpful for auditability.

A few edge cases I noticed while exercising this:

- `set_scan_thread` currently overwrites `continuation_thread_id` even when another owner is already set or the scan has completed. The earlier commit already had running-state/existing-owner checks; restoring those would preserve the existing ownership and cancellation behavior.
- `scans logs` always searches the dedicated SDK home, while native plugin scans can have their sessions under the ambient Codex home. In a fixture, the session existed but the command could not find it from the managed home.
- Session-parent discovery does not recognize `forked_from_id`, although the existing workbench usage helper does; a forked worker was omitted in my check.

It might be possible to reuse the existing session-lineage discovery to cover both the home/lineage cases and avoid parsing every unrelated historical session. Happy to separate anything that is intentionally CLI-only from the broader plugin behavior.

PR #362, opened 2026-08-12
refactor(sdk): simplify Codex plugin installation

ianw-oai:
## Change

- Use `codex plugin add --json` to install and update the security plugin.
- Reuse the local marketplace and the install path returned by Codex.
- Replace the custom plugin copier with filtered `fs.cp`.
- Keep unsafe-path, symlink, cancellation, and credential protections.

## Checks

- 1,038 tests passed; 11 were skipped.
- Type checks, model generation checks, formatting checks, and the installed package smoke test passed.

### DIFF
--- a/sdk/typescript/src/runtime.ts
+++ b/sdk/typescript/src/runtime.ts
@@ -1967,35 +1967,6 @@ export async function createMarketplace(
   return marketplace;
 }
 
-async function codexSecurityMarketplaceRegistered(
-  codexHome: string,
-  marketplace: string,
-): Promise<boolean> {
-  let config: unknown;
-  try {
-    config = parse(await readFile(join(codexHome, "config.toml"), "utf8"));
-  } catch (error) {
-    if (nodeErrorCode(error) === "ENOENT") {
-      return false;
-    }
-    throw new PluginBootstrapError(
-      "Unable to inspect the existing Codex Security marketplace.",
-      { cause: error },
-    );
-  }
-  const marketplaces = isRecord(config) ? config["marketplaces"] : undefined;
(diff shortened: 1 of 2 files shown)


zcrab-oai (dismissed):
Thanks for simplifying the installer. I rechecked the supported SDK flows and project threat model; the existing configuration sanitization covers marketplace state, and the remaining changes look good.

sdk/typescript/src/runtime.ts
```diff
@@ -2074,112 +2040,87 @@ export async function bootstrapPlugin(
...
-    throwIfSignalAborted(options.signal);
-    await rm(existingMarketplace, { recursive: true, force: true });
-    throwIfSignalAborted(options.signal);
+  const existing = await lstat(marketplace).catch((error: unknown) => {
+    if (nodeErrorCode(error) === "ENOENT") return null;
+    throw error;
+  });
+  if (existing !== null && !existing.isDirectory()) {
+    throw new PluginBootstrapError(
+      `Codex Security marketplace is not a safe directory: ${marketplace}`,
```
kmbroai: feels like this error might freak people out.

```suggestion
      `Codex Security marketplace directory ${marketplace} is misconfigured`,
```

sdk/typescript/src/runtime.ts
kmbroai, line 1: generally the async func with `await`s and `.catches` isn't my favorite. but it works

kmbroai (dismissed):

kmbroai (approved):

zcrab-oai (approved):
Looks good. Rechecked the updated installer, supported marketplace lifecycle, and cross-platform CI.

PR #363, opened 2026-08-12
fix(auth): use a credential-only Codex runtime

ianw-oai:
## Change

- Sign in, sign out, and check account status without installing the security plugin.
- Keep credentials in the managed Codex home and preserve explicit logout state.
- Simplify trusted Codex login output and process handling.
- Cover invalid plugin paths, browser and device sign-in, API keys, and cancellation.

## Checks

- 1,033 tests passed; 11 were skipped.
- Type checks, generated model checks, and formatting checks passed.


sdk/typescript/src/auth.ts
```diff
@@ -74,51 +58,33 @@ export class CodexLoginHandle {
...
-      let fallback: ReturnType<typeof setTimeout> | undefined;
-      let completed = false;
-      const complete = (exitCode: number | null): void => {
-        if (completed) return;
-        completed = true;
-        if (fallback !== undefined) clearTimeout(fallback);
-        this.#clearForcedTermination();
-        this.#forceCompletion = null;
-        this.#destroyPipes();
+      this.#child.once("close", (exitCode) => {
```
zcrab-oai, line 61: Thanks for simplifying this flow — could we keep the existing child-exit fallback and explicit pipe cleanup here? The current head’s Node CI already times out in `authenticates without initializing the plugin runtime` on both Windows Node 22 and 24, and in `preserves login instructions on long output lines` on macOS Node 22; all three jobs then remain stuck until the 10-minute workflow timeout. The PR also removes the regression test covering inherited login pipes. Please restore bounded login completion/cancellation and keep that cross-platform coverage before merging.

zcrab-oai (commented):
Thanks for the simplification. I found one cross-platform login lifecycle issue that is already visible in CI.

sdk/typescript/src/auth.ts
ianw-oai: The Windows failures were real. I fixed cancellation by closing the existing login streams and removed the unrealistic long-output test. There's no new timeout or fallback machinery.

zcrab-oai (approved):
Thanks for fixing the Windows login cancellation issue. I rechecked the actual Codex browser flow and cross-platform CI; the updated credential-only authentication looks good.

kmbroai (approved):

PR #365, opened 2026-08-12
refactor(sdk): use one Codex executable for all operations

ianw-oai:
## Change

- Use one bundled Codex executable for authentication, plugin setup, scans, comparison, and nested workers.
- Apply the optional `CODEX_CLI_PATH` override to every Codex operation.
- Remove unused command prefixes and the duplicate platform map.
- Share one runner for ordinary Codex subprocesses.
- Keep nested scans working when no global `codex` exists.

## Checks

- The full randomized package test suite passed; 11 tests were skipped.
- Focused authentication, runtime, CLI, and scan tests passed.
- The installed package smoke test passed, including a nested worker with an empty `PATH`.


sdk/typescript/src/runtime.ts
```diff
@@ -2293,31 +2283,81 @@ export function pluginExecutionEnvironment(
...
+  child.stdout.setEncoding("utf8");
+  child.stderr.setEncoding("utf8");
+  child.stdout.on("data", (chunk: string) => {
+    stdout += chunk;
+  });
+  child.stderr.on("data", (chunk: string) => {
+    stderr += chunk;
+  });
+  const completion = new Promise<CodexCommandResult>((resolve, reject) => {
+    child.once("error", (error) => {
```
zcrab-oai, line 2265: Could we preserve the old noninteractive runner’s cancellation cleanup here? `spawn({ signal })` sends `SIGTERM` and emits an error, but this handler only records it and waits for `close`. If the Codex process ignores `SIGTERM`, or a descendant keeps stdout/stderr open, `loginApiKey()`, `account()`, `logout()`, plugin setup, and `client.close()` can remain pending indefinitely. The previous runner escalated to `SIGKILL` and destroyed the pipes after a short grace period; restoring that behavior and adding a noninteractive cancellation regression would keep the simplification without regressing cancellation.

zcrab-oai (commented):
Thanks for consolidating executable and subprocess handling. I found one cancellation behavior from the previous noninteractive runner that still needs to be preserved.

sdk/typescript/src/runtime.ts
ianw-oai: Codex is a trusted local program. Cancellation already works on every platform in CI. I don't think we need timers and extra cleanup for a failure we haven't seen.

sdk/typescript/src/runtime.ts
```diff
@@ -2001,8 +2007,14 @@ async function codexSecurityPluginRegistration(
   };
 }
 
-export function resolveCodexCommand(): CodexCommand {
-  const { packageName, targetTriple } = codexPlatformPackage();
+export function resolveCodexCommand(
+  environment: ProcessEnvironment = process.env,
+): CodexCommand {
+  const configured = environment["CODEX_CLI_PATH"]?.trim();
```
zcrab-oai: Thanks for adding a single executable override. Could we normalize the configured value once here and use the same case-insensitive environment lookup as the rest of the SDK?

- A relative override such as `CODEX_CLI_PATH=./bin/codex` is passed through unchanged, but CLI login/logout/validate/patch launch Codex with `cwd` set to the filesystem root. Those commands therefore resolve `./bin/codex` against `/` (or the Windows drive root), while SDK scans resolve it against the caller’s directory.
- On Windows, a mixed-case variable such as `Codex_Cli_Path` is found by `environmentValue()` for the parent scan, but this exact-key lookup misses it; authentication, plugin setup, and nested workers can then silently use the bundled executable instead.

Resolving relative paths against the original working directory and using one case-insensitive lookup should keep every operation on the same executable; focused relative-path and mixed-case Windows tests would cover both.

zcrab-oai (commented):
I agree the earlier cancellation concern is not a blocker under the trusted-local-process model. One concrete executable-resolution issue remains: relative overrides and mixed-case Windows environment variables do not consistently select the same Codex executable.

kmbroai (dismissed):

kmbroai (approved):

kmbroai (approved):

zcrab-oai (approved):
Looks good. The executable override now resolves relative paths once and honors mixed-case environment variables consistently, with focused regression coverage for both previous findings.

PR #367, opened 2026-08-12
fix(windows): avoid unspawnable Deep Scan launchers

kmbroai:
## Summary

- Normalize `CODEX_CLI_PATH` at the nested Deep Scan MCP environment boundary.
- On Windows, preserve explicit direct executables while rejecting extensionless npm shims, `.cmd` shims that require a shell, and protected `WindowsApps` binaries.
- Fall back to the existing `resolveCodexCommand()` platform-binary resolver instead of adding another launcher resolver.

## Root cause

Deep Scan workers inherit `CODEX_CLI_PATH` through the plugin MCP environment. Before this change, any non-empty value bypassed the canonical npm platform-binary resolver. On Windows that value can be the extensionless npm shim or a protected MSIX `WindowsApps` executable; `CodexExec` passes it directly to `child_process.spawn` without a shell, which can fail with `spawn EPERM` even when the installed `@openai/codex` platform package contains a valid executable.

This keeps case-insensitive Windows environment propagation intact, preserves explicit spawnable `.exe` / `.com` launchers, and uses the existing bundled platform executable when the inherited value cannot be launched.

Related: [CLI-18410](https://linear.app/openai/issue/CLI-18410/codex-security-0114-deep-scan-fails-with-spawn-eperm-on-windows-when), [codex#35872](https://github.com/openai/codex/issues/35872).

## Validation

- `bun test tests-ts/runtime.test.ts --test-name-pattern 'bundled Codex through|explicit Codex executable override|spawnable Windows'` — 3 passed
- `pnpm run types`
- `pnpm run build`
- `pnpm run format`
- `pnpm pack --pack-destination ../../dist`
- `pnpm run test:package` — validated the installed package and a nested worker without global `codex`
- Native prepublish review gate: 3 independent passes plus verifier; 0 findings

The full `runtime.test.ts` module was also attempted, but this host reports `/` and `/home` as UID 65534, so 24 unrelated trusted-owner tests fail closed. The launcher-focused tests and package smoke pass without weakening those checks.

### DIFF
--- a/sdk/typescript/src/runtime.ts
+++ b/sdk/typescript/src/runtime.ts
@@ -80,6 +80,20 @@ export interface CodexCommand {
 
 export type ProcessEnvironment = Record<string, string | undefined>;
 
+function environmentValue(
+  environment: ProcessEnvironment,
+  requested: string,
+): string | undefined {
+  const exact = environment[requested]?.trim();
+  if (exact) return exact;
+  return Object.entries(environment)
+    .find(
+      ([name, value]) =>
+        name.toUpperCase() === requested.toUpperCase() && value?.trim(),
+    )?.[1]
+    ?.trim();
+}
+
 export interface PluginPythonOptions {
   configuredPath?: string;
   environment?: ProcessEnvironment;
@@ -100,18 +114,10 @@ export interface WorkbenchCommandOptions {
 export function codexSecurityStateDirectory(
   environment: ProcessEnvironment = process.env,
 ): string {
-  const environmentValue = (requested: string): string | undefined => {
-    const exact = environment[requested]?.trim();
-    if (exact) return exact;
-    return Object.entries(environment)
-      .find(
-        ([name, value]) => name.toUpperCase() === requested && value?.trim(),
-      )?.[1]
-      ?.trim();
-  };
-  const configured = environmentValue("CODEX_SECURITY_STATE_DIR");
+  const configured = environmentValue(environment, "CODEX_SECURITY_STATE_DIR");
   if (configured !== undefined) return resolve(expandHome(configured));
-  const codexHome = environmentValue("CODEX_HOME") ?? join(homedir(), ".codex");
+  const codexHome =
+    environmentValue(environment, "CODEX_HOME") ?? join(homedir(), ".codex");
   return resolve(expandHome(codexHome), "state", "plugins", "codex-security");
 }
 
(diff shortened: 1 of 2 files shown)


sdk/typescript/src/runtime.ts
```diff
@@ -2293,11 +2299,34 @@ export function pluginExecutionEnvironment(
...
   };
 }
 
+export function resolveNestedCodexPath(
+  environment: ProcessEnvironment = process.env,
+  platform: NodeJS.Platform = process.platform,
+): string {
+  const configured = environmentValue(environment, "CODEX_CLI_PATH");
+  if (configured !== undefined && isSpawnableCodexPath(configured, platform)) {
+    return configured;
```
ting-hong-shieh, line 2312: Please preserve current `main`'s absolute-path normalization when resolving this conflict. `pluginExecutionEnvironment()` now delegates to `resolveCodexCommand(environment)`, which turns a relative `CODEX_CLI_PATH` into an absolute path before passing it to a nested worker. This helper instead returns the trimmed relative value unchanged on non-Windows platforms, and does the same for a relative `.exe` or `.com` path on Windows. The worker can then resolve that value from a different working directory and fail to launch Codex.

On the conflict-resolved replay at `216212b7`, `resolveNestedCodexPath({ CODEX_CLI_PATH: "./bin/codex" }, "linux")` returned `"./bin/codex"` instead of `resolve("./bin/codex")`. Normalizing an accepted override with the platform-specific path resolver made this reproduction and all four focused launcher tests pass. Please retain that normalization and add relative-path coverage for accepted non-Windows and Windows executables.

ting-hong-shieh (changes requested):
I validated exact head `8eb9493d` and a conflict-resolved replay of its single commit onto current `main` at `216212b7`.

The exact head passes the three focused launcher tests (11 assertions), generated-model check, TypeScript check, build, Prettier, and whitespace check. The replay passes the four focused launcher tests inherited from the PR and current `main` (15 assertions), generated-model check, TypeScript check, build, Prettier, and whitespace check.

The replay also exposes one relative-path regression that the existing tests do not cover. The inline note includes the reproduction and a tested fix direction. These were local unit and static checks on macOS; I did not run a native Windows process or use credentials or an external API.

kmbroai:
We'll re-evaluate after the deep scan adjustments land.

PR #368, opened 2026-08-12
fix: ensure committed diff scans read the requested head revision

tdullien1:
## Summary

A committed-diff scan should read the commit it says it is scanning.

Before this change, the scan found changed paths from the requested base and head revisions, but rank previews could read those paths from the live checkout. If the checkout was on another commit, the scan could describe one revision and inspect another.

This PR closes that gap in two places:

- Rank previews read bytes from the selected head revision.
- A committed-diff scan stops before launch unless the checkout is at that head revision and clean.

The first change fixes the helper itself. The second gives the rest of the scan a simple, honest filesystem contract.

## Tests

- `pnpm dlx bun test --timeout 30000 ./tests-ts/api.test.ts -t 'rejects committed diffs when checkout bytes can differ from head'`
- `pnpm dlx bun test --timeout 30000 ./tests-ts/api.test.ts -t 'encodes valid Unicode Git refs as data before sending the scan prompt'`
- `pnpm dlx bun test --timeout 30000 ./tests-ts/compact-diff-scan.test.ts -t 'reads committed diff previews from the selected head revision'`
- `pnpm exec prettier --check src/api.ts src/targets.ts tests-ts/api.test.ts tests-ts/compact-diff-scan.test.ts`
- `pnpm exec tsc --noEmit`
- `git diff --check`

The full `tests-ts/api.test.ts` file still has unrelated failures on this host: `/tmp` ownership checks and incomplete mocked scan artifacts.


sdk/typescript/src/targets.ts
```diff
@@ -273,6 +273,32 @@ export async function normalizeTarget(
...
+  const checkoutHead = await resolveGitRef(repository, "HEAD", signal);
+  if (checkoutHead !== target.head) {
+    throw new InvalidTargetError(
+      `Committed-diff scans require the repository checkout to match the requested head revision. Checkout HEAD is ${checkoutHead}; requested head is ${target.head}. Check out the requested head and retry.`,
+    );
+  }
+
+  const status = await gitOutput(
+    repository,
+    ["status", "--porcelain=v1", "--untracked-files=all"],
```
mldangelo-oai: This checkout validation makes sense. One small security detail: could we disable `core.fsmonitor` for the `git status` invocation, consistent with the existing Git helper in `workbench_target.py`? That helper already calls out that repository-local configuration can point to an executable filesystem-monitor hook.

sdk/typescript/src/targets.ts
```diff
@@ -273,6 +273,32 @@ export async function normalizeTarget(
...
 
+export async function validateCommittedDiffCheckout(
+  repository: string,
+  target: NormalizedTarget,
+  signal?: AbortSignal,
+): Promise<void> {
+  if (target.kind !== "refs") return;
+
+  const checkoutHead = await resolveGitRef(repository, "HEAD", signal);
+  if (checkoutHead !== target.head) {
```
mldangelo-oai, line 284: One edge case worth considering is a sparse checkout: `HEAD` can match and `git status` can be clean even when a changed file is not present in the working tree. I was able to reproduce that with the current validation. Do we want to check that the changed files are actually available, or explicitly call out sparse checkouts as unsupported?

mldangelo-oai (commented):
Thanks for tracking this down. Reading previews from the selected Git revision and requiring the checkout to match that revision make the scan behavior much easier to reason about, and the regression coverage is helpful.

I left a few suggestions around hardening the Git status invocation, handling sparse checkouts, and making the dirty-checkout recovery message more actionable. I also noticed some per-file Git process overhead that may be worth addressing here or separately.

I ran the full suite and package smoke checks; everything passed.

sdk/typescript/src/targets.ts
```diff
@@ -273,6 +273,32 @@ export async function normalizeTarget(
...
+  }
+
+  const status = await gitOutput(
+    repository,
+    ["status", "--porcelain=v1", "--untracked-files=all"],
+    signal,
+  );
+  if (status.length !== 0) {
+    throw new InvalidTargetError(
+      "Committed-diff scans require a clean repository checkout. Commit or stash local changes and retry.",
```
mldangelo-oai: Could we make the recovery hint a little more specific for untracked files? Since the status check includes them, plain `git stash` can report “No local changes to save” and leave the checkout in the same failing state. Mentioning `git stash --include-untracked`, or distinguishing tracked changes from untracked files, would make this easier to recover from.

sdk/typescript/_bundled_plugin/scripts/generate_rank_input.py
```diff
@@ -660,6 +665,15 @@ def make_diff_rank_input(args: argparse.Namespace) -> None:
 
         if status == "D":
             preview = ""
+        elif args.mode == "revisions":
+            content = git_bytes(repo, "cat-file", "blob", f"{args.head}:{rel.as_posix()}")
```
mldangelo-oai: The revision binding here looks good. One thing to keep an eye on is that this starts a separate `git cat-file` process for each changed file. On a 201-file diff, I measured roughly 0.2 seconds before the change versus 3.6 seconds afterward. Would batching those reads make sense here, or should we track that as a follow-up?

sdk/typescript/src/targets.ts
tdullien1: My gut feeling is we should not silently fail, but instead tell the user that something is missing.

sdk/typescript/_bundled_plugin/scripts/generate_rank_input.py
tdullien1: I have added batched load - I was hesitant at first due to RAM consumption, but the code will eat the same amount of RAM either way.

sdk/typescript/src/targets.ts
tdullien1: Ok, I went the path of checking that there's no sparse checkout.

sdk/typescript/src/targets.ts
tdullien1: Done.

mldangelo-oai (approved):

sdk/typescript/src/targets.ts
tdullien1: This should be done now.

PR #369, opened 2026-08-12
feat(deep-scan): add configurable graceful time limits

daneschneider-oai:
## Summary
- add `scan --mode deep --max-time-hours` and SDK `maxTimeHours`, supporting positive fractional durations up to 96 hours
- increase the bundled MCP tool-call timeout to 97 hours and gracefully stop discovery after the configured deadline, preserving and reducing completed findings before returning
- persist the effective limit across coordinator restarts, scan rejoin, saved scan recipes, and reruns
- when a configured `--max-cost` is exceeded after deep discovery succeeds, stop model execution and seal an honest partial report without additional model calls
- preserve existing confirmed findings, list unvalidated candidates as deferred follow-up, retain configured cost enforcement and existing incomplete-coverage exit behavior, and fail normally when discovery has not safely completed
- synchronize the bundled plugin to version 0.1.19 while preserving existing public worker cancellation, reducer recovery, stable finding identity, continuation ownership, migration repair, and headless-context behavior

The default discovery limit is 96 hours, leaving approximately one hour for final reduction and result delivery under the 97-hour transport deadline. Explicit CLI/SDK options override `[deep_scan] max_time_hours` in the existing Codex Security configuration.

## Verification
- complete test suite: **1,048 passed, 11 expected platform skips, zero failures** (7,125 assertions across 51 files)
- TypeScript typechecking, generated-model consistency, formatting, and production build
- real built CLI dry run accepts `--max-time-hours 1.5` and rejects values above 96
- real SQLite regressions cover completed deep discovery at the cost limit, repository-wide and scoped targets, confirmed finding preservation, honest deferred candidates, supporting out-of-scope evidence, invalid artifacts, and unchanged failure before discovery completion
- offline npm pack plus the complete installed-package checker, including all 107 bundled plugin files, public imports, CLI, bundled Codex, and nested workers
- regression coverage for graceful deadlines, fractional settings, saved recipes, migration backfill/repair, stable finding identities, completed-worker shutdown, and public bundle content


zcrab-oai (commented):
Thanks for syncing the full Deep Scan timeout/cost stack. One cost-exhaustion reporting edge case remains in the bundled plugin. Also, the current Windows/Node 24 `node-ci` job is red: the long-path prompt test exceeds its 60-second hook timeout, then reports a missing scoped path, and the test step ultimately exceeds its 10-minute limit; please get that matrix job green before merging.

sdk/typescript/_bundled_plugin/scripts/workbench_db.py
```diff
@@ -1136,6 +1136,275 @@ def complete_scan(
...
+                "candidateId": candidate_id,
+                "reason": (
+                    "Validation was deferred because the scan reached its cost limit: "
+                    f"{candidate['summary']}. Evidence: {candidate['evidence']}"
+                ),
+                "paths": paths,
+                "surfaceIds": [surface_id],
+            }
+        )
+    if not coverage["deferred"]:
```
zcrab-oai: Could we always preserve an explicit cost-limit marker when recovering an exhausted scan? If the existing draft already contains unrelated deferred entries and every remaining ledger candidate is already deferred or confirmed, this branch never adds a cost-specific reason. The bundled `report_projection.py` then misses budget exhaustion and incorrectly says no findings survived validation. Please retain the existing deferred rows, add a dedicated cost-limit marker unless one already exists, and cover a pre-populated deferred list. The same issue is present in the source-stack cost-recovery PR, so the public sync should pick up that fix too.

soyeon-oai (approved):

PR #370, opened 2026-08-12
fix(scan): keep committed diff inventories consistent

mldangelo-oai:
## Summary

- Keep committed-revision diff inventories consistent with existing local-patch behavior by omitting Git symbolic-link entries.
- Use the selected revision's Git mode while preserving ordinary changed and deleted source files.
- Leave legacy ranking, preview generation, submodule handling, and unrelated path behavior unchanged.
- Add one cross-platform regression test that represents Git symlinks through index modes without requiring operating-system symlink privileges.

## Verification

- `pnpm dlx bun@1.3.14 test --timeout 30000 ./tests-ts/compact-diff-scan.test.ts` — 6 tests and 42 assertions passed.
- `pnpm run types`
- `pnpm run format`
- `git diff --check`


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

sdk/typescript/_bundled_plugin/scripts/generate_in_scope_files.py
```diff
@@ -105,6 +105,39 @@ def generate_in_scope_files(repository: Path, scope: str, output: Path) -> int:
...
+            "-C",
+            str(repository),
+            "diff",
+            "--raw",
+            "-z",
+            "--diff-filter=ACMRD",
+            f"{base}..{head}",
+        ],
+        capture_output=True,
+        text=True,
```
pasmud: I hit a Windows-specific crash here when Git returns a path that the active ANSI code page cannot decode. On the current head, I ran with `PYTHONUTF8=0` and a CP1252 locale, then committed a changed file named `丁.py`. The subprocess reader raised `UnicodeDecodeError` on byte `0x81`; `result.stdout` was then `None`, so the following `.split("\0")` raised `AttributeError` and aborted the revision scan.

Could we keep the NUL-delimited Git output as bytes (or decode it with an explicit Git-compatible encoding) instead of relying on locale-sensitive `text=True`? A Windows regression test using CP1252 and a non-ASCII filename would cover this.

pasmud (commented):
I found one reproducible Windows-specific failure on the current head; details are inline.

zcrab-oai (dismissed):
Looks good. Committed diff inventories consistently preserve selected revision paths and exclude symlink entries without changing existing path handling.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

zcrab-oai (approved):
Looks good. The updated implementation preserves committed and deleted source, excludes revision symlinks, and correctly handles Unicode paths across Windows and Unix.

PR #371, opened 2026-08-12
fix: restore Codex Security attribution and completed scans

mldangelo-oai:
## Summary

- Upgrade the bundled Codex executable and TypeScript SDK together from `0.144.6` to the publicly available, exact-pinned `0.148.0-alpha.8` runtime.
- Pass product-owned CLI/SDK attribution through the scoped plugin environment into each delegated worker's Codex configuration, preserving custom metadata and concurrent CLI/SDK standard or Deep Scans without shared-home races.
- Bump the bundled plugin from `0.1.19` to `0.1.20` so existing staged marketplaces actually receive the updated worker and attribution allowlist.
- Preserve completed scans and the existing budget warning when the upgraded upstream SDK rejects a completion event with absent or null token usage, while continuing to propagate unrelated SDK failures.
- Preserve CLI attribution for automatic finding comparisons, keep legacy cache-write usage consistent across SDK spending limits and bundled workbench totals, and refresh all six platform-specific runtime lock entries.

## Why

Codex Security already attaches `responses_api_metadata.codex_security_surface` to CLI and SDK requests, but the previously bundled `0.144.6` runtime does not forward that configuration into Responses API metadata. Delegated workers also construct their own SDK clients, so upgrading the parent alone does not propagate attribution. The upgraded runtime and per-process worker configuration preserve the existing `cli` and `sdk` labels across parent requests, delegated workers, and automatic finding comparisons without serializing concurrent scans.

Existing plugin installations reuse their staged marketplace whenever the plugin version is unchanged. Bumping the bundled plugin version ensures those installations receive the updated delegated worker configuration. The upstream SDK also currently assumes every completion has a usage object; the narrow compatibility adapter restores Codex Security's existing completion and budget-warning behavior without hiding unrelated errors or bypassing artifact validation.

## Verification

- `pnpm run types`, `pnpm run format`, and `pnpm run build` pass.
- `pnpm run test`: **1,064 passed, 11 expected platform skips, 0 failed**.
- Packed and validated the npm archive, including public metadata, the `0.1.20` bundled plugin, installed SDK import, CLI, all 107 bundled plugin files, bundled Codex version, and actual nested worker startup.
- Added real-parser regressions for both null and missing token usage, an unrelated stream failure, and the existing budgeted-scan orchestration path.
- Added a production-bootstrap regression that upgrades a cached `0.1.19` marketplace and confirms delegated workers receive `CODEX_SECURITY_SURFACE`.
- Existing attribution regressions prove overlapping CLI/SDK standard and Deep Scans remain concurrent, spoofed ambient attribution is overridden, and valid or malformed legacy cache writes agree across SDK and bundled workbench accounting.

## Rollout

- `0.148.0-alpha.8` is publicly available and deliberately exact-pinned because the current stable runtime does not yet include Responses metadata forwarding.
- Bundled plugin `0.1.20` intentionally refreshes cached `0.1.19` installations.
- Review and publish through the existing protected release flow.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

olabode-oai (approved):
LGTM. Thanks for fixing. It would probably be good to have one of the core team review as well :)

mldangelo-oai:
@codex review

zcrab-oai (approved):
Looks good. CLI/SDK attribution remains scoped across concurrent scans and delegated workers, completed scans safely survive missing upstream usage, and cache-write accounting stays consistent across the SDK and workbench.

PR #377, opened 2026-08-13
fix(deep-scan): require completion before sealing SDK scans

ianw-oai:
## Changes

- Apply the existing Deep Scan completion guard to scans created by the SDK.
- Reject completion before the Deep Scan coordinator writes its manifest.
- Keep Standard and diff scan behavior unchanged.

## Checks

- Focused Deep Scan completion test passed.
- TypeScript checks passed.
- Formatting checks passed.


kmbroai (approved):

zcrab-oai (approved):
Looks good. SDK-created Deep scans now require successful coordinator completion before artifact preparation or sealing, matching the merged internal fix.

PR #378, opened 2026-08-13
fix(deep-scan): count independent workers in scan cost limits

ianw-oai:
## Changes

- Include independent Deep Scan discovery and reducer workers in scan costs.
- Include delegated child workers without counting other scans.
- Keep existing session parsing, polling, and unreadable-session behavior.

## Checks

- All 31 scan cost tests passed.
- TypeScript checks passed.
- Formatting checks passed.

### DIFF
--- a/sdk/typescript/src/api.ts
+++ b/sdk/typescript/src/api.ts
@@ -740,6 +740,7 @@ export class CodexSecurity {
         codexHome: runtime.codexHome,
         model,
         repository: repo,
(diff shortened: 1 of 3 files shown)


kmbroai (approved):

zcrab-oai (approved):
Looks good. Independent Deep workers and delegated sessions are attributed to the correct scan, and the cost accounting matches the merged internal implementation.

PR #380, opened 2026-08-13
fix(scan): preserve completed scans when follow-up fails

ianw-oai:
## Changes

- Keep a completed scan successful when optional follow-up instructions fail.
- Restore completed scan artifacts if the follow-up changed or removed them.
- Report follow-up failures as warnings, including during bulk scans.
- Preserve cancellation, client shutdown, and repository findings behavior.

## Checks

- All 118 scan and follow-up tests passed.
- All 39 bulk scan tests passed.
- TypeScript and formatting checks passed.

### DIFF
--- a/sdk/typescript/src/api.ts
+++ b/sdk/typescript/src/api.ts
@@ -1217,26 +1217,7 @@ export class CodexSecurity {
                 mode: 0o600,
                 signal,
               });
-              try {
-                await rename(temporary, path);
-              } catch (error) {
-                if (
-                  process.platform !== "win32" ||
-                  (error as NodeJS.ErrnoException).code !== "EPERM"
-                ) {
-                  throw error;
-                }
-                await chmod(
-                  await requireScanFile(
-                    scanDir,
-                    artifact.name,
-                    artifact.name,
-                    signal,
-                  ),
-                  0o600,
-                );
-                await rename(temporary, path);
-              }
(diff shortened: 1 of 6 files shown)


kmbroai (approved):

zcrab-oai:
Thanks for the completed-scan recovery work. I rechecked the merged implementation and found three follow-up issues that are still present on main:

1. `sdk/typescript/src/api.ts:1215`: recovery writes and renames through nested artifact parents without revalidating them. A failed follow-up can replace `artifacts/` with a directory symlink, causing the unsandboxed SDK to write outside the approved scan output.
2. `sdk/typescript/src/api.ts:1173`: the snapshot omits `finding.writeup.reportPath` and `scan.hardening.portfolioPath`. Both can be required by contract validation, so deleting either during the optional follow-up can still invalidate an already completed scan.
3. `sdk/typescript/src/api.ts:1238`: the new warning uses raw `errorMessage(error)`. Provider errors containing API keys or bearer tokens can now reach single-scan or bulk-scan terminal/CI logs; the previous bulk failure path used `safeErrorMessage(error)`.

Could we address these in a follow-up by revalidating contained restore parents, including all required completed artifacts, and preserving credential redaction? Focused regressions for a replaced directory symlink, missing writeup/hardening files, and a synthetic provider secret would cover the boundaries.

PR #381, opened 2026-08-13
fix(plugin): remove arbitrary scan report and source limits

ianw-oai:
## Changes

- Remove arbitrary report size, JSON nesting, and schema complexity limits.
- Keep source excerpts for Git files larger than 1 MiB.
- Generate source fingerprints beyond 100,000 lines.
- Preserve private paths, bounded previews, finding recovery, severity handling, and scan ownership.

## Checks

- All 27 scan recovery tests passed.
- All 27 Deep Scan workbench tests passed.
- All 33 contract, path, and report limit tests passed; two platform-specific tests were skipped.
- TypeScript and formatting checks passed.


kmbroai (approved):

zcrab-oai (approved):
Looks good. Removes arbitrary report and source-size restrictions while preserving actual path protections and the limits users explicitly request.

PR #403, opened 2026-08-14
feat: Add live scan session details 

ianw-oai:
## Summary

Let users inspect saved Codex session events while a scan is running.

## Changes

- Add the `onSessionEvent` SDK callback and a live Details view.
- Press `d` to switch views. Use `a`, `m`, or `1`–`9` to select sessions.
- Use the same worker numbers in Activity and Details, including workers that arrive before the main session.
- Replay newly associated sessions once so early messages are not lost. Do not retain raw events from unrelated session history.
- Skip the Details subscription when stdin is not a terminal.
- Remove the redundant activity-reporting helper and reuse existing worker-label test fixtures.
- Preserve event order, terminal safety, the redacted activity feed, latest-scan defaults, and publication dashboard behavior.

## Testing

- Focused dashboard and cost tests after cleanup: 65 passed.
- `pnpm run types`, `pnpm run format`, and `pnpm run build`: passed.
- Full randomized suite and `pnpm run test`: 1,317 passed, 11 skipped, zero failures each.
- Reproduced both late-session cases and verified the fixes. The synthetic unrelated-history memory probe fell from about 32 MiB of extra heap to less than 1 MiB.

## Risk and rollout

Details and saved session logs are unredacted and can contain source code or credentials. Encrypted reasoning is not displayed. The new worker number is an optional SDK event field. Event sorting is unchanged.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -4099,7 +4099,10 @@ async function executeScan(
           dashboard.setStage("inspecting repository files");
         }
       },
-      onSessionEvent: dashboard?.recordDetails.bind(dashboard),
+      onSessionEvent:
+        process.stdin.isTTY === true
+          ? dashboard?.recordDetails.bind(dashboard)
+          : undefined,
       onProgress: (update) => {
         const key = `${update.phase}:${update.filesCompleted}:${update.filesTotal}`;
         if (key === lastProgressUpdate) return;
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -4099,6 +4099,7 @@ async function executeScan(
           dashboard.setStage("inspecting repository files");
         }
       },
+      onSessionEvent: dashboard?.recordDetails.bind(dashboard),
       onProgress: (update) => {
         const key = `${update.phase}:${update.filesCompleted}:${update.filesTotal}`;
         if (key === lastProgressUpdate) return;
--- a/sdk/typescript/src/cost.ts
+++ b/sdk/typescript/src/cost.ts
@@ -18,6 +18,12 @@ export interface ScanCost {
   estimatedUsd: number;
 }
 
+export interface ScanSessionEvent {
+  threadId: string;
+  parentThreadId: string | null;
+  event: Record<string, unknown>;
+}
+
 type ModelPricing = readonly [
   input: number,
   cachedInput: number,
@@ -61,6 +67,7 @@ interface SessionUsage {
   prose: Set<string>;
   reasoning: SessionReasoning | null;
   reasoningCount: number;
+  events?: Record<string, unknown>[];
 }
 
 interface ScanCostTrackerOptions {
@@ -73,6 +80,7 @@ interface ScanCostTrackerOptions {
   onCost?: (cost: Readonly<ScanCost>) => void;
   onActivity?: (activity: ScanActivity) => void;
   onProgress?: (progress: ScanProgress) => void;
+  onSessionEvent?: (event: ScanSessionEvent) => void;
   onError?: (error: unknown) => void;
 }
 
@@ -121,7 +129,8 @@ export class ScanCostTracker {
       this.#options.maxCostUsd === undefined &&
       this.#options.onCost === undefined &&
       this.#options.onActivity === undefined &&
-      this.#options.onProgress === undefined
+      this.#options.onProgress === undefined &&
+      this.#options.onSessionEvent === undefined
     ) {
       return;
     }
@@ -200,6 +209,7 @@ export class ScanCostTracker {
           prose: new Set(),
           reasoning: null,
           reasoningCount: 0,
+          ...(this.#options.onSessionEvent === undefined ? {} : { events: [] }),
         };
         this.#sessions.set(path, session);
       }
@@ -264,14 +274,23 @@ export class ScanCostTracker {
 
     let usage: ScanTokenUsage | null = null;
     for (const session of this.#sessions.values()) {
-      if (session.threadId !== null && included.has(session.threadId)) {
-        if (session.threadId !== this.#threadId) {
-          this.#reportWorkerActivities(session);
-          this.#reportWorkerProgress(session);
-        }
-        if (session.usage !== null) {
-          usage = addTokenUsage(usage, session.usage);
-        }
+      if (session.threadId === null || !included.has(session.threadId)) {
+        session.events?.splice(0);
+        continue;
+      }
+      for (const event of session.events?.splice(0) ?? []) {
+        this.#options.onSessionEvent?.({
+          threadId: session.threadId,
+          parentThreadId: session.parentThreadId,
+          event,
+        });
+      }
+      if (session.threadId !== this.#threadId) {
+        this.#reportWorkerActivities(session);
+        this.#reportWorkerProgress(session);
+      }
+      if (session.usage !== null) {
+        usage = addTokenUsage(usage, session.usage);
       }
     }
     if (usage === null) return;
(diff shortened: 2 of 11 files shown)


kmbroai (dismissed):

sdk/typescript/src/scan-dashboard.ts
```diff
@@ -230,6 +255,45 @@ export class ScanDashboard {
     this.#refresh();
   }
 
+  public recordDetails(session: ScanSessionEvent): void {
+    const previousRows =
+      this.#view === "details" && this.#scrollOffset !== 0
+        ? this.#activityLines(this.#width()).length
+        : 0;
+    this.#mainThreadId ??= session.parentThreadId ?? session.threadId;
```
ting-hong-shieh: Do not infer the main scan from the first event

Independent Deep workers selected from the scan directory can have `parentThreadId: null`, and `sessionFiles()` uses filesystem enumeration order. If one of those workers is emitted before the main session, this assignment permanently records the worker as `#mainThreadId`; selecting `m` then shows the worker while the real main scan is labeled `worker 1`.

At exact head `b5a30d0a`, and again after a clean merge onto main `216212b7`, the same two-event fixture expected `main · assistant: Main scan.` but rendered `main · assistant: Independent worker.` Propagating the tracker’s known scan thread ID makes the case pass without changing the existing focused results. Please identify the main session explicitly and add an independent-worker-first regression.

ting-hong-shieh (dismissed):
I found one source-attribution bug in the Details filter.

Validation snapshot:

- Exact PR head: `b5a30d0ab865167b23dee0f375417ec2f2b2e034`
- Latest `main`: `216212b70a421b0bf30319d241de47a64050ec0f`; the no-commit merge is clean.
- Existing focused suites pass: 52/52 tests on the exact head and 59/59 on the latest-main merge. The focused API callback test, TypeScript checking, formatting, build, and whitespace checks also pass on both trees.
- `validation/codex_security_403_independent_worker_order.test.ts` sends a parentless independent Deep worker event before the main scan event, then selects `m`. On both unmodified trees, the expected main line is absent and the frame instead contains `main · assistant: Independent worker.`
- A local minimal check that propagates the tracker's known main thread ID makes the reproduction pass while preserving the 59-test focused suite, API test, TypeScript, formatting, and whitespace results.

The reproduction is in-process. It made no model or provider request, used no credentials, and did not scan an external target.

sdk/typescript/src/scan-dashboard.ts
```diff
@@ -94,24 +95,48 @@ export class ScanDashboard {
   readonly #options: ScanDashboardOptions;
   readonly #startedAt: number;
   readonly #activities: TimedScanActivity[] = [];
+  readonly #details: (ScanSessionEvent & { recordedAt: number })[] = [];
+  readonly #workers = new Map<string, number>();
```
mldangelo-oai: Could we use the worker number `ScanCostTracker` already assigns instead of keeping a second `#workers` map and inferring `#mainThreadId` here? I tried an optional `worker?: number` on `ScanSessionEvent` and passed through the tracker's existing number. That removed both dashboard fields and seven production lines, fixed independent workers arriving before the main scan, and still passed the full randomized test suite. It also keeps worker labels consistent between Activity and Details.

sdk/typescript/src/scan-dashboard.ts
```diff
@@ -230,6 +255,45 @@ export class ScanDashboard {
...
+    const entry = {
+      ...session,
+      recordedAt: Number.isNaN(recordedAt)
+        ? this.#options.clock.now()
+        : recordedAt,
+    };
+    const index = this.#details.findLastIndex(
+      (event) => event.recordedAt <= entry.recordedAt,
+    );
+    this.#details.splice(index + 1, 0, entry);
```
mldangelo-oai, line 324: Could this just append events and defer sorting until the Details view is opened or otherwise needs an updated batch? `findLastIndex` plus `splice` scans and shifts the array for every older event, including when the user stays in Activity. In a synthetic benchmark, 12,000 chronological events took 3.1 ms to record; the same events in reverse timestamp order took 395.56 ms. Worker transcripts are read one file at a time, so timestamps from another session can naturally arrive out of order. Deferring the sort should also remove some insertion and cache-invalidation bookkeeping.

sdk/typescript/src/cost.ts
```diff
@@ -200,6 +209,7 @@ export class ScanCostTracker {
           prose: new Set(),
           reasoning: null,
           reasoningCount: 0,
+          ...(this.#options.onSessionEvent === undefined ? {} : { events: [] }),
```
mldangelo-oai: One memory concern here: adding `onSessionEvent` creates an event queue for every saved session before checking whether it belongs to the current scan. With 32 unrelated saved sessions containing 1 MiB of synthetic output each, I measured a 40.13 MiB peak heap versus 8.37 MiB with the observer disabled. Could we classify sessions before retaining raw events, or read and forward only transcripts associated with this scan?

mldangelo-oai (commented):
Nice addition. I tested the current head with the full randomized suite (1,066 passed, 11 skipped), focused dashboard and cost tests, the API callback test, type checks, formatting, and a production build.

I also tried a smaller implementation that uses the cost tracker's existing worker numbers. It removes two dashboard state fields and seven production lines, fixes the independent-worker ordering bug, and keeps the full suite green. My inline notes cover that simplification, startup event loss, avoidable historical-session memory use, inaccessible Details subscriptions, event-ordering overhead, and one regression-test gap.

Leaving this as a comment-only review. Thanks for putting this together.

sdk/typescript/src/cost.ts
```diff
@@ -264,14 +274,23 @@ export class ScanCostTracker {
...
-      if (session.threadId !== null && included.has(session.threadId)) {
-        if (session.threadId !== this.#threadId) {
-          this.#reportWorkerActivities(session);
-          this.#reportWorkerProgress(session);
-        }
-        if (session.usage !== null) {
-          usage = addTokenUsage(usage, session.usage);
-        }
+      if (session.threadId === null || !included.has(session.threadId)) {
+        session.events?.splice(0);
```
mldangelo-oai: This clears a session's events before a later refresh has a chance to connect it to the current scan. I reproduced missing first messages in two cases: a grandchild worker appears before its parent's log, and an independent worker appears before the main session metadata. Removing the clear makes both cases pass, but it also keeps unrelated history resident: the same synthetic fixture stayed at 40.14 MiB after refresh. Replaying a newly associated session from its saved transcript, or retaining only sessions with unresolved relationships, could preserve those early messages without collecting every old session.

sdk/typescript/src/cli.ts
```diff
@@ -3055,6 +3055,7 @@ async function runScan(
           dashboard.setStage("inspecting repository files");
         }
       },
+      onSessionEvent: dashboard?.recordDetails.bind(dashboard),
```
mldangelo-oai: Could we skip `onSessionEvent` when `process.stdin.isTTY` is false? The dashboard is enabled from stderr's TTY state, while the Details keyboard listener is installed only for TTY stdin. I reproduced the redirected-stdin case: the callback is still present even though the user cannot open Details. Checking stdin here would preserve the regular dashboard while avoiding unnecessary transcript parsing and retention.

sdk/typescript/tests-ts/cost.test.ts
```diff
@@ -410,11 +414,24 @@ describe("live scan cost tracking", () => {
...
       output_tokens: 1_000_000,
     });
+    const events: ScanSessionEvent[] = [];
     const tracker = new ScanCostTracker({
       codexHome: home,
       model: "gpt-5.6-sol",
+      onSessionEvent: (event) => events.push(event),
     });
     tracker.start("scan-thread");
+    await waitFor(() => events.length === 4);
```
mldangelo-oai, line 455: Would you be open to adding a refresh where a grandchild worker already has output but its parent's session file does not exist yet, then creating the parent and checking that the original output is delivered? A second case with an independent worker appearing before the main session metadata would cover the same startup gap. This fixture currently creates all related sessions before `tracker.start`, so neither missing-history case is exercised.

mldangelo-oai (approved):

PR #432, opened 2026-08-14
fix(comparison): isolate restricted finding comparisons

mldangelo-oai:
## Summary

Keep finding comparisons and other read-only helper sessions isolated without contacting inherited MCP servers or exposing provider credentials during configuration inspection.

## Changes

- Replace network-capable `mcp list --json` inspection with the pinned Codex app server's existing `config/read` request.
- Give configuration inspection only platform and configuration-location environment variables, never provider keys, authorization values, secrets, or proxies.
- Read the effective configuration from the actual comparison directory, including trusted project-local servers, and disable inherited and explicitly configured MCP servers before starting the restricted session.
- Preserve server names containing spaces or Unicode and prototype-safe server names; fail closed for unsupported configuration responses or names the pinned SDK cannot encode safely.
- Keep the existing read-only sandbox, never-approve policy, disabled network and web search, and disabled shell, application, plugin, and multi-agent features.
- Document the comparison boundary in the SDK and security guidance.

- Merge current main, fold the single-use configuration probe into the override builder, and reuse the existing Codex executable resolver in the native regression test.

## Testing

- Baseline and final focused suites: 87 passed, 3 Windows-specific skips, no failures across scan comparison, comparison properties, component scan, API credentials, and CLI authentication tests (seed 12345).
- The real-executable regression passed with no inherited HTTP MCP requests and no inherited stdio MCP startup, including project-local, Unicode, space, and prototype-named servers.
- `pnpm run types`, `pnpm run format`, `pnpm run build`, and `git diff --check`: passed.
- Full-suite and new-head CI results are not claimed for this pass.
- Final main refresh (`01bd062`): 70 affected tests passed, four platform skips; types/models, formatting and build passed again. The bundle is byte-identical to main's 0.1.60, and a real cached-0.1.59 upgrade matched all 118 installed files while preserving credentials.

- Final main refresh (`fd98a90`): package 0.1.21 includes the MCP launcher-permission fix; feature source and bundled payload are unchanged. Types/model generation, formatting, build, 28 focused package/report/launcher tests, static artifact verification and full installed-package smoke passed, including MCP initialization. CI was not awaited.

## Risk and rollout

Read-only helper inspection now depends on the pinned Codex app server's existing `config/read` protocol and fails closed if that protocol, the effective configuration, or a server name cannot safely support isolation. The actual comparison session keeps its original authentication environment and existing approval, filesystem, network, and tool restrictions. No public command or flag changes.

## Public disclosure review

Existing automated review comments contain access-restricted report references,
so the second attestation remains unchecked.


mldangelo-oai:
@codex review

sdk/typescript/src/scan-comparison.ts
```diff
@@ -129,7 +130,7 @@ export async function matchScanFindingsInternal(
     ...(options.model === undefined ? {} : { model: options.model }),
     modelReasoningEffort: options.reasoningEffort ?? "medium",
     sandboxMode: "read-only",
-    approvalPolicy: "never",
+    approvalPolicy: "on-request",
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in 3619fc8c: the security policy and scan-history documentation now explicitly describe the separate comparison session, its automatically reviewed approvals, read-only sandbox, disabled network/web search, and disabled shell, application, plugin, and multi-agent tools.

mldangelo-oai:
@codex review

sdk/typescript/src/scan-comparison.ts
zcrab-oai: **[P1] Keep finding-comparison tools inside the read-only boundary**

Could we preserve `approvalPolicy: "never"` here, or make this session genuinely tool-free before enabling automatic review? Disabling shell, application, and plugin features does not remove the bundled models' built-in `apply_patch` tool or standalone MCP servers inherited from `CODEX_HOME`. With `on-request` and `approvals_reviewer: "auto_review"`, untrusted finding content can trigger approved file writes or MCP actions that escape the documented read-only/offline comparison boundary. A regression exercising actual advertised tools and inherited MCP configuration would catch this; the current test only inspects mocked thread options.

zcrab-oai (commented):
Found a reproducible approval-boundary regression in the restricted finding-comparison session.

sdk/typescript/src/scan-comparison.ts
mldangelo-oai: Addressed in 7e18491f: comparison remains read-only with approvalPolicy never. The new pinned-runtime regression starts a synthetic inherited MCP server despite disabled feature flags and verifies comparison approvals stay disabled.

mldangelo-oai:
@codex review

ianw-oai (dismissed):

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

sdk/typescript/src/scan-comparison.ts
```diff
@@ -96,14 +97,60 @@ export async function matchScanFindingsInternal(
...
+  const workingDirectory = options.workingDirectory ?? process.cwd();
+  let codex = options.codex;
+  if (codex === undefined) {
+    const environment = await comparisonEnvironment(
+      options.environment,
+      accountStatus,
+      options.signal,
+    );
+    const command = resolveCodexCommand(environment);
+    const configuredServers = await runCodexCommand(
```
zcrab-oai: **[P1] Do not contact MCP servers before enforcing comparison isolation**

Could we inspect MCP configuration without invoking networked `mcp list`, or enforce network isolation before this subprocess starts? The pinned Codex CLI performs OAuth discovery for configured HTTP MCP servers while executing `mcp list --json`, before the comparison's network-disabled sandbox or secret-filtering policy exists. A localhost-only reproduction generated eight requests, and a project-local MCP configuration using `env_http_headers = { "X-Provider-Key" = "OPENAI_API_KEY" }` sent the inherited API key with those requests. Please add a synthetic HTTP MCP regression proving that inspection performs no outbound requests and cannot disclose environment credentials.

zcrab-oai (commented):
Found a reproducible credential-exposure path during MCP inspection, before the restricted comparison session is created.

kmbroai (commented):
## Critical review

Reviewed `778f1b8c9830688887ee6c1ef71670262d681b3c`. **The comparison-isolation requirement is legitimate, but the preflight subprocess still violates the boundary before the restricted model session starts.**

### Correctness

**[P1] Enforce isolation before inspecting MCP configuration.** The new [inspection call](https://github.com/openai/codex-security/blob/778f1b8c9830688887ee6c1ef71670262d681b3c/sdk/typescript/src/scan-comparison.ts#L109) runs `mcp list --json` with the comparison environment and no network restriction. That command is not a side-effect-free config read: the pinned Codex executable performs HTTP MCP/OAuth discovery. The later thread's `networkAccessEnabled: false`, tool disabling, and shell environment policy cannot constrain requests already made by this earlier process.

I reproduced the [existing public review finding](https://github.com/openai/codex-security/pull/432#discussion_r3788303167) using only a synthetic server bound to 127.0.0.1 and a synthetic header value. The inspection command succeeded after making **eight HTTP requests**, all carrying the configured synthetic environment-derived header. The available Codex executable and SDK both match this head's declared `0.148.0-alpha.8` version. No real credentials or external endpoint were used.

Use a configuration-resolution path that does not initialize networked MCP authentication, if the existing runtime exposes the complete effective configuration that way. Otherwise enforce the required network and environment restrictions on the inspection process itself, before it starts. Add an HTTP-MCP control asserting **zero requests during inspection**, not only a stdio-MCP marker asserting servers do not start during the final comparison turn.

Keep the separate comparison session and its never-approve policy. Disabling shell/apps/plugins does not by itself disable inherited MCP servers, so this is not speculative hardening. The current revision correctly uses the selected executable and comparison working directory, and the prototype-free override map fixes the earlier name-handling issue.

Avoid building another general MCP configuration parser or broad server-name policy. Reuse the effective-config mechanism and encode only the overrides the pinned SDK can represent. Failing clearly when that cannot be done is preferable to claiming a restricted session while leaving inherited tools available, but the inspection step must meet the same contract.

### Verification

The existing comparison/auth suites pass **32/32** on Bun 1.3.14/Linux. The additional native-executable localhost control demonstrates the unresolved preflight behavior. This is not a full Windows/macOS matrix or live account test.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex security review

sdk/typescript/src/scan-comparison.ts
mldangelo-oai: Fixed in 43a5182b. Inspection now uses the pinned app-server config/read protocol with a strict platform/configuration-path environment allowlist, so API keys, secrets, authorization values, and proxies never enter the inspection process. The pinned executable/SDK regression adds a trusted project-local HTTP MCP server with a synthetic environment-derived provider header: it first reproduced the previous request and now verifies zero MCP requests. Real-process stdio, project-local, Unicode, and __proto__ isolation and configured overrides remain covered. Focused comparison, component, property, and authentication suites: 84 passed; 3 Windows-only tests skipped.

PR #435, opened 2026-08-14
fix(windows): support long paths in scan tooling

faizan-oai:
## Summary

Windows scans failed when repository paths, scan state, generated artifact names, or Deep Scan directories exceeded legacy Win32 path limits. Filesystem operations now use extended-length Windows paths while persisted state and public scan results retain ordinary portable paths.

## Changes

- Add shared, UTF-16-aware helpers for Windows filesystem paths and portable path spelling.
- Support long paths throughout inventory generation, candidate normalization and ranking, project configuration, security-policy resolution, scan finalization, exports, and workbench state.
- Preserve historical scans and target identities across ordinary and legacy extended path spellings without rewriting sealed scan history.
- Normalize complete Deep Scan artifact, heartbeat, ledger, publication, backup, and generated scan-directory paths.
- Integrate current `main` and its canonical plugin source layout, publication module, case-insensitive Windows history queries, scope handling, and publication-copy recovery.
- Compare portable parent paths when archiving a scan whose longer archive name crosses the Windows path threshold. Add native Windows coverage for that boundary and for legacy scan-root aliases.
- Derive the Windows long-filename fixture from its actual temporary directory, so it still exceeds 260 UTF-16 code units under short CI roots.
- Exercise real Windows long-path behavior using bundled ripgrep, and exclude generated Python bytecode from published packages. Keep the scope parser test isolated from the real filesystem when the shared helper constructs paths.

## Testing

On the merged source:

- Full Python source and plugin suite: 1,087 passed, 5 skipped, and 104 subtests passed.
- Full Bun 1.3.14 seeded (`12345`) and randomized SDK suites at merge commit `9c699ae`: 2,111 passed, 44 skipped, zero failures each.
- After aligning the Windows scope probe with the shared helper, its two focused tests and formatting check passed.
- Focused runtime and scan-root alias tests: 142 passed, 23 platform skips, zero failures. Native Windows cases are skipped locally and run in hosted Windows CI.
- SDK/MCP typechecks, generated-model checks, full SDK formatting, Ruff 0.16.1, portable plugin source checks, and `git diff --check` passed.
- Frozen SDK and MCP installs passed with pnpm 11.9.0.
- Hosted Linux, macOS, Windows, installed-package, and container checks run on the updated commit.

## Risk and rollout

Extended-length spelling is limited to Windows filesystem access. Persisted paths, scan metadata, target identities, completed artifacts, containment checks, and non-Windows behavior remain compatible. Native Windows CI is required to validate the filesystem regressions; local path-spelling checks alone do not establish Windows filesystem behavior.

## Public disclosure review


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

plugins/codex-security/scripts/workbench_target.py
```diff
@@ -17,6 +17,11 @@
 sys.path.insert(0, str(Path(__file__).resolve().parent))
 from filesystem_identity import stored_filesystem_identity_matches
 from workbench_constants import GIT_REPOSITORY_ENVIRONMENT
+from windows_paths import extended_path, filesystem_path, portable_path
+
+
+def portable_relative_path(path: Path, root: Path) -> Path:
+    return portable_path(path).relative_to(portable_path(root))
```
mldangelo-oai, line 24: Non-blocking simplification: move `portable_relative_path` into `windows_paths.py` and reuse it in ranking, candidate normalization, policy resolution, and workbench code. Those changes repeat the same `portable_path(path).relative_to(portable_path(root))` operation. One helper would keep the portable-path containment rules consistent without removing any checks.

plugins/codex-security/scripts/deep_scan_workbench.py
```diff
@@ -757,10 +776,17 @@ def begin_deep_scan_for_target(
...
+            else state_dir() / "scans"
         )
-        target_root = (root / safe_segment(target.name)).resolve()
-        if target_root == target or target in target_root.parents:
+        target_root = filesystem_path(root / safe_segment(target.name)).resolve()
+        portable_target = portable_path(target)
+        portable_target_root = portable_path(target_root)
+        if (
+            portable_target_root == portable_target
+            or portable_target in portable_target_root.parents
```
mldangelo-oai, line 860: Non-blocking simplification: this is the same scan-root construction and containment check as [`workbench_db.scan_target_root`](https://github.com/openai/codex-security/blob/5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2/sdk/typescript/_bundled_plugin/scripts/workbench_db.py#L798-L813). Expose that helper through the existing Deep Scan dependencies and call it here, then create the directory. That keeps Standard and Deep scans on one path-normalization implementation.

mldangelo-oai (commented):
Reviewed [5c9c09fd24b4](https://github.com/openai/codex-security/commit/5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2) for correctness and simplification. Three optional cleanups for shared path handling and focused tests. Existing review threads were checked to avoid repeating the earlier long-path findings.

sdk/typescript/tests-ts/runtime.test.ts
```diff
@@ -395,15 +429,1052 @@ describe("plugin runtime preparation", () => {
...
+        inventoryOptions,
+      );
       expect(inventory.status, inventory.stderr).toBe(0);
       const posixRows = (await readFile(posixOutput, "utf8"))
         .trimEnd()
         .split(/\r?\n/u);
       expect(posixRows).toContain(String.raw`./literal\backslash.txt`);
       expect(posixRows).toContain("./literal:colon.txt");
+    } else {
+      expect(rows).toContain(`./${mixedLongPath}`);
```
mldangelo-oai, line 542: Non-blocking test cleanup: this Windows branch now runs inventory, ranking, policy, export, database-history, and Deep Scan recovery scenarios inside one test. Split them into focused tests, with a small shared long-path fixture and Python runner. An early assertion currently prevents the later scenarios from running, and none can be rerun or sharded independently.

mldangelo-oai (approved):
Rechecked 5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2. The earlier simplification suggestions remain non-blocking. Approving the reviewed head.

kmbroai (commented):
## Critical review

Reviewed draft head `5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2`. **Long-path support is justified, but there is a remaining archive correctness bug. Keep this draft until that boundary is fixed.**

### Correctness

**[P2] Compare archive siblings using the same path spelling.** In [archive_scan](https://github.com/openai/codex-security/blob/5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2/sdk/typescript/_bundled_plugin/scripts/workbench_scan_start.py#L97), the new canonicalization can return an ordinary path for the scan directory but an extended `\\?\` path for its longer `.previous-...` sibling. The unchanged comparison at line 105 then compares `archived_scan_dir.parent != scan_dir.parent` without normalizing either side. The same physical parent has different lexical spellings, and a valid archive is rejected as not being a previous sibling.

I reproduced the guard failure using the **actual `windows_paths.py` functions and actual `archive_scan` function with Windows lexical paths**, with filesystem canonicalization supplied at the boundary: a 229-unit scan path and its 255-unit archive path have equal portable parents but unequal filesystem-spelled parents. The ordinary-spelling control reaches the database; the PR's threshold-dependent spelling raises `The archived scan must be a previous sibling of the scan directory.` before any database call. This is a deterministic path-logic reproduction, **not native Windows filesystem execution**.

The failure is consequential because [prepareOutputDir already renames the old output](https://github.com/openai/codex-security/blob/5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2/sdk/typescript/src/runtime.ts#L1500) before [registration passes the archive path](https://github.com/openai/codex-security/blob/5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2/sdk/typescript/src/api.ts#L833) into the workbench. Rejection aborts startup before history/artifact paths are repointed. Compare `portable_path(archived_scan_dir).parent` with `portable_scan_dir.parent`, retaining the existing canonical/non-symlink checks. Add a native regression where the **archive suffix alone** crosses the helper threshold; testing both paths already above the threshold will miss this.

Do not solve this by requiring users to enable a machine-wide long-path policy. Supporting ordinary Windows paths across inventory, ranking, finalization, and history is a coherent feature. The portable-for-storage/filesystem-for-I/O distinction is the right foundation, but mixing those representations in equality and containment checks is the main maintenance risk.

I agree with the existing non-blocking simplification feedback: move `portable_relative_path` into `windows_paths.py` and reuse it across consumers; reuse the existing scan-root construction helper rather than maintaining a second copy in Deep Scan. Centralize these specific invariants, not a new general-purpose path abstraction.

The approximately 1,100-line expansion inside one runtime test should be split into focused inventory, archive/history, export, and Deep Scan tests sharing a small fixture. An early failure currently prevents later scenarios from running, and individual cases cannot be selected or sharded independently. The missing mixed-length archive case illustrates why separate boundary tests would help.

### Verification

Ran the focused Deep Scan workbench, reducer-recovery, history, and diff-ranking suites with seed 12345: **34 passed, 0 failed** on Bun 1.3.14/Linux with locally available dependencies, outside the sandbox and with a normal `022` umask. Also ran the path-logic control above. I did not run native Windows long-path I/O, the full runtime suite, or the packaging/platform matrix.

mldangelo-oai (changes requested):
Re-reviewed 5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2. The [archive-sibling bug already reported here](https://github.com/openai/codex-security/pull/435#pullrequestreview-5011515376) is still present, so this should stay unmerged. I independently reproduced it with the actual archive_scan and windows_paths functions: a 229-unit scan path and its 255-unit archive sibling pass on the base, but this head rejects them because only the archive gets the extended prefix. Normalizing both parents to portable spelling fixes that control; short/short and long/long controls still pass. This is a Windows path-logic reproduction on macOS, not native Windows filesystem execution.

Please fix that comparison and add a focused case where the archive suffix alone crosses the threshold. The existing suggestions to share portable_relative_path and scan_target_root, and to split the large runtime test into independently runnable cases, remain useful.

This branch also predates the canonical plugin-source move and now conflicts with main. Merge current main, carry the changes into plugins/codex-security instead of restoring tracked generated payloads, and rerun the portable source checks and Windows coverage on the resulting head before marking it ready.

sdk/typescript/tests-ts/runtime.test.ts
```diff
@@ -342,6 +353,13 @@ describe("plugin runtime preparation", () => {
       join(repository, "nested", "tracked-secret.py"),
       "secret = True\n",
     );
+    const mixedLongPath =
+      process.platform === "win32" ? `long-file-${"c".repeat(180)}.ts` : null;
+    if (mixedLongPath !== null) {
+      const mixedLongFile = join(repository, mixedLongPath);
+      expect(mixedLongFile.length).toBeGreaterThan(260);
```
mldangelo-oai: [P2] Size the long-path fixture from the actual temporary directory

The fixed 180-character filename does not guarantee a path longer than 260 characters. When `tmpdir()` returns `D:\tmp`, the current temporary-directory prefix, six-character suffix, and `repository` component produce a 248-character `mixedLongFile`, so this assertion fails before the test exercises long-path handling. Derive the filename padding from the actual repository path, ideally in the shared fixture used by the split Windows tests, so the result crosses the intended threshold independently of the temp directory.

mldangelo-oai (changes requested):
Completed a fresh PR/deslop review of `5c9c09fd24b48a6bdab4f70b02bcddf6ae8139e2`. There is one additional test failure, noted inline. The [previously reported archive-path bug](https://github.com/openai/codex-security/pull/435#pullrequestreview-5059951597) also still reproduces, and the branch still conflicts with current main.

The useful simplifications remain the existing requests to [share portable_relative_path](https://github.com/openai/codex-security/pull/435#discussion_r3799038463), [reuse scan_target_root](https://github.com/openai/codex-security/pull/435#discussion_r3799038470), and [split the large Windows test into independently runnable cases](https://github.com/openai/codex-security/pull/435#discussion_r3799038472). Keep the real Windows and legacy-path coverage while doing that; those compatibility cases are justified.

Validation: inspected all 20 changed files, reran the archive comparison with short/long path controls, and checked the new fixture failure with Node's Windows path implementation. These are path-logic checks on macOS, not native Windows filesystem test results.

mldangelo-oai:
@codex review

Please review the updated head `9c699ae19f9c6172d2358e6ec53c9212713523ab`, including the merge of current `main`, the canonical plugin source layout, and the archive path comparison.

mldangelo-oai:
@codex review

Please review the updated head `c22f1f6b85527d159139b1dfaabfdb80915e3488`. The follow-up keeps the Windows scope test's path double consistent in the shared path helper; it does not change production behavior.

sdk/typescript/tests-ts/runtime.test.ts
mldangelo-oai: Fixed in `9c699ae` and included in the current head `c22f1f6`: the filename padding is derived from the actual repository path, so the fixture crosses the intended Windows path threshold even under a short temporary root. The full inventory regression passed in [native Windows CI](https://github.com/openai/codex-security/actions/runs/33328249582/job/99310343700).

mldangelo-oai:
Conflict resolution is verified at `c22f1f6b85527d159139b1dfaabfdb80915e3488`. The branch includes `main` at `f714a6582ec0b7fe32fc06894a3eb9cdd768b48f` and also merges cleanly with the newer `774da259` base.

- All current checks are green, including [cross-platform node-ci](https://github.com/openai/codex-security/actions/runs/33328249582), installed-package checks, and both container workflows. An unchanged process-launch test passed on retry.
- Native Windows CI passed the full long-path inventory regression, legacy scan-root aliases, archive-name threshold case, and corrected scope parser probe. The long-filename fixture review thread is resolved with that evidence.
- The merged runtime at `9c699ae` passed both full local SDK runs: 2,111 passed, 44 skipped, zero failures each. The full Python suite passed 1,087 tests and 104 subtests, with 5 skips.
- The final test-double adjustment passed its two focused tests and formatting check. Additional local full runs were affected by system sleep and loopback socket failures; all eight affected CLI server tests subsequently passed on both this branch and unchanged SDK source.
- SDK/MCP typechecks, Ruff, and portable plugin source checks passed. Codex code and security reviews completed on the current commit; existing review feedback is not otherwise dismissed.

PR #446, opened 2026-08-15
feat(sdk): publish security findings through the connected Linear app

kmbroai:
## Summary

Publish prepared security findings through an existing authenticated Codex session and its connected Linear application, without introducing a separate Linear credential flow.

## Changes

- Expose the programmatic `publishScan(scanDirectory, options)` API and typed `onProgress` publication events.
- Accept an optional `AbortSignal`, terminate and await the entire Codex process tree on Unix and Windows, and preserve verified partial receipts before reporting cancellation.
- Invoke the existing Codex runtime with the ambient user configuration and stream exact prepared issue data through stdin.
- Stream Codex activity and independently verified Linear issue-creation results as they arrive, including fragmented JSONL output.
- Verify completed connected-app issue-creation events against the selected team, optional project, title, description, and priority.
- Support team-only publication without project lookup or project mutation arguments while retaining exact project/team validation when a project is supplied.
- Recognize actual dotted `linear.save_issue` connected-app events, one explicitly verified legacy alias, and Linear responses that expose issue identifiers as `id`.
- Support dry runs, empty scans, partial failures, more than 25 findings, repeated publications without deduplication, and failing optional progress observers.
- Write publication receipts outside sealed scan artifacts using hashed scan-ID filenames.
- If optional receipt storage fails after independently verified issue creation, return accurate created and failed results with a sanitized structured warning instead of encouraging duplicate publication.
- Add the new compiled modules to the public package allowlist.

## Testing

- Focused connected-app event verification and publishing SDK tests: 46 passed with 142 assertions.
- `tsc --noEmit`: passed.
- Generated-model validation, formatting, and an independent SDK build: passed.
- One real synthetic child process verifies fragmented, malformed, CRLF, and unterminated Codex JSONL, two dotted-tool successes, one failure, ordered progress, and the persisted receipt.
- Real subprocess regressions verify interrupted parents, signal-resistant grandchildren, forced termination, partial receipts, and Windows process-tree cleanup under Node 22 and Node 24.
- Receipt-failure regressions cover complete and partial verified publication, credential redaction, fabricated issue claims, zero created issues, and cancellation.
- Preserved the complete eight-case destination, payload, priority, and update-prevention security matrix without duplicating it for the legacy alias.

## Risk and rollout

- Uses the existing authenticated connected application and preserves its configured approval policies.
- Does not add API tokens, fallback destinations, issue readback, duplicate matching, or hardcoded application identifiers.
- Progress observers cannot interrupt publication, and issue progress is emitted only for verified destination-specific mutation results.
- Stacked on the sealed-scan publication foundation.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/publication-events.ts
+++ b/sdk/typescript/src/publication-events.ts
@@ -39,7 +39,8 @@ export function collectPublicationEvents(
       !isRecord(item) ||
       item["type"] !== "mcp_tool_call" ||
       item["server"] !== "codex_apps" ||
-      item["tool"] !== "linear_save_issue"
+      (item["tool"] !== "linear.save_issue" &&
+        item["tool"] !== "linear_save_issue")
     ) {
       continue;
     }
@@ -171,7 +172,8 @@ function savedIssue(
       isRecord(data) ? data["issue"] : undefined,
     ]) {
       if (!isRecord(value)) continue;
-      const identifier = value["identifier"] ?? value["issueIdentifier"];
+      const identifier =
+        value["identifier"] ?? value["issueIdentifier"] ?? value["id"];
       if (typeof identifier !== "string" || identifier.trim().length === 0) {
         continue;
       }
--- a/sdk/typescript/src/publication-events.ts
+++ b/sdk/typescript/src/publication-events.ts
@@ -47,21 +47,10 @@ export function collectPublicationEvents(
 
     const args = item["arguments"];
     const issue = isRecord(args)
-      ? publication.issues.find(
-          (candidate) =>
-            candidate.title === args["title"] &&
-            candidate.description === args["description"],
-        )
+      ? matchPublicationIssue(publication, args)
       : undefined;
     if (issue === undefined) {
       unexpected.push("Codex attempted to create an unexpected Linear issue.");
-      continue;
-    }
-    if (!isRecord(args) || !hasExpectedArguments(args, publication, issue)) {
-      failed.set(
-        issue.findingId,
-        "Codex attempted to create a Linear issue with unexpected arguments or destination.",
-      );
       continue;
     }
     if (failed.has(issue.findingId) || created.has(issue.findingId)) {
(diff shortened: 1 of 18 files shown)


wholley-oai (dismissed):

sdk/typescript/src/publication-events.ts
```diff
@@ -0,0 +1,189 @@
...
+export function matchPublicationIssue(
+  publication: PreparedScanPublication,
+  arguments_: Record<string, unknown>,
+): PreparedPublicationIssue | undefined {
+  const description = arguments_["description"];
+  if (typeof description !== "string") {
+    return undefined;
+  }
+
+  const matches = publication.issues.filter((issue) =>
```
soyeon-oai, line 122: [P1] Verify the complete mutation arguments. This identifies a finding only from IDs embedded in `description`, so a completed `linear.save_issue` event with a different team, project, title, description, or priority (or with extra fields) is accepted and counted as the requested publication. That can send finding details outside the approved destination while reporting success. Require an exact permitted argument object with no extras, backed by a runtime-authenticated tool receipt.

sdk/typescript/src/publication-events.ts
```diff
@@ -0,0 +1,189 @@
...
+        issue.findingId,
+        "Codex attempted to create more than one Linear issue for this finding.",
+      );
+      continue;
+    }
+    if (item["status"] !== "completed") {
+      const error = item["error"];
+      failed.set(
+        issue.findingId,
+        isRecord(error) && typeof error["message"] === "string"
```
soyeon-oai, line 67: [P2] Sanitize connector failures at ingestion. A connected-app error message is copied verbatim into the SDK result and then the durable receipt, so diagnostics containing credentials or private payload text cross the existing output-redaction boundary. Apply the shared safe-error sanitizer here (and cap stored diagnostic size) before returning or persisting it.

sdk/typescript/src/publish.ts
```diff
@@ -0,0 +1,397 @@
...
+  const events = collectPublicationEvents(
+    invocation.stdout,
+    prepared,
+    failureMessage,
+  );
+  result.created = events.created;
+  result.failed = events.failed;
+  result.counts.created = events.created.length;
+  result.counts.failed = events.failed.length;
+  await (dependencies.writeReceipt ?? writePublicationReceipt)(
```
soyeon-oai: [P2] Do not turn a completed remote mutation into an apparent failed publication when receipt writing fails. If this `await` throws after Linear issues were created, the caller receives an exception and the natural retry creates another full set. Treat the optional receipt failure as a warning with the successful result, or retain durable recovery evidence that prevents a blind retry.

sdk/typescript/src/publish.ts
```diff
@@ -0,0 +1,397 @@
...
+
+async function runPublicationCodex(
+  command: CodexCommand,
+  args: readonly string[],
+  input: string,
+  environment: NodeJS.ProcessEnv,
+  onEvent?: (event: unknown) => void,
+): Promise<PublicationCodexResult> {
+  return new Promise((resolve, reject) => {
+    const child = spawn(command.command, [...args], {
```
soyeon-oai, line 351: [P1] Propagate cancellation to the spawned Codex process. There is no `AbortSignal` or child/process-tree cleanup here, so terminating the parent can leave the child running and creating Linear issues after the CLI reports cancellation. Thread an abort signal through publication, terminate and await the child on SIGINT/SIGTERM (including Windows behavior), and preserve partial recovery state before exit.

soyeon-oai (commented):
Reviewed the exact PR delta. I found four publication correctness and safety issues.

sdk/typescript/src/publication-events.ts
kmbroai: this introduced a bunch of instability when I did this in a previous commit. so i'll leave it as it is

sdk/typescript/src/publish.ts
kmbroai: good call

sdk/typescript/src/publish.ts
kmbroai: I like it. might help with the weird flakiness. moves it to a data issue rather than a runtime error.

daneschneider-oai (approved):

PR #453, opened 2026-08-15
feat(sdk): publish findings through durable concurrent batches

kmbroai:
## Summary

Publish prepared security findings in bounded concurrent batches and derive final results from validated durable records stored in local scan history.

## Changes

- Group connected Linear issue creation into batches of at most 20 findings.
- Write finding-to-issue mappings into a private durable handoff instead of trusting model-generated summaries.
- Validate exact finding, occurrence, destination, team, optional project, title, description, and priority before importing publication records.
- Preserve team-only publication through private handoffs, recovered issue mappings, SQLite records, receipts, and results without inventing a project identifier.
- Read final issue identities, URLs, and receipts back from the transactional workbench store.
- Keep database-backed issue creation successful if its optional receipt fails, returning a sanitized warning without rerunning remote mutations.
- Preserve partial successes, recover incomplete handoffs, and retain recoverable mappings if database persistence fails.
- After cancellation, await Codex process-tree termination, persist verified completed issues, write an accurate partial receipt, and retain the private handoff with duplicate-avoidance recovery instructions.
- Keep CLI commands, interactive presentation, and end-to-end CLI integration out of this SDK layer.

## Testing

- Real SQLite publication-store, connected-app event validation, and durable publisher tests: 70 passed with 286 assertions.
- Verified a 41-finding 20 + 20 + 1 batching boundary, malformed or absent Codex output, out-of-order completion, exact destination validation, partial failures, and durable recovery.
- Consolidated recovery assertions into the stronger case preserving both written and independently salvaged issue mappings when database persistence fails.
- Verified pre-aborted requests, cancellation during preparation, interrupted partial imports, database failures, receipt failures, retained private handoffs, and real signal-resistant child processes.

## Risk and rollout

- Publication still uses the existing authenticated Codex profile and connected Linear app.
- No issues are created before validating the completed local scan and every finding association.
- Stacked on the workbench publication-persistence layer.

## Public disclosure review


wholley-oai (dismissed):

sdk/typescript/src/publish.ts
```diff
@@ -290,13 +365,285 @@ function publicationPrompt(publication: PreparedScanPublication): string {
...
+      created.delete(issue.findingId);
+      failed.set(
+        issue.findingId,
+        "Codex wrote more than one Linear publication for this finding.",
+      );
+      continue;
+    }
+    observed.add(issue.findingId);
+
+    if (
```
soyeon-oai, line 566: [P1] Validate the handoff's complete `arguments` object. The record is matched only by scan/finding/occurrence IDs and a nonempty issue identifier; `record.arguments` is never compared with the prepared team, project, title, description, and priority. A model-authored handoff can therefore claim success for a mutation made with different arguments, which is then persisted under the requested destination. Require exact arguments with no extras and bind them to an authoritative runtime tool receipt.

sdk/typescript/src/publish.ts
```diff
@@ -290,13 +365,285 @@ function publicationPrompt(publication: PreparedScanPublication): string {
...
+  const eventFailed = new Map(
+    events.failed.map((issue) => [issue.findingId, issue.error]),
+  );
+  for (const issue of publication.issues) {
+    const saved = created.get(issue.findingId);
+    const verified = eventCreated.get(issue.findingId);
+    const eventFailure = eventFailed.get(issue.findingId);
+    if (
+      saved === undefined &&
+      !observed.has(issue.findingId) &&
```
soyeon-oai: [P1] Do not let a model-authored failure override a verified successful tool event. In a live synthetic E2E, the runtime observed and displayed a created Linear issue, but the handoff writer then failed and emitted a failure record; `observed` blocked this salvage path, so the CLI returned `created: 0`, persisted no association, deleted the handoff, and left the real issue remote. A retry would duplicate it. Prefer the exact-argument completed tool receipt, or retain recovery state and report an indeterminate outcome.

sdk/typescript/src/publish.ts
```diff
@@ -189,10 +207,51 @@ export async function publishScanInternal(
...
+      throw new CodexSecurityError(
+        `Could not persist created Linear issues: ${detail}. The publication handoff remains at ${handoff.file}; recover it before retrying to avoid creating duplicate issues.`,
+        { cause: error },
+      );
+    }
+  }
+  result.failed = handoffResults.failed;
+  result.counts.created = result.created.length;
+  result.counts.failed = result.failed.length;
+  await rm(handoff.directory, { recursive: true, force: true }).catch(
```
soyeon-oai, line 282: [P2] Keep recovery state until the receipt is durable. The database commit can succeed, this removes the handoff, and then `writeReceipt` can fail. The command reports an error with neither a handoff nor a receipt even though remote creation and local persistence completed, encouraging a duplicate retry. Write the receipt atomically before cleanup, or return the successful result with a warning after the DB commit.

soyeon-oai (commented):
Reviewed the exact PR delta and exercised the stacked CLI against a real synthetic Linear publication. I found three durability/verification issues.

sdk/typescript/src/publish.ts
soyeon-oai: This was reproduced in my e2e.

soyeon-oai: image

failed in cli, but ticket was created.

sdk/typescript/src/publish.ts
```diff
@@ -310,13 +404,285 @@ function publicationPrompt(publication: PreparedScanPublication): string {
...
+      continue;
+    }
+    const issue = expectedIssues.get(record["findingId"]);
+    if (issue === undefined) {
+      unexpected.push(
+        "Codex wrote a Linear publication for an unknown finding.",
+      );
+      continue;
+    }
+    if (observed.has(issue.findingId)) {
```
soyeon-oai, line 533: [P1] Preserve both successful issue IDs when duplicate handoffs are detected.

Found during E2E review:
```text
completed issues: SYNTH-DUPLICATE-A, SYNTH-DUPLICATE-B
CLI: created=0, failed=1 | SQLite associations=0 | recovery handoffs=0
```
Detecting duplicates is correct, but discarding both IDs and deleting the handoff makes already-created issues unrecoverable. Persist the verified IDs or retain the handoff and report an indeterminate outcome.

sdk/typescript/src/publish.ts
```diff
@@ -200,10 +219,69 @@ export async function publishScanInternal(
...
+    failureMessage,
+  );
+  if (handoffResults.created.length > 0) {
+    await preserveVerifiedHandoff(
+      handoff.file,
+      prepared,
+      handoffResults.created,
+    );
+    try {
+      result.created = await (
```
soyeon-oai, line 250: [P2] Prevent concurrent publications from overwriting successful receipts.

Found during E2E review:
```text
concurrent publication A: success
concurrent publication B: success
SQLite associations=2 | issues in the shared scan receipt=1
```
Fresh issues on republish are documented; silently losing one successful run from the scan-keyed receipt is the defect. Serialize same-scan publication or write attempt-scoped receipts.

sdk/typescript/src/publish.ts
```diff
@@ -144,13 +157,19 @@ export async function publishScanInternal(
   }
   if (prepared.issues.length === 0) return result;
 
+  const environment = dependencies.environment ?? process.env;
+  await (dependencies.preparePublicationStore ?? preparePublicationStore)(
+    prepared,
+    environment,
+  );
+  options.signal?.throwIfAborted();
+  const handoff = await createPublicationHandoff(prepared, environment);
```
soyeon-oai, line 170: [P2] Remove source-bearing handoffs when the publisher cannot start.

Found during E2E review with a missing `CODEX_CLI_PATH`:
```text
CLI exit=2 | spawned processes=0 | remote mutations=0
retained publication.json=1447 bytes | finding description=1011 bytes
```
The startup rejection bypasses cleanup, leaving finding details and source locations on disk despite no recoverable remote operation. Delete pre-mutation handoffs on startup failure.

soyeon-oai (commented):
Additional issues reproduced during E2E review on the latest stack (#453 86384dbf; integrated CLI #455 1fd8b0de; identical publisher source).

daneschneider-oai (approved):

PR #447, opened 2026-08-15
feat(cli): publish completed security scans to Linear

kmbroai:
## Summary

Add `codex-security publish scan [SCAN_DIR]` so completed security scans can be published to an explicitly selected connected Linear team, with an optional project.

## Changes

- Add the nested publication command with `--to linear`, `--linear-team`, optional `--project`, `--dry-run`, and structured JSON output.
- Select an existing completed scan interactively when no scan directory is provided; exclude missing, replaced, inaccessible, and symlinked scan directories before showing the picker.
- Resolve destination defaults from `CODEX_SECURITY_LINEAR_TEAM` and `CODEX_SECURITY_LINEAR_PROJECT` while letting explicit flags take precedence.
- Preserve database-backed created-issue results, partial failures, accurate exit codes, and scan-ID provenance.
- Display sanitized optional-receipt warnings on stderr without hiding created issues, contaminating JSON stdout, or changing publication exit codes.
- Handle SIGINT and SIGTERM without exiting early: await child shutdown and durable recovery, preserve recovery guidance, and return conventional exit codes 130 and 143.
- Document source-code evidence, destination requirements, 20-item concurrent batches, and existing local scan-history requirements.
- Add an installed-package CLI publication dry-run smoke.
- Preserve sealed-file identity on Windows Node 22 using the independently verified fail-closed, two-handle identity check from #454.
- Exercise actual sealed scans, bundled Python, SQLite associations, durable handoff, CLI JSON, receipts, and partial failures end to end.
- Keep dashboard styling, relative scan ages, aligned columns, terminal rendering, and live interactive presentation in the follow-up UX PR.

## Testing

- Focused CLI routing, real-filesystem picker filtering, receipt warnings, interruption handling, Windows sealed-file validation, and real SQLite-backed integration: 90 passed with 563 assertions.
- The CLI command layer retains 15 focused routing, destination, dry-run, real-directory selection, receipt-warning, interruption, and exit-code tests; cross-batch behavior is verified by real SQLite integrations.
- Real end-to-end sealed scans verified 23 successful project-scoped findings across 20 + 3 concurrent batches and 21 persisted team-only successes with one failure across 20 + 2 batches.
- A real sealed-scan interruption persists one durable mapping and one independently verified mutation in SQLite, writes the partial receipt, retains its recovery handoff, and exits 130 with clean JSON stdout.
[...]

PR #454, opened 2026-08-15
fix(sdk): preserve sealed-file identity on Windows Node 22

kmbroai:
## Summary

Accept the same checked regular file on Windows Node 22 when path and open-handle metadata represent the volume identifier differently.

## Changes

- Keep exact device identity checks on non-Windows platforms and whenever ordinary metadata already agrees.
- On Windows device mismatches only, compare the original file with a second independently opened no-follow handle using exact full-precision descriptor device and inode identities.
- Continue requiring a regular non-symlink file, unchanged original numeric inode, identical full-precision path inode, unchanged checked parents, and unchanged final path identity.
- Add regressions for volume identifiers larger than JavaScript's safe integer range, differently represented numeric inodes, changed exact inodes, non-regular files, and strict non-Windows behavior.
- Load the installed SDK's bundled sealed-scan example with the actual Node runtime during package verification so Windows Node 22 genuinely exercises this path.

## Testing

- Complete sealed-scan contract suite: 30 passed with 97 assertions.
- Built, packed, installed, and validated the public SDK, including its real Node-runtime sealed-scan contract smoke.
- Captured actual Windows Node 22 diagnostics proving path and file-handle device identifiers differ even in their low 32 bits, then verified exact same-API descriptor identity instead.

## Risk and rollout

- File access remains fail-closed; no symlink, ownership, parent, inode, or scan-seal checks are removed.
- Kept separate from the security-finding publication feature stack because it also protects existing scan consumers.

## Public disclosure review

### DIFF
--- a/sdk/typescript/scripts/smoke-package.mjs
+++ b/sdk/typescript/scripts/smoke-package.mjs
@@ -1,8 +1,12 @@
 import assert from "node:assert/strict";
 import { spawnSync } from "node:child_process";
 import {
+  chmod,
+  cp,
+  lstat,
   mkdir,
   mkdtemp,
+  open,
   readFile,
   readdir,
   rm,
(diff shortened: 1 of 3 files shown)


kmbroai:
Superseded by current `main`: the Windows Node 22 sealed-file identity fix and its exact regression coverage landed through #447. Current package verification also installs the SDK and loads the bundled sealed scan through the installed CLI on the Windows Node 22 lane. There is no distinct change left in this PR, so I’m closing it rather than publishing duplicate coverage.

PR #463, opened 2026-08-15
fix(cli): return safe structured scan failure details

mldangelo-oai:
## Summary

Return complete, privacy-safe JSON and JSONL error envelopes when scan execution or saved-scan replay fails, while preserving detailed stderr diagnostics and existing exit codes.

## Changes

- Emit `SCAN_FAILED` or `SCAN_REPLAY_UNAVAILABLE` with fixed messages. Keep both ordinary and `--full-output` shapes complete even when token-output options are present.
- Preserve detailed local errors, cancellation, partial-output guidance, and the distinction between local plugin setup and installer failures.
- Keep `LocalPluginBootstrapError` compatible with `PluginBootstrapError` catches and retain underlying error causes.
- Inline three single-use private implementations into their existing public error boundaries and consolidate duplicate typed-error test loops. Remove 26 net lines from the merged baseline, including documentation and cache checks.
- Document the structured failure and SDK error contracts. Merge current main and preserve the draft state.

## Testing

Current main integration and CI repair:

- Authentication, CLI, runtime, and knowledge-base tests: 333 passed, 21 platform skips. The two failing authentication cases now assert the structured error object while retaining the checks on detailed stderr output.
- Generated-model checks, MCP typechecking, TypeScript checking, formatting, bundled-plugin source checks, and `git diff --check`: passed. TypeScript checking was also run without its incremental cache after a temporary-storage error.
- The full local seed-12345 run encountered exhausted temporary storage and is not counted as a pass. Full-suite and installed-package verification are pending on the new-head CI run.
- No live authentication or model calls were performed. The PR remains a draft.

## Risk and rollout

JSON/JSONL callers receive a structured error instead of empty stdout for these failures. Successful output and plain-text diagnostics retain their existing behavior. Detailed stderr may contain local diagnostic context. This does not change scan execution, policy decisions, authentication, permissions, or partial-result sealing, and adds no CLI flags.

The bundled plugin is byte-for-byte identical to current main and keeps main's version; no branch-specific plugin or npm version is introduced. Preserve draft status until review and current-head CI are complete.

## Public disclosure review

Historical commit contact metadata and automated restricted report/settings links remain in the existing PR history, so the second attestation stays unchecked. New fixtures are synthetic; the new commit uses GitHub noreply metadata.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

Please review the current head, `301e54159886185274b956d18ac84756477100cc`.

sdk/typescript/src/runtime.ts
```diff
@@ -1911,6 +1917,19 @@ export async function resolvePluginPath(
   pluginPath: string | undefined,
   workspace: string,
   signal?: AbortSignal,
+): Promise<string> {
+  try {
+    return await resolveLocalPluginPath(pluginPath, workspace, signal);
+  } catch (error) {
+    if (signal?.aborted || error instanceof ConfigurationError) throw error;
+    throw new ConfigurationError(errorMessage(error), { cause: error });
```
zcrab-oai: **[P2] Preserve the public plugin-bootstrap error contract**

Could we keep `PluginBootstrapError` for the publicly exported plugin helpers and classify local failures only at the CLI boundary? `resolvePluginPath`, `createMarketplace`, `pluginMetadata`, and `bootstrapPlugin` now replace established bootstrap failures with the sibling `ConfigurationError`, and existing tests had to change their expected error class. SDK integrations that catch the exported bootstrap error will stop handling malformed plugins and unsafe marketplace paths.

zcrab-oai (commented):
Draft feedback: the CLI classification change should not break the public SDK error contract.

mldangelo-oai:
@codex review

Please review commit `2a37ea13b64dcab5648ec50cdbbfef01b4efe0a8`.

kmbroai (commented):
## Critical review

Reviewed head `2a37ea13b64dcab5648ec50cdbbfef01b4efe0a8`.

**Recommendation: keep the structured-failure fix. No new blocking correctness defect found in the reviewed changes.** A nonzero exit without a usable JSON failure object is a real integration problem; successful-result formatting and failure formatting cannot be treated interchangeably.

### Correctness and compatibility

The [scanFailure projection](https://github.com/openai/codex-security/blob/2a37ea13b64dcab5648ec50cdbbfef01b4efe0a8/sdk/typescript/src/cli.ts#L1110) explicitly owns the failure envelope and avoids Incur invocation metadata. Keeping errors complete even with token controls is justified: truncating or offsetting a machine-readable failure would defeat the advertised contract. Scan and rerun failures preserve their distinct codes, and unavailable replay now produces a failure rather than empty apparent success.

The current head also resolves the earlier public error-class concern: `LocalPluginBootstrapError` extends `PluginBootstrapError`, so SDK consumers catching the established superclass continue to work. Knowledge-base staging failures are classified as local configuration failures rather than misdiagnosed as model transport errors. The machine-readable message is a fixed category while stderr retains the existing local troubleshooting behavior; this is not a promise that all diagnostics are safe to publish.

### Simplification and necessity

Keep the small closed projection instead of serializing arbitrary Error objects or inventing another error schema. Likewise, keep the bootstrap subclass specific to a meaningful local-versus-remote distinction; there is no need for a general taxonomy framework. There are now separate rendering overrides for history, publication, and scans. That duplication is worth watching, but unifying them should wait for genuinely identical envelope contracts—not become a prerequisite for this fix.

The tests should continue to emphasize JSON parseability, exit status, credential/metadata exclusion, and superclass compatibility rather than exact diagnostic prose. Those are the externally observable guarantees.

### Verification

Ran `cli.test.ts`, `knowledge-base.test.ts`, and `runtime.test.ts`: **277 passed, 7 skipped, 0 failed**, including structured failures, full-output/token controls, replay unavailability, and bootstrap behavior. Linux, Bun 1.3.14 / Node 22.13.1, cached dependencies. No native Windows run, live authentication, or model request was performed.

sdk/typescript/src/runtime.ts
mldangelo-oai: Verified on the current head: LocalPluginBootstrapError extends PluginBootstrapError, so existing SDK catches retain the public error contract. Runtime coverage verifies both the specialized subclass and existing public PluginBootstrapError cases. Resolving this outdated discussion.

mldangelo-oai:
@codex review

Please review current head `aa98e1b82b8785c304c085c35c9492d129b22542` after the main integration and authentication-test repair. Check that JSON failures remain structured and exclude upstream details while stderr retains the existing recovery guidance. The PR remains a draft.

PR #469, opened 2026-08-15
fix(cli): preserve ambient Codex filesystem denials

daneschneider-oai:
## Summary

Codex Security uses a generated Codex home. Preserve filesystem denials from the caller's effective Codex permission configuration when starting a scan, without changing the scanner's intended write access.

## Changes

- Pin the Codex CLI and TypeScript SDK to the matching `0.148.0-alpha.21` release.
- Pass each scan's filesystem profile through process-local raw SDK overrides instead of rewriting a shared configuration for each turn.
- Carry forward active and explicitly configured deny rules, including inherited profiles, scoped paths, globs, and supported glob-depth settings. Preserve literal paths without silently applying a rule to additional workspace roots.
- Inspect the selected scan profile before starting a turn. Refuse a scan when its restrictions cannot be preserved or organization policy selects an uninspectable profile.
- Include concrete administrator setup steps in `sdk/typescript/README.md` and link to that guide from the CLI error message.
- Keep generated-home authentication, concurrent scans, follow-up turns, and saved explicit scan restrictions working with the same override path.
- Apply the same permission preparation and verification to standalone SDK finding validation, with regression coverage for preserved and missing denials.

## Testing

Local checks used Node `24.15.0` and Bun `1.3.14`.

- Focused API, runtime, concurrency, custom-validation, and API-surface tests (`--timeout 30000 --seed 12345`): 298 passed, 13 skipped, zero failures.
- `pnpm --ignore-workspace run test --seed 12345`: 1,727 passed, 30 skipped, zero failures.
- `pnpm --ignore-workspace run generate:models:check` and `pnpm --ignore-workspace exec tsc --noEmit`: passed.
- `pnpm --ignore-workspace run format` and `pnpm --ignore-workspace run build`: passed.
- Frozen-lockfile installation and `git diff --check`: passed.

## Risk and rollout

The administrator guide ships with the runtime change, so the CLI error's README link resolves when this PR merges. The Codex dependencies remain exact prerelease pins: the published stable `0.148.0` TypeScript SDK does not contain the required raw-override API. Configurations whose deny rules cannot be represented exactly fail before a scan starts. Permission inspection and execution are separate processes, so a concurrent policy change between them is still possible; this is not an atomic runtime enforcement guarantee.

## Public disclosure review


ianw-oai:
Since this adds a thousand lines of code to api.ts, can we factor out the change and try to simplify it?

sdk/typescript/src/api.ts
```diff
@@ -3322,6 +3415,1158 @@ export function classifyConnectionFailure(
...
+): boolean {
+  for (const [path, access] of Object.entries(required)) {
+    if (path === "glob_scan_max_depth") continue;
+    const actualAccess = actual[path];
+    if (access === "deny") {
+      if (actualAccess !== "deny") return false;
+      continue;
+    }
+    if (!isRecord(access)) continue;
+    if (actualAccess === "deny") {
```
mldangelo-oai, line 4066: **[P1] Validate effective access beneath required denies**

An effective profile containing `/data = "deny"` and `/data/public = "read"` passes this check: `filesystemDenials()` removes the child read, `filesystemWritesCoveredBy()` does not examine reads, and this branch accepts the broad deny. Pinned Codex resolves access using the most-specific matching path, so `/data/public` remains readable despite the required `/data` denial; recursive config merging can preserve both entries. Please compare the compiled effective Codex policy—ideally through a Codex-provided containment API—instead of duplicating path/glob precedence here, or at minimum reject overlapping read/write carve-outs beneath required denies.

sdk/typescript/src/api.ts
```diff
@@ -3322,6 +3415,1158 @@ export function classifyConnectionFailure(
...
+    visited.add(current);
+    const profile = profiles[current];
+    if (!isRecord(profile)) return null;
+    const parent = profile["extends"];
+    if (parent === undefined || parent === null) {
+      return { ancestors, builtin: null };
+    }
+    if (typeof parent !== "string") return null;
+    if (parent.startsWith(":")) {
+      if (parent === ":read-only" || parent === ":workspace-write") {
```
mldangelo-oai, line 4106: **[P2] Do not hard-code Codex built-ins in TypeScript**

`:workspace-write` is not a permission-profile ID in pinned Codex 0.148.0-alpha.21; the built-in is `:workspace` (`workspace-write` is the legacy sandbox-mode value). The test at `api.test.ts:5563` therefore fabricates a profile the real parser cannot produce, while a real `extends = ":workspace"` chain falls through to `null` here and is rejected without evaluating its effective rules. This fails closed, but it rejects safely narrowed profiles and keeps the regression green for the wrong reason. Please avoid fixing only the spelling: use Codex's typed compiled policy/validation and add a contract test with the real built-in ID.

sdk/typescript/src/runtime.ts
```diff
@@ -2438,6 +2464,211 @@ export async function runCodexCommand(
...
+  const child = spawn(
+    command.command,
+    [
+      ...configOverrides.flatMap((override) => ["--config", override]),
+      "app-server",
+      "--disable",
+      "plugins",
+    ],
+    {
+      env: environment,
```
mldangelo-oai, line 2483: **[P2] Spawn the inspection app-server in the inspected directory**

The target `cwd` is sent only in later RPC payloads; the process itself starts in the caller's directory. App-server loads startup authentication and managed configuration before those RPCs, so scanning a different repository can inspect target config while using startup policy from an unrelated checkout. This reproduces with the pinned runtime: `config/read` reports the target, but a caller-side forced login method still controls `account/login/start`. Add `cwd` to these spawn options and make the fake app-server test assert `process.cwd()` as well as the RPC fields.

mldangelo-oai (commented):
**Direction: restart on current main and move policy semantics into Codex**

The denial-preservation gap is real, but this shape should be restarted on current main rather than incrementally conflict-resolved. The branch is 111 commits behind, conflicts with current API/runtime files, and pins 0.148.0-alpha.21 while main uses 0.149.1. More importantly, `api.ts:3418-4568` and `runtime.ts:2467-2669` recreate Codex's permission inheritance, built-ins, path/glob expansion, Windows normalization, managed requirements, and effective-policy precedence in roughly 1,400 production lines. The inline findings are already concrete semantic drift from that duplication.

Please move the primitive into Codex/app-server: compile the scanner allow side and preserve ambient deny restrictions in Codex's typed effective policy, returning a typed result or validation outcome. Then this PR can delete roughly 1,300-1,450 production lines and most of the 1,213 added test lines, retaining a small integration matrix for exact path/glob/workspace denials, managed rejection, follow-up turns, and concurrency. This is not a request for another local path/glob checker; the simplification is to stop translating Codex filesystem policy in TypeScript.

PR #471, opened 2026-08-15
[codex] Show security scans in the Codex app

ianw-oai:
## Summary

Security scans use a private Codex home, so the Codex app cannot see their tasks. Show each scan in the app without moving its credentials or session files.

## Changes

- Start an app-visible Codex task and resume that task for the scan.
- Share the app task database while keeping scanner credentials and session files private.
- Continue the scan if app task creation fails.
- Add focused coverage and include the new module in package checks.

## Testing

- `bun test --timeout 30000 tests-ts/api.test.ts tests-ts/api-attribution-concurrency.test.ts tests-ts/desktop-session.test.ts` — 122 passed.
- `pnpm run types`
- `pnpm run format`
- `pnpm run build`
- `pnpm run check:package /private/tmp/openai-codex-security-0.1.12.tgz`
- Started a scan against a synthetic repository and confirmed its task appears in the app task list.

## Risk and rollout

Scan authentication, private session storage, and repository path checks do not change. If app task setup fails, scanning falls back to the existing behavior.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/desktop-session.ts
+++ b/sdk/typescript/src/desktop-session.ts
@@ -1,0 +1,81 @@
+import { spawn } from "node:child_process";
+import { once } from "node:events";
+import { createInterface } from "node:readline";
+import type { CodexOptions } from "@openai/codex-sdk";
+import type { CodexCommand } from "./runtime.js";
+import { CODEX_SDK_VERSION } from "./version.js";
+
+export interface DesktopSessionOptions {
+  command: CodexCommand;
+  environment: Record<string, string>;
+  config: NonNullable<CodexOptions["config"]>;
+  workingDirectory: string;
+  title: string;
+  signal: AbortSignal;
+}
+
+export async function prepareDesktopSession(
+  options: DesktopSessionOptions,
+): Promise<string> {
+  const child = spawn(options.command.command, ["app-server"], {
+    env: options.environment,
+    signal: options.signal,
+    stdio: ["pipe", "pipe", "pipe"],
+    windowsHide: true,
+  });
+  child.stderr.resume();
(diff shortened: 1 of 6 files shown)


sdk/typescript/src/desktop-session.ts
```diff
@@ -0,0 +1,81 @@
...
+    const result = await request(1, "thread/start", {
+      cwd: options.workingDirectory,
+      approvalPolicy: "never",
+      config: options.config,
+    });
+    const threadId = (result["thread"] as { id: string }).id;
+    await request(2, "thread/name/set", {
+      threadId,
+      name: options.title,
+    });
```
mldangelo-oai, line 71: Non-blocking: the task is visible and resumable, but naming it through the private-home app server did not make the name appear in the app-home database-only list with the pinned runtime. Naming the same task from the app-home side did. Could we use the supported shared-metadata path, or do this metadata-only operation from the app-home side, while leaving scan execution and credentials isolated? I'd check the name from the app's listing context in the regression test.

mldangelo-oai (commented):
Non-blocking QA notes on `eb941990`. The shared-database approach worked with the desktop's database-only task list. I could read and continue the saved scan task. I'd keep scan execution in its isolated home.

The name was the remaining issue I reproduced; details are inline. It would also be useful to share the small app-server client with #477 instead of maintaining two request loops.

The full local suite passed 1,139 tests with 11 skipped. The branch currently conflicts with main, and the title and Windows checks still fail. A branch refresh, Conventional Commit title, and Windows rerun would help. The observed Windows failure occurs during credential-home setup, so I wouldn't attribute it to this feature without a clean rerun.

PR #475, opened 2026-08-16
feat: patch Linear issues and projects

ianw-oai:
## Summary

Import Linear issues and patch them with the existing CLI workflow.

## Changes

- Patch selected issues or every open issue in a project, with native Linear filters and automatic pagination.
- Keep imported content literal, retain each issue's ID and URL, and check the workspace selected by an issue URL.
- Share Linear client setup with publishing and keep credentials out of the patch subprocess.
- Use `--linear-*` options consistently. Publishing still accepts `--project` as an alias.

```bash
codex-security patch --linear-issue SEC-123 --linear-issue SEC-124
codex-security patch --linear-project "Security backlog"
codex-security patch --linear-project "Security backlog" \
  --linear-filter '{"labels":{"name":{"eq":"security"}}}'
```

Set `CODEX_SECURITY_LINEAR_API_KEY` for Linear access.

## Testing

- Full randomized test suite: 1,296 passed, 11 skipped.
- TypeScript, formatting, build, CLI help, and installed npm package checks passed.

## Risk and rollout

Importing issues is read-only. Existing positional patch inputs and publication authentication behavior are unchanged. Linear credentials are not passed to Codex. No live Linear-to-PR run was performed.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -82,6 +82,7 @@ import {
   ScanInterruptedError,
 } from "./errors.js";
 import type { SeverityLevel } from "./models.js";
+import { importLinearIssues, type LinearClientFactory } from "./linear.js";
 import { runMultiscan } from "./multiscan.js";
 import {
   publishScan,
@@ -204,6 +205,9 @@ const VALUE_OPTIONS = new Set([
   "--plugin-path",
   "--python",
   "--codex",
+  "--linear",
+  "--linear-project",
+  "--filter",
   "--fail-on-severity",
   "--max-cost",
   "--workers",
@@ -715,6 +719,7 @@ interface CliDependencies {
     environment?: NodeJS.ProcessEnv,
   ): Promise<number>;
   bulkScan?: BulkScanDiscoveryDependencies;
+  linearClient?: LinearClientFactory;
   runWorkbench(args: readonly string[]): Promise<JsonObject>;
   matchFindings: typeof matchScanFindings;
   checkForUpdate(signal: AbortSignal): Promise<UpdateNotice | undefined>;
@@ -2450,10 +2455,21 @@ export async function main(
         "issues...": z
           .string()
           .min(1, "An issue must not be empty.")
+          .optional()
           .describe("Issue text or a file containing issues."),
       }),
       options: z.object({
         effort: effortOption(),
+        linear: z
+          .array(optionValue("--linear"))
+          .default([])
+          .describe("Linear issue identifier or URL; repeat for more issues."),
+        linearProject: optionValue("--linear-project")
+          .optional()
+          .describe("Patch every open issue in this Linear project."),
+        filter: optionValue("--filter")
+          .optional()
+          .describe("JSON Linear issue filter for --linear-project."),
         codex: z
           .array(optionValue("--codex"))
           .default([])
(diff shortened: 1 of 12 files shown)


sdk/typescript/src/cli.ts
```diff
@@ -2450,10 +2455,21 @@ export async function main(
         "issues...": z
           .string()
           .min(1, "An issue must not be empty.")
+          .optional()
           .describe("Issue text or a file containing issues."),
       }),
       options: z.object({
         effort: effortOption(),
+        linear: z
+          .array(optionValue("--linear"))
```
kmbroai: noop comment, calling out so we are aware

earlier we did `--to linear` to make it more generic for publishing. but then I let codex add `--linear-...` to the prefix of things bc they were very linear specific. IMO we should either go full generic with `--to` and stuff like `--ticket-api-key` or more specific with `--linear-...` prefixed flags.

TL;DR I don't like how I did a mix of the 2.

sdk/typescript/src/linear.ts
```diff
@@ -0,0 +1,125 @@
...
+          );
+        }
+        issues.push(issue);
+      }
+    }
+
+    return issues.map(
+      ({ identifier, title, url, description }) =>
+        `Linear issue: ${identifier}\nTitle: ${title}\nURL: ${url}\n\n${description ?? ""}`,
+    );
```
mldangelo-oai: Non-blocking: when this feeds the common patch result, could we retain the issue ID and URL alongside the text? A small `{ source, id, url, text }` value, formatted at the prompt boundary, would save later stages from parsing this string. I'd keep the SDK adapter and pagination as they are.

mldangelo-oai (commented):
Non-blocking QA notes on `4d3aeb5c`. The adapter is already a good size. Pagination, credential precedence, error handling, and removing Linear credentials before starting the patch subprocess passed with the real Linear SDK against a synthetic service. The full local suite passed: 1,292 tests, 11 skipped. The current GitHub checks are green.

I'd keep the adapter small and retain the issue ID and URL when connecting it to the common patch result. This covers intake and patch handoff, not a live Linear-to-PR run.

sdk/typescript/src/cli.ts
```diff
@@ -2463,14 +2479,50 @@ export async function main(
...
+                    ([name]) =>
+                      !/^(?:CODEX_SECURITY_)?LINEAR_(?:API_KEY|ACCESS_TOKEN)$/iu.test(
+                        name,
+                      ),
+                  ),
+                );
           exitCode = await runSkill(
             "fix-finding",
-            positionals,
+            [...positionals, ...imports],
```
zcrab-oai, line 2554: **[P1] Keep imported Linear issue content literal**

Could imported Linear issues take a literal-only path into `runSkill` instead of sharing positional file-or-text resolution? User-selected local files and remotely supplied issue descriptions have different trust boundaries, and imported descriptions must never trigger local file lookup. Please add Unix and Windows regressions that ensure imported ticket content is always preserved literally.

sdk/typescript/src/linear.ts
```diff
@@ -0,0 +1,125 @@
...
+      while (page.pageInfo.hasNextPage) await page.fetchNext();
+      issues.push(...page.nodes);
+      if (issues.length === 0) {
+        throw new CodexSecurityError(
+          `No open Linear issues matched project "${options.project}" and its filter.`,
+        );
+      }
+    } else {
+      for (const input of options.issues) {
+        const id = linearIssueIdentifier(input);
```
zcrab-oai: **[P2] Preserve the workspace selected by an issue URL**

Could we retain the workspace from a supplied Linear issue URL and verify it against the fetched issue's canonical URL? Looking up only the team-and-number identifier can silently return a different workspace's issue when the configured credential belongs elsewhere, causing the patch to target something other than the issue the user selected.

zcrab-oai (commented):
Found two input-handling issues affecting imported issue content and explicitly selected Linear workspaces.

sdk/typescript/src/linear.ts
kmbroai, line 1: I think this is similar to the linear client I added for the publish workflow. We might want to put those together.

kmbroai (dismissed):

sdk/typescript/src/cli.ts
ianw-oai: Addressed in b3e0df8. Patch now uses --linear-issue, --linear-project, --linear-filter, and --linear-api-key. Publishing keeps --to linear for the destination and uses --linear-project, with --project retained as an alias.

sdk/typescript/src/linear.ts
ianw-oai: Addressed in b3e0df8. Intake now returns { source, id, url, text }, and runSkill formats it only when building the prompt. The SDK adapter and pagination are unchanged.

sdk/typescript/src/cli.ts
ianw-oai: Fixed in b3e0df8. Imported issues bypass positional file resolution and remain literal. Added Unix and Windows path regressions that verify imported content is not probed as a local file. The full suite passed with 1,296 tests and 11 skips; CI is green.

sdk/typescript/src/linear.ts
ianw-oai: Fixed in b3e0df8. URL imports retain the selected workspace and compare it with the fetched issue URL. A mismatch stops the import. Tests cover a different workspace and a changed title slug in the same workspace.

sdk/typescript/src/linear.ts
ianw-oai: Addressed in b3e0df8. Intake and publishing now share API-key resolution and redirect-safe Linear SDK client setup. Existing publication authentication behavior and cancellation are preserved.

zcrab-oai (approved):

PR #492, opened 2026-08-17
fix: normalize Windows configuration paths

faizan-oai:
## Summary

Severity: P1–P2. Windows path spelling, environment-key casing, home-relative values, and machine-wide configuration could cause valid Codex Security settings to be missed or resolved against the wrong location. This makes those inputs resolve consistently while preserving existing configuration precedence.

## Changes

- Expand home-relative paths at CLI, runtime, comparison, and Deep Scan configuration boundaries.
- Match Windows environment keys and equivalent filesystem aliases using Windows semantics.
- Load machine-wide preflight configuration from ProgramData and recognize aliased project trust paths.
- Add focused coverage for path spelling, casing, aliases, and configured Python selection.

## Testing

- `pnpm run format` — passed.
- `pnpm run types` — passed.
- `pnpm run build` — passed.
- Focused Windows configuration-path regressions — 10 passed.

## Risk and rollout

The changes are limited to configuration discovery and path normalization. Existing precedence remains unchanged, and non-Windows path behavior is preserved apart from explicit home expansion where the CLI already accepts a path. No migration or rollout step is required.

## Public disclosure review


mldangelo-oai (commented):
Reviewed [6c185159fefe](https://github.com/openai/codex-security/commit/6c185159fefe26149365efe282065c7babb8b4bb) for correctness and simplification. One optional test simplification. The production changes preserve the intended configuration precedence.

sdk/typescript/tests-ts/api.test.ts
```diff
@@ -2211,6 +2211,64 @@ describe("CodexSecurity orchestration", () => {
...
+        join(ambientHome, "codex-security", "config.toml"),
+        "[deep_scan]\nworkers = 5\n",
+      );
+      const previousUserProfile = process.env["USERPROFILE"];
+      process.env["USERPROFILE"] = root;
+      try {
+        const client = new TestClient(
+          {},
+          {
+            environment: { CODEX_HOME: "~\\ambient-home" },
```
mldangelo-oai: Non-blocking simplification: `expandHome` now accepts the supplied environment, so this test can inject `USERPROFILE: root` beside `CODEX_HOME` instead of changing `process.env`. That removes the save/restore block and avoids shared process state during an async test. The new `comparisonEnvironment` home-expansion test can use the same approach.

mldangelo-oai (dismissed):
Rechecked 6c185159fefe26149365efe282065c7babb8b4bb. The earlier simplification suggestions remain non-blocking. Approving the reviewed head.

sdk/typescript/tests-ts/api.test.ts
faizan-oai: Addressed in 020c2b3. Both home-relative tests now inject `USERPROFILE` through the supplied environment alongside `CODEX_HOME`; neither mutates `process.env`. The two focused regressions pass, along with type and format checks.

mldangelo-oai (commented):
Re-reviewed 020c2b33060d72922db70a2ceb750fa2323dfea1, including the test-only update since the earlier approval. The previous review thread is addressed, and I found no new blocking issue.

CI is green on this head, including Windows, but the branch still conflicts with main in `sdk/typescript/src/cli.ts`. The resolution should retain both `resolveCliPath(directory, input)` and `lstat(path, { bigint: true })`. A fresh approval should be attached to the conflict-free replacement head after its checks pass.

Local validation: the diff whitespace check and focused Python preflight smoke checks passed. I did not run the full SDK suite or native Windows tests locally.

mldangelo-oai (approved):

PR #495, opened 2026-08-17
fix: reject Windows-ambiguous portable paths

faizan-oai:
## Summary

Severity: P1–P2. Windows reserved names, trailing dots or spaces, alternate streams, case-insensitive collisions, and other ambiguous components can make distinct logical paths address the same filesystem object. This rejects those inputs at the relevant contract, archive, repository, output, and multiscan boundaries.

## Changes

- Add one shared TypeScript validator for Windows-unsafe path components.
- Reject non-portable contract artifacts and case-insensitive artifact collisions.
- Reject ambiguous plugin ZIP entries, runtime paths, repository paths, and multiscan task IDs.
- Reject NTFS alternate-stream scopes on Windows while preserving valid colon-containing POSIX paths.
- Mirror contract checks in the bundled Python finalizer.

## Testing

- `pnpm run format` — passed.
- `pnpm run types` — passed.
- `pnpm run build` — passed.
- Focused portable-path regressions — 12 passed and 1 platform-opposite test skipped on Windows.

## Risk and rollout

The change intentionally rejects names that are not portable or are ambiguous on Windows. POSIX colon-containing path components remain accepted outside Windows. No stored-data migration is required.

## Public disclosure review


sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py
```diff
@@ -224,11 +237,23 @@ def _require_safe_relative_path(value: str, context: str, *, allow_dot: bool = F
         or "\0" in value
         or path.is_absolute()
         or ".." in path.parts
+        or any(_windows_unsafe_path_component(part) for part in value.split("/"))
```
mldangelo-oai: Non-blocking correctness note: this helper also validates [source locations](https://github.com/openai/codex-security/blob/5de0011508c473d98198c92c2ba39c132170e0a2/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py#L646-L647) and [include/exclude scopes](https://github.com/openai/codex-security/blob/5de0011508c473d98198c92c2ba39c132170e0a2/sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py#L1350-L1358). On Unix, paths such as `src/app.ts:fixture` and `src/CON.py` are valid and target normalization still accepts them, but finalization now rejects them. I reproduced the acceptance change against the base and this head. Keep the portable-name rule for scan artifacts and archives; separate it from validation of paths inside a Unix source repository.

sdk/typescript/_bundled_plugin/scripts/generate_rank_input.py
```diff
@@ -293,6 +293,21 @@ def path_is_excluded(path: Path) -> bool:
...
+def windows_stream_component(path: Path) -> str | None:
+    """Return the first NTFS alternate-data-stream component."""
+
+    if os.name != "nt":
+        return None
+    return next(
+        (
+            component
+            for component in path.parts
+            if component != path.anchor and ":" in component
```
mldangelo-oai, line 305: Non-blocking simplification: this new `windows_stream_component` is identical to the one in [`generate_in_scope_files.py`](https://github.com/openai/codex-security/blob/5de0011508c473d98198c92c2ba39c132170e0a2/sdk/typescript/_bundled_plugin/scripts/generate_in_scope_files.py#L18-L30). Share the predicate so both inventory paths enforce the same rule. The Windows-path helper module from [#435](https://github.com/openai/codex-security/pull/435) would be a natural place once that change is available.

mldangelo-oai (commented):
Reviewed [5de0011508c4](https://github.com/openai/codex-security/commit/5de0011508c473d98198c92c2ba39c132170e0a2) for correctness and simplification. One reproduced Unix-path regression and one optional deduplication. The source-path regression was checked against both the base and this head.

mldangelo-oai (dismissed):
Rechecked 5de0011508c473d98198c92c2ba39c132170e0a2. [The Unix source-path regression](https://github.com/openai/codex-security/pull/495#discussion_r3799032840) still reproduces at this head. Keeping it as a non-blocking follow-up. Approving the reviewed head.

sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py
faizan-oai: Addressed in 71cdc0f. Repository-relative source and scope paths are now validated separately from portable scan-local artifact paths in both the Python finalizer and TypeScript loader. The regression accepts `src/app.ts:fixture` and `src/CON.py`, while the portable artifact checks remain enforced. Focused tests, types, format, and Python compilation pass.

sdk/typescript/_bundled_plugin/scripts/generate_rank_input.py
faizan-oai: Agreed, but intentionally deferred until the shared Windows-path helper from #435 is available. Adding a new one-function module or a cross-PR dependency here would increase coupling without reducing behavior today; the two predicates remain identical in this branch and can be consolidated after integration.

sdk/typescript/_bundled_plugin/scripts/finalize_scan_contract.py
```diff
@@ -66,6 +66,19 @@
     "json": "exports/findings.json",
     "sarif": "exports/results.sarif",
 }
+WINDOWS_INVALID_PATH_CHARACTERS = frozenset('<>:"|?*')
+WINDOWS_RESERVED_DEVICE_NAMES = {
```
mldangelo-oai: Can you ask your agent to simplify this part? IMHO we should let codex handle this

faizan-oai: Addressed in 0a2b2a1. The Python check now uses one compiled regular expression matching the existing TypeScript helper instead of maintaining separate character, suffix, and reserved-name logic. The focused contract tests, typecheck, and formatting checks pass.

mldangelo-oai (approved):

mldangelo-oai (approved):

PR #513, opened 2026-08-17
fix(plugin): reject SQLite snapshot self-copy

### LINKED ISSUE #512: `snapshot_sqlite.py` hangs when source and destination are the same database
## Summary

`sdk/typescript/_bundled_plugin/scripts/snapshot_sqlite.py` can hang indefinitely when the destination refers to the same SQLite database file as the source.

The helper opens the source read-only, opens the destination separately, and then calls `source_connection.backup(destination_connection)`. SQLite's backup API does not make progress when both connections refer to the same database file.

## Reproduction

```bash
python3 - <<'PY'
import sqlite3

path = "/tmp/codex-security-snapshot-repro.sqlite3"
with sqlite3.connect(path) as connection:
    connection.execute("CREATE TABLE IF NOT EXISTS t(value TEXT)")
PY

python3 sdk/typescript/_bundled_plugin/scripts/snapshot_sqlite.py \
  /tmp/codex-security-snapshot-repro.sqlite3 \
  /tmp/codex-security-snapshot-repro.sqlite3
```

The second command does not return.

Aliases to the same file can reach the same condition, so comparing path strings alone would not be sufficient.

## Expected behavior

The helper should reject an existing destination that identifies the same filesystem object as the source before calling SQLite backup.

## Suggested fix

Use filesystem identity (`Path.samefile` / `os.path.samefile`) when the destination already exists, fail fast with a clear diagnostic, and add a regression with a process timeout so this cannot silently become an unbounded test hang.

sylvesterkaczmarek:
## Summary

Fail fast when `snapshot_sqlite.py` is asked to back up a SQLite database onto the same filesystem object.

Fixes #512.

## Problem

The helper opens the source database read-only, opens the destination separately, and calls `sqlite3.Connection.backup()`. When both connections refer to the same database file, the backup does not make progress and the command can remain stuck indefinitely.

A string comparison is insufficient because another path, including a hard-link alias, can identify the same database.

## Changes

- Rebased onto current `main` and moved the fix to the canonical plugin source under `plugins/codex-security`.
- If the destination already exists, use `Path.samefile()` to compare filesystem identity with the resolved source.
- Reject a self-copy through `argparse` before either SQLite connection is opened.
- Add Python regression coverage for both the exact source path and a hard-link alias, with a subprocess timeout so a recurrence fails instead of hanging.

## Validation

The regression covers:

- exact source path as destination: exits 2 with the new diagnostic;
- hard-link alias as destination: exits 2 with the new diagnostic;
- the normal snapshot path remains unchanged.

Pushed-head CI is required for the full repository validation.

## Risk

Low. The normal snapshot path is unchanged. The new check runs only when the destination already exists and rejects only destinations that the operating system reports as the same file as the source.

### DIFF
--- a/plugins/codex-security/scripts/snapshot_sqlite.py
+++ b/plugins/codex-security/scripts/snapshot_sqlite.py
@@ -16,6 +16,8 @@ def main() -> None:
     source = args.source.expanduser().resolve(strict=True)
     destination = args.destination.expanduser().absolute()
     destination.parent.mkdir(parents=True, exist_ok=True)
+    if destination.exists() and source.samefile(destination):
+        parser.error("source and destination must refer to different database files")
     with sqlite3.connect(f"{source.as_uri()}?mode=ro", uri=True) as source_connection:
         with sqlite3.connect(destination) as destination_connection:
             source_connection.backup(destination_connection)
--- a/plugins/codex-security/tests/test_snapshot_sqlite.py
+++ b/plugins/codex-security/tests/test_snapshot_sqlite.py
@@ -27,7 +27,4 @@ def test_sqlite_snapshot_rejects_source_identity_as_destination(tmp_path: Path) 
             check=False,
         )
         assert completed.returncode == 2
-        assert (
-            "source and destination must refer to different database files"
-            in completed.stderr
-        )
+        assert "source and destination must refer to different database files" in completed.stderr
--- a/plugins/codex-security/tests/test_snapshot_sqlite.py
+++ b/plugins/codex-security/tests/test_snapshot_sqlite.py
@@ -1,0 +1,33 @@
+from __future__ import annotations
+
+import os
+import sqlite3
+import subprocess
+import sys
+from pathlib import Path
+
+from workbench_test_support import SNAPSHOT_SCRIPT
+
+
+def test_sqlite_snapshot_rejects_source_identity_as_destination(tmp_path: Path) -> None:
+    source = tmp_path / "source.sqlite3"
+    alias = tmp_path / "source-alias.sqlite3"
+    with sqlite3.connect(source) as connection:
+        connection.execute("CREATE TABLE records (value TEXT NOT NULL)")
+        connection.execute("INSERT INTO records VALUES ('sealed')")
+        connection.commit()
+    os.link(source, alias)
+
+    for destination in (source, alias):
+        completed = subprocess.run(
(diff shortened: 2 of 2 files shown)


Hughhhhcoder:
I reproduced this against current `main`; the filesystem-identity fix and the same-path/hard-link regression still apply. One repository-layout update may help unblock this PR: the canonical plugin source is now `plugins/codex-security`, while `sdk/typescript/_bundled_plugin` is generated and current `AGENTS.md` says not to edit it directly. The equivalent current-main patch belongs in `plugins/codex-security/scripts/snapshot_sqlite.py`, with regression coverage in `plugins/codex-security/tests/test_workbench_setup_and_migrations.py`. I verified that placement locally with the required Ruff 0.16.1 lint/format checks, plugin source compatibility check, the 65-test workbench setup/migrations file, and the complete plugin suite (`1110 passed, 5 skipped, 110 subtests passed`). I will not open a competing PR; sharing this only to make the rebase/port easier.

sylvesterkaczmarek:
Thanks for the pointer. I rebased onto current `main` and moved the fix into the canonical `plugins/codex-security` source with Python regression coverage for the same-path and hard-link cases. The PR is mergeable again; the fork CI runs are currently waiting at `action_required` for workflow approval.

Hughhhhcoder (commented):
I checked the rebased head `3655b37a` locally. The focused regression passes (`1 passed`) and Ruff lint is clean, but the repository-pinned formatter (`ruff==0.16.1`) still reports `plugins/codex-security/tests/test_snapshot_sqlite.py` as unformatted: the multiline assertion at the end is collapsed to one line. Running `uv run --project plugins/codex-security --extra test ruff format plugins/codex-security/tests/test_snapshot_sqlite.py` (then the corresponding `ruff format --check`) should clear the gate before fork CI is approved.

sylvesterkaczmarek:
Thanks for catching the formatter gate. I ran the repository-pinned Ruff formatter on the regression, verified ruff format --check, reran the focused snapshot test (1 passed), and pushed 606ee03 to the same branch.

Hughhhhcoder (approved):
Rechecked the updated head `606ee037`. The focused self-copy/hard-link regression passes, Ruff lint and the repository-pinned format check are clean, and `git diff --check` passes. The formatter issue is resolved; I do not see a remaining correctness blocker in this scoped patch.

sylvesterkaczmarek:
Thanks for rechecking the updated head and the formatter gate. Appreciate the approval.

PR #562, opened 2026-08-18
fix(cli): emit ok:false envelope for failed history commands

pasmud:
## Summary

A failed structured history command exited non-zero and reported an error on stderr, but still emitted an `ok: true` envelope on stdout. This made automation that consumes structured stdout incorrectly classify failed commands as successful. This change makes the CLI emit an `ok: false` error envelope on failure instead.

## Changes

In `sdk/typescript/src/cli.ts`, the `history` helper previously caught workbench failures, wrote the error to stderr, set the exit code to 2, and returned `undefined`. Because it returned normally, the Incur command framework produced a successful envelope with `ok: true`. The helper now re-throws the error so the framework emits an `ok: false` envelope. The `scans match` catch block was updated the same way for consistency.

## Testing

- Added a regression test in `sdk/typescript/tests-ts/cli.test.ts` covering `scans show`, `scans list`, `scans compare`, and `scans match` with `--json`: each asserts exit code 2, empty stdout (no `ok: true`), and the error on stderr.
- `bun test --timeout 30000 tests-ts/cli.test.ts`: 140 pass, 0 fail.
- `tsc --noEmit`: clean.

## Risk and rollout

Low. The change only affects the structured envelope emitted for already-failing history commands; successful commands are unchanged. No schema or persisted-state changes.

## Public disclosure review


pasmud:
Hi @soyeon-oai , could you please review this PR when you have a chance?

mldangelo-oai (approved):
Looks great! thank you

PR #566, opened 2026-08-19
feat(cli): add read-only security verify-fix

kmbroai:
## Summary

Add a standalone, read-only workflow and dedicated discoverable skill for verifying whether an existing security fix actually resolves a reported finding.

## Changes

- Add `codex-security verify-fix` for finding text, saved scans or finding identifiers, Linear issues, and filtered Linear projects.
- Add a dedicated bundled `verify-fix` skill with user-facing discovery metadata, explicit remediation-verification guidance, a strict read-only boundary, and an evidence-backed result contract.
- Inline the concise verification skill and existing shared static finding assessment reference into one agent prompt without invoking or changing the `fix-finding` or `validation` skills, and without extra model tool calls.
- Include the verification skill and its discovery metadata in the existing exact-file bundled plugin contract.
- Run verification through one read-only Codex app-server agent, preserving existing patch behavior and batching multiple findings into one invocation.
- Disable repository-local MCP servers before verification starts, preserve user-configured integrations, and fail safely when a repository overrides an existing integration.
- Stream Codex reasoning, repository commands, tool calls, and commentary through the existing live activity dashboard, with plain stderr progress in CI and noninteractive terminals.
- Return evidence-backed `fixed`, `still_vulnerable`, or `inconclusive` results with machine-readable output and meaningful exit codes.
- Remove Linear credentials from the verifier environment, reject incomplete or mismatched results, and sanitize human-readable model output.
- Preserve durable regression coverage for dedicated skill loading, shared verification methodology, repeated Linear issues in one agent invocation, saved findings, read-only MCP isolation, result validation, and interactive/noninteractive progress.

## Testing

- `bun test --timeout 30000 tests-ts/cli-verify-fix.test.ts tests-ts/cli-skills.test.ts tests-ts/cli-patch-trust.test.ts tests-ts/cli-patch.test.ts tests-ts/cli-publish.test.ts tests-ts/linear.test.ts tests-ts/scan-dashboard.test.ts tests-ts/scan-activity.test.ts` — 172 passing.
- `bun test --timeout 30000 tests-ts/runtime.test.ts --test-name-pattern 'projects only the unchanged external payload from the source checkout'` — 1 passing, including all bundled skill files.
- `bun test --timeout 30000 tests-ts/cli.test.ts --test-name-pattern 'exposes Incur help, schemas, manifests, and completions|documents every public command argument and option|uses Incur version and command help'` — 3 passing.
- New `verify-fix` skill structure, frontmatter, UI metadata, and default invocation prompt — validated.
- `tsc --noEmit --pretty false`, `node --run format`, `node scripts/generate-models.cjs --check`, and `node --run build` — passing.
- Packed npm tarball contract validation — passing with 233 entries, including the new skill, its discovery metadata, and the existing shared assessment reference.
- Full installed-package smoke could not run locally because the existing trusted-owner check rejects `/tmp` in this environment.

## Risk and rollout

Additive CLI command and bundled skill; existing `fix-finding` and `validation` skills remain unchanged. Verification explicitly uses a read-only sandbox, blocks repository-local MCP execution, never receives Linear credentials, and fails closed when evidence is missing or does not match the requested finding. Progress stays on stderr, preserves machine-readable stdout, and falls back safely outside interactive terminals. Available after the next package release.

## Public disclosure review

### DIFF
--- a/sdk/typescript/_bundled_plugin/skills/fix-finding/SKILL.md
+++ b/sdk/typescript/_bundled_plugin/skills/fix-finding/SKILL.md
@@ -46,6 +46,8 @@ Use this guidance whenever reproducing the finding, running tests, or validating
 
 When the caller explicitly requests standalone verification-only mode, inspect the current checkout without creating, modifying, or deleting repository files. Do not implement a fix, apply a patch, commit changes, or modify issue trackers.
 
+In this mode, this section and the caller's requested JSON result envelope are the complete workflow and output contract. Skip the patch workflow, workbench remediation stages, patch-only outcome statuses, and report-writing instructions elsewhere in this skill. Apply shared evidence and safety requirements only when they are compatible with read-only verification; return `inconclusive`, not `blocked`, for an unresolved proof gap.
+
 For each supplied finding, identify the original source, broken control, sensitive operation, preconditions, and legitimate behavior. Trace those same security properties through the current implementation, accounting for moved or refactored code. Run the original reproducer, focused regression checks, bypass checks, and legitimate-behavior checks when they can run without writing to the repository. If a necessary check requires an unavailable writable directory, report its exact proof gap instead of weakening the read-only boundary.
 
 Return exactly one result for every expected identifier, in the supplied order:
--- a/sdk/typescript/_bundled_plugin/skills/fix-finding/SKILL.md
+++ b/sdk/typescript/_bundled_plugin/skills/fix-finding/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: fix-finding
-description: Use when the user explicitly asks to fix and verify a validated or plausible security finding, or to verify an existing fix without modifying the repository. Do not use as the primary trigger for full PR, commit, branch, patch, or repository scans.
+description: Use when the user explicitly asks to fix and verify a validated or plausible security finding. Do not use as the primary trigger for full PR, commit, branch, patch, or repository scans.
 ---
 
 # Fix Finding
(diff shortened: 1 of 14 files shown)


sdk/typescript/_bundled_plugin/skills/fix-finding/SKILL.md
```diff
@@ -42,6 +42,20 @@ Use this guidance whenever reproducing the finding, running tests, or validating
...
+
+For each supplied finding, identify the original source, broken control, sensitive operation, preconditions, and legitimate behavior. Trace those same security properties through the current implementation, accounting for moved or refactored code. Run the original reproducer, focused regression checks, bypass checks, and legitimate-behavior checks when they can run without writing to the repository. If a necessary check requires an unavailable writable directory, report its exact proof gap instead of weakening the read-only boundary.
+
+Return exactly one result for every expected identifier, in the supplied order:
+
+- `fixed`: current source or focused runtime evidence proves the original security boundary is closed and legitimate behavior remains intact.
+- `still_vulnerable`: current source or focused runtime evidence proves the same vulnerable path remains reachable.
+- `inconclusive`: repository mismatch, missing original context, unavailable checks, or another proof gap prevents a supported decision.
```
kmbroai: this is inconsistent with the below
`- outcome: `fixed`, `no_change`, or `blocked``

but it is a different wflow, so maybe that is ok

sdk/typescript/_bundled_plugin/skills/fix-finding/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: fix-finding
-description: Use when the user explicitly asks to fix and verify a validated or plausible security finding. Do not use as the primary trigger for full PR, commit, branch, patch, or repository scans.
+description: Use when the user explicitly asks to fix and verify a validated or plausible security finding, or to verify an existing fix without modifying the repository. Do not use as the primary trigger for full PR, commit, branch, patch, or repository scans.
```
ianw-oai: I think we can undo this change?

ianw-oai:
[P2] Preserve the documented verification exit codes

An authentication, rate-limit, or other Codex failure can return `1` [here](https://github.com/openai/codex-security/blob/833dd5cfa747e989435e203525164652a9fdd7d3/sdk/typescript/src/cli.ts#L2939) without a result. `verify-fix` documents `1` as a confirmed `still_vulnerable` result, so callers cannot distinguish a runtime failure from a verified vulnerability. Return `2` for verification failures, while preserving `130` and `143` for interruption. Add a test where `runCodex` returns `1`.

sdk/typescript/src/cli.ts
```diff
@@ -4001,6 +4421,19 @@ export async function readSkillCommandOutput(
...
     appServer?.input.write(`${JSON.stringify(request)}\n`);
   };
+  const startThread = (config?: JsonObject): void => {
+    send({
+      id: 2,
+      method: "thread/start",
+      // An explicit cwd makes Codex persist trust for a new project.
+      // Inherit the child process cwd and preserve the user's decision.
+      params: {
+        approvalPolicy: "never",
```
kmbroai: I think we have to match auto approver here. not "never"

kmbroai:
@codex review

sdk/typescript/src/cli.ts
```diff
@@ -4035,13 +4482,83 @@ export async function readSkillCommandOutput(
...
+            continue;
+          }
+          startThread(
+            repositoryServers.size === 0
+              ? undefined
+              : {
+                  mcp_servers: Object.fromEntries(
+                    [...repositoryServers].map((server) => [
+                      server,
+                      { enabled: false },
```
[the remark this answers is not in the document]
kmbroai: This is existing app-server trust behavior, not a new attack vector introduced by `verify-fix`.

The earlier patch app-server change in #477 explicitly documents: "User configuration and explicitly trusted project configuration remain active." Both `patch` and `verify-fix` launch the same app-server, inherit trusted project configuration, and can use a project-selected model provider when no provider is explicitly pinned. Therefore, the provider-redirection scenario described here already applies to the existing `patch` workflow.

Existing patch behavior is actually less restrictive: trusted repository-local MCP servers are allowed to execute during patching, whereas `verify-fix` explicitly disables them; the trust regression tests assert this distinction.

If project-controlled model providers should be rejected, that is a broader shared app-server hardening change that should be applied consistently to both workflows, not a regression specific to this PR.

ianw-oai (approved):

sdk/typescript/_bundled_plugin/skills/fix-finding/SKILL.md
kmbroai: old comment? i don't think this file is changed any more

PR #572, opened 2026-08-19
fix(scan): allow large git output during scan prep

### LINKED ISSUE #565: Scan prep crashes with "stdout maxBuffer length exceeded" on large repositories (0.1.11+)
## Summary

Since 0.1.11, `codex-security scan <dir> --diff <sha>` exits with code 2 within a second or two on repositories with large tracked-file counts, before any scan work happens. The only diagnostic is:

```
[00:00] Preparing scan
stdout maxBuffer length exceeded
```

Reproduced against a large public monorepo whose tracked-file listing is just over 1MB (roughly 14,000 files). The failure is independent of diff size: committed diffs changing fewer than a dozen files fail identically. Expected behavior: prep completes and the scan proceeds, as it does on 0.1.10 against the same repository.

## Cause

Commit ecc608e2 (PR #368, first released in 0.1.11) added a sparse-checkout guard to committed-diff scans that lists every tracked file:

```js
const tracked = await gitOutput(repository, ["ls-files", "-t", "-z"], signal);
```

`gitOutput()` in `dist/targets.js` runs git through `execFile` without a `maxBuffer` option, so Node's default 1MB cap applies:

```js
const { stdout } = await execFile(command.executable, ["-c", "core.fsmonitor=false", "-C", repository, ...args], {
    encoding: "utf8",
    signal,
    env: command.environment,
});
```

(Quoted from 0.1.14; 0.1.11's `gitOutput` differs slightly but equally omits `maxBuffer`.)

On a repository whose `git ls-files -t -z` output exceeds 1MB, Node kills the child and the scan dies. The code is unchanged on main as of 2026-08-19 (`sdk/typescript/src/targets.ts`).

## Reproduction

1. Create a repository whose tracked-file listing exceeds 1MB, for example 30,000 files with 60-character paths, and commit twice so a committed diff exists.
2. `npm i @openai/codex-security@0.1.14` (`--ignore-scripts` is fine)
3. `mkdir -m 700 /tmp/out` (the CLI refuses an output directory with group or other permission bits), then `OPENAI_API_KEY=placeholder codex-security scan <repo> --diff <first-commit-sha> --auth api-key --output-dir /tmp/out --max-cost 1 --json`
4. The CLI exits 2 with `stdout maxBuffer length exceeded` during "Preparing scan", before any API call. Repeat with `@openai/codex-security@0.1.10`: the prep phase completes.

Any real-world monorepo of that size reproduces it the same way.

## Suggested fix

Pass an explicit `maxBuffer` in `gitOutput()` (`dist/runtime.js` already uses `maxBuffer: Infinity` on one of its own call sites), or stream the listing instead of buffering it. Related: #249 tracks a similar whole-output buffering pattern for `git diff --binary` content digests.

Downgrading to 0.1.10 is not a clean workaround: it predates the head-revision guard PR #368 added, and its `scan --diff` save path has its own failure (#299).

## Environment

- @openai/codex-security 0.1.11 through 0.1.14 (0.1.10 and earlier unaffected by this specific failure)
- Node 22.x; reproduced on Linux x86_64 and macOS arm64

sylvesterkaczmarek:
## Summary

Allow committed-diff scan preparation to handle repositories whose tracked-file inventory exceeds Node's default `execFile` stdout buffer.

`validateCommittedDiffCheckout()` runs `git ls-files -t -z` to detect sparse checkouts. On large repositories, that output can exceed Node's default 1 MiB buffer and terminate scan preparation with `ERR_CHILD_PROCESS_STDIO_MAXBUFFER` before scanning begins.

Fixes #565.

## Changes

- remove Node's implicit 1 MiB stdout ceiling from the shared trusted Git execution path with `maxBuffer: Infinity`, matching the unbounded subprocess-buffer convention already used elsewhere in this package
- add an end-to-end regression that creates a synthetic clean Git repository whose tracked-file listing is confirmed to exceed 1 MiB
- exercise normal committed-diff target normalization and checkout validation against that repository
- keep Git invocation, sparse-checkout detection, target validation, and parsing semantics unchanged

## Testing

- added a focused synthetic-repository regression that produces more than 1 MiB of real `git ls-files -t -z` output and verifies committed-diff validation completes
- reviewed the branch against current upstream `main`; the diff is limited to one production line in `src/targets.ts` and the regression file
- fork workflow execution did not produce a completed verification result in this environment, so pushed-head CI remains the authoritative full typecheck, formatting, build, and test validation

## Risk and rollout

Low. Git command output was already collected in memory; this removes Node's default rejection threshold rather than changing the command, output format, or validation behavior. The change follows the same `maxBuffer: Infinity` convention already present in the SDK runtime and avoids introducing an arbitrary repository-size limit.

## Public disclosure review


mldangelo-oai:
Hi @sylvesterkaczmarek thank you so much for the contribution. Can you please merge in main and address the ci failures?

sylvesterkaczmarek:
> Hi @sylvesterkaczmarek thank you so much for the contribution. Can you please merge in main and address the ci failures?

Sure thing! Done, I merged current main into the branch. 

The newly triggered workflows are all action_required before any jobs execute, so there are no CI failure logs to address yet. Could you approve the workflows? I’ll address any actual failures once they run.

PR #587, opened 2026-08-21
fix: stream workbench payloads through stdin

faizan-oai:
## Summary

Severity: P1. Large scan recipes, preflight details, user context, comparison results, and recovery payloads can exceed the Windows process command-line limit before the workbench starts. This second split moves those structured payloads onto the stdin transport introduced by #500.

## Stack

This is PR 2 of 3. Review and merge in this order:

1. [#500](https://github.com/openai/codex-security/pull/500) — stdin-capable workbench transport, merged into `main`.
2. #587 — structured workbench and SDK payload migration (this PR), targeting `main`.
3. [#588](https://github.com/openai/codex-security/pull/588) — CLI validation-prompt migration, currently stacked on this PR's branch.

After this PR merges, #588 will be retargeted to `main`, updated with current `main`, simplified, rechecked, and approved before merging.

## Changes

- Send scan registration recipes, preflight issues, user context, and scan-comparison results through stdin.
- Update Deep Scan, progress, history, validation, and recovery workbench paths to consume stdin while preserving legacy argv transports.
- Decode workbench stdin as UTF-8 on every platform, and remove the redundant stdin-to-argv rewrite so option-like context remains literal text.
- Regenerate the checked-in bundled MCP runtime.
- Add focused regressions for maximum-size payloads, Unicode repository paths and preflight details, option-like Unicode context, empty-context clearing, and stdin forwarding.
- Avoid Bun's immediate synchronous `.cmd` timeout on hosted Windows while retaining the enclosing test timeout.
- Run the repository-index test probe asynchronously with its existing 10-second deadline to avoid a premature synchronous subprocess timeout reproduced locally and in Windows CI.
- Merge current `main` without rewriting branch history.

## Testing

- `pnpm run test` on the final head (seed `727007122`) — 1,500 passed, 59 platform skips, 0 failed across 84 files.
- Previously failing Windows CI shard, seed `3603982969` — 463 passed, 17 platform skips, 0 failed across 37 files after the async probe fix. The earlier full seeded run reproduced that premature timeout as its only failure.
- Affected stdin and Windows suites — 79 passed across 6 files. All 3 strengthened Unicode/option-like regressions failed before the production fixes and passed afterward.
- Affected Windows subprocess tests repeated five times — 10 passed.
- `pnpm run types` — passed.
- `pnpm run format` — passed.
- `git diff --check` — passed.
[...]

PR #500, opened 2026-08-17
fix: support stdin in workbench commands

faizan-oai:
## Summary

Severity: P1. Large helper payloads can exceed the Windows process command-line limit before the helper starts. This first split adds a reusable stdin-capable workbench process boundary; follow-up PRs move individual payloads onto it.

## Stack

This is PR 1 of 3. Review and merge in this order:

1. **#500** — stdin-capable workbench transport (this PR), targeting `main`.
2. **[#587](https://github.com/openai/codex-security/pull/587)** — structured workbench and SDK payload migration, currently stacked on this PR's branch.
3. **[#588](https://github.com/openai/codex-security/pull/588)** — CLI validation-prompt migration, currently stacked on #587's branch.

After each parent merges, the next PR will be retargeted to `main`, updated with current `main`, rechecked, and approved before merging.

## Changes

- Add an optional stdin payload to the TypeScript workbench runner.
- Use the existing bounded child-process transport while preserving credential filtering, cancellation, JSON parsing, and error reporting.
- Extend the runtime regression to cover a 64 KiB stdin payload together with a 5 MiB response.

## Testing

- `pnpm run types` — passed.
- `pnpm run format` — passed.
- `git diff --check` — passed.

## Risk and rollout

This PR only introduces the transport primitive; existing callers continue to omit stdin. The workbench and CLI payload migrations are split into follow-up PRs for separate review.

## Public disclosure review

- [x] No customer, partner, prospect, or user identities, data, or identifying details are included.
- [x] No credentials, personal data, private source, scan findings, or nonpublic links or tickets are included.
- [x] I reviewed the branch name, title, description, commits, changes, comments, logs, screenshots, attachments, and links for public disclosure.

PR #588, opened 2026-08-21
fix: stream CLI validation prompts through stdin

faizan-oai:
## Summary

Severity: P1. Large validation prompts can exceed the Windows process command-line limit before Codex starts. Send the prompt through stdin instead of placing the entire prompt in the argument list.

The preceding splits, #500 and #587, are merged. This PR targets `main` and includes main `22a72a7c8d06603ee5c1a69ff91a2fb2343f5f2c` through normal merges. Current head: `ae8c74eb452d98bf4a991cea65023a4b731ddee9`.

## Changes

- Pass validation prompts through stdin using `exec … -`, without adding public CLI flags or changing defaults.
- Preserve the live app-server protocol for both `patch` and `verify-fix`, and inherited terminal behavior for login/logout.
- Exercise an oversized Unicode prompt through a real child process and assert that app-server commands receive no raw stdin payload.
- Use the existing asynchronous process helper for both workbench comparison test probes after each reproduced a premature synchronous Python timeout on Windows. Both retain their 10-second deadlines, payloads, and result assertions.
- Retain the already-merged main implementations when resolving old stack conflicts. The complete four-file diff remains unchanged after the latest main synchronization; no further simplification was warranted.

## Testing

- Full local suites with seeds `12345` and `1894337911` on `eef22d26e16780cba285d58c1f902a1616292719` — each passed 1,511 tests, with 59 skipped and zero failures across 85 files.
- On synchronized head `ae8c74e`, focused CLI skill, verification, patch-trust, signal, and comparison tests — 54 passed, 1 skipped.
- Incoming main's SARIF export and recovery regressions on `ae8c74e` — 6 passed.
- The `verify-fix` transport assertion failed before the app-server input fix and passed after it. Both comparison probes passed three isolated repetitions and both full suites after switching to the async helper.
- `pnpm run types`, `pnpm run format`, and `git diff --check` — passed on `ae8c74e`.
- Fresh packed-package content checks and installed-package smoke — passed on `ae8c74e`, including NodeNext types, public imports, CLI, bundled plugin files, and a nested worker.
- All GitHub workflows on exact head `ae8c74eb452d98bf4a991cea65023a4b731ddee9` — passed: Node CI (all 24 jobs), container CI, container release checks, and PR title validation.

## Risk and rollout

Only non-app-server validation prompts change production transport. Patch and verification retain their JSON-RPC app-server connection; login/logout retain terminal input/output. The Unicode regression checks complete prompt delivery and EOF without a command-line-sized payload. No credential or permission protections change.

## Public disclosure review


mldangelo-oai (approved):
I don't know what this solves but I trust you

PR #589, opened 2026-08-21
fix(sdk): align scan cost session attribution with log inclusion

Drewwb:
## Summary

`ScanCostTracker.#readSessions` (`sdk/typescript/src/cost.ts`) and `belongsToScan` in `readScanLogs` (`sdk/typescript/src/scan-logs.ts`) both decide whether a Codex session belongs to a scan from the deep-worker output directory layout, but the cost-side check is weaker and disagrees with the log-side one. It matches any working directory whose path relative to `<scanDir>/artifacts/deep_discovery/workers` splits into two components ending in `"output"`, without `belongsToScan`'s guards. Two directories the log path deliberately rejects are therefore counted toward scan cost:

- a sibling such as `<scanDir>/artifacts/deep_discovery/output` relativizes to `../output`; and
- on Windows a directory like `D:\output` on a different drive than the scan relativizes to the absolute `D:\output`.

Both inflate reported cost/token totals, and because `ScanCostTracker` gates `--max-cost`, a scan can abort early against its budget on tokens it never spent.

## Changes

- `cost.ts`: reuse `belongsToScan`'s guarded, exact-path worker-output check — require a non-absolute relative path, reject a `..` first component, and confirm the directory is exactly `workers/<id>/output` — so cost attribution matches log attribution. Adds the `isAbsolute` import.
- `cost.test.ts`: add a regression test (`"excludes sessions beside the deep worker output directories"`) asserting a 1,000,000/1,000,000-token bystander session in `<scanDir>/artifacts/deep_discovery/output` is not counted; expected usage stays `input_tokens: 1_250, output_tokens: 12`.

## Testing

- Compared the two predicates' expressions on identical inputs under both `path.win32` and `path.posix`; after the change they agree on every case — real `workers/<id>/output` and `artifacts` directories are still included, and the sibling / cross-drive directories are excluded.
- Added the regression test above (fails on `main`, passes with this change).
- Note: I did not run the full `bun test` suite locally; the new test follows the existing `cost.test.ts` harness.

## Risk and rollout

Low. The change strictly narrows over-inclusion in scan cost attribution — legitimate worker, artifacts, and child-thread sessions are unaffected. No CLI, API, schema, or output-format changes.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/cost.ts
+++ b/sdk/typescript/src/cost.ts
@@ -1,5 +1,5 @@
 import { open, readdir } from "node:fs/promises";
-import { join, relative, sep } from "node:path";
+import { isAbsolute, join, relative, sep } from "node:path";
 import {
   scanActivityFromSessionEvent,
   type ScanActivity,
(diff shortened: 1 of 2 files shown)


mldangelo-oai (approved):
Thanks @Drewwb for the clear report, focused fix, and regression test! I reproduced the incorrect attribution and verified that this change excludes the unrelated sessions while still enforcing the budget for legitimate workers.

Reviewed 81eff9f4d6646a71a455cd939b9dc86b51b8d1aa: two full SDK runs passed (1,461 passed and 23 skipped each), along with the focused cost/log tests, typechecking, formatting, and build. Additional SDK budget tests and Node POSIX/Windows path checks also passed.

This fixes the reported directory-matching cases. I'll keep the shared-matcher cleanup, additional Windows regression coverage, and remaining timestamp/parentage differences in a separate follow-up so this fix stays focused.

PR #598, opened 2026-08-21
fix(cli): order update-notice prereleases by SemVer precedence

### LINKED ISSUE #534: Update notices use locale collation instead of SemVer prerelease ordering
## Summary

`isNewerVersion()` compares prerelease suffixes with:

```ts
new Intl.Collator("en", { numeric: true }).compare(candidate[4], installed[4])
```

SemVer prerelease precedence is not locale collation. Non-numeric identifiers are compared by ASCII code points and are case-sensitive, while numeric identifiers compare numerically and have lower precedence than non-numeric identifiers.

A concrete mismatch is case ordering:

- current: `1.0.0-A`
- npm latest: `1.0.0-a`

SemVer says `1.0.0-a` has higher precedence because ASCII `a` sorts after `A`. English `Intl.Collator` orders `a` before `A`, so the current update check suppresses a valid update. The reverse pair can produce a false update notice.

## Impact

This only affects prerelease versions, but update notices can be missing or incorrect for valid SemVer identifiers containing case differences. The same helper also delegates dot-separated identifier semantics to a locale API rather than implementing the SemVer rules it is intended to enforce.

## Expected behavior

Compare prerelease identifiers according to SemVer:

1. split on `.`;
2. numeric vs numeric: compare integer value;
3. numeric has lower precedence than non-numeric;
4. non-numeric vs non-numeric: compare ASCII lexically;
5. if all shared identifiers are equal, the longer identifier list has higher precedence.

## Suggested fix

Replace the `Intl.Collator` comparison with a small deterministic prerelease comparator and add regressions for case-sensitive ASCII ordering, numeric identifiers, numeric-vs-text precedence, and prefix-length ordering.

Hughhhhcoder:
## Summary

`isNewerVersion()` compared prerelease suffixes with `Intl.Collator("en", { numeric: true })`, which applies locale collation instead of the SemVer precedence rules the helper is meant to enforce. Non-numeric identifiers must compare by ASCII code points (case-sensitive), while numeric identifiers compare numerically and have lower precedence than non-numeric identifiers. The locale collator orders lowercase `a` before uppercase `A`, so a valid update such as `1.0.0-A` → `1.0.0-a` was suppressed, and the reverse pair produced a false update notice.

## Changes

- Replace the `Intl.Collator` comparison with a deterministic `comparePrerelease()` helper implementing SemVer §11.4 precedence: split on `.`, compare numeric identifiers numerically, give numeric identifiers lower precedence than non-numeric identifiers, compare non-numeric identifiers by ASCII code points, and let the longer identifier list win when shared identifiers are equal.
- Add a regression test covering ASCII case ordering, numeric-vs-text precedence, prefix-length ordering, and numeric width.

## Testing

- `bun test --timeout 30000 tests-ts/update-notice.test.ts` — 12 pass, 0 fail
- `pnpm run test` — 1504 pass, 28 skip, 0 fail
- `pnpm run types` — clean
- `pnpm run format` — clean

## Risk and rollout

Low risk. The change only affects update-notice detection for prerelease versions; stable version ordering is unchanged. No public CLI surface, schema, or default changes.

## Public disclosure review

Fixes #534

### DIFF
--- a/sdk/typescript/src/version.ts
+++ b/sdk/typescript/src/version.ts
@@ -159,12 +159,31 @@ function isNewerVersion(latest: string, current: string): boolean {
   if (candidate[4] === installed[4]) return false;
   if (candidate[4] === undefined) return true;
   if (installed[4] === undefined) return false;
-  return (
-    new Intl.Collator("en", { numeric: true }).compare(
-      candidate[4],
-      installed[4],
-    ) > 0
-  );
+  return comparePrerelease(candidate[4], installed[4]) > 0;
+}
+
+/** Compare two SemVer prerelease identifiers according to spec precedence. */
+function comparePrerelease(latest: string, current: string): number {
+  const candidate = latest.split(".");
+  const installed = current.split(".");
+  const length = Math.max(candidate.length, installed.length);
+  for (let index = 0; index < length; index += 1) {
+    const candidateId = candidate[index];
+    const installedId = installed[index];
+    if (candidateId === undefined) return -1;
+    if (installedId === undefined) return 1;
+    const candidateNumeric = /^\d+$/u.test(candidateId);
+    const installedNumeric = /^\d+$/u.test(installedId);
+    if (candidateNumeric && installedNumeric) {
+      const difference = Number(candidateId) - Number(installedId);
+      if (difference !== 0) return Math.sign(difference);
+    } else if (candidateNumeric !== installedNumeric) {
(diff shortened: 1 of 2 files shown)


mldangelo-oai (approved):
Thanks for tracking this down and adding focused regression tests! The case-ordering example makes the issue easy to understand. I verified the update-notice tests and typecheck on this head, and the fix stays nicely scoped. Happy to merge once CI is green.

PR #448, opened 2026-08-15
fix(runtime): recover malformed credential-home lock owners

mldangelo-oai:
## Summary

Recover stale credential-home locks when the recorded owner cannot identify a real process.

## Changes

- Accept only positive signed-32-bit integer process IDs before checking whether a lock owner is alive.
- Route malformed owner values through the existing stale-lock age check.
- Preserve live-owner, permission-denied, exited-process, and lock-release behavior.
- Retain the existing focused coverage and remove duplicated explanatory comments.
- Merge current main additively. The original runtime and test changes are byte-for-byte unchanged.

## Testing

- On `1cea3fe7`: `pnpm run types`, `pnpm run format`, and `git diff --check` passed.
- Build, pack, the static package contract, and a bounded installed-package check passed. The installed bytes match the archive; public SDK imports, package/plugin versions, `--version`, `--help`, and `patch --help` were checked.
- The source comparison confirms that the original two-file PR patch is unchanged after merging main.
- Historical validation at `52dfe1da`: 5 focused tests passed; the fixed-seed SDK suite passed 1,294 tests with 11 skips, 0 failures, and 9,156 assertions.

No new credential-lock scenarios, full SDK suite, or full installed-package smoke were run for this mechanical main integration. Three fresh source-only publication reviews and independent verification found no actionable issues. Post-push CI remains pending.

## Risk and rollout

Only malformed owner process IDs change behavior. Legitimate live locks, existing credential protections, and the 30-second incomplete-lock threshold are unchanged. Recovery of a genuinely reused live process ID remains separate work. No migration or configuration change is needed.

## Public disclosure review

- [x] No customer, partner, prospect, or user identities, data, or identifying details are included.
- [ ] No credentials, personal data, private source, scan findings, or nonpublic links or tickets are included.
- [x] I reviewed the branch name, title, description, commits, changes, comments, logs, screenshots, attachments, and links for public disclosure.

Newly authored material was reviewed for public disclosure. Existing automated review comments contain access-restricted report references; maintainer cleanup remains outstanding.

PR #611, opened 2026-08-21
fix(runtime): reclaim credential-home locks held by a reused PID

### LINKED ISSUE #228: Stale credential-home lock is never reclaimed when owner.json names a live or unusable pid, and the wait is silent and unbounded
## Summary

`recoverStaleCredentialHomeLock` decides whether a credential-home lock is stale from `process.kill(owner.pid, 0)` alone. Two separate problems follow:

1. Values that are not process identifiers at all — `0`, `-1`, fractional, out-of-range — are treated as "owner is alive" or throw a raw argument error, so the lock is never reclaimed.
2. A genuinely reused pid is indistinguishable from the original owner, so a lock left by a killed scan is never reclaimed either.

In both cases `acquireCodexSecurityCredentialHomeLock` spins at `CREDENTIAL_LOCK_POLL_MILLISECONDS` (25 ms) forever, with no message and no timeout. Every other stale path — missing or corrupt `owner.json` — has a 30 s escape hatch; the pid branch has none.

## Environment

- Reproduced against the exported `acquireCodexSecurityCredentialHomeLock`; no network or credentials required

## The code

`sdk/typescript/src/runtime.ts`:

```ts
if (isRecord(owner) && typeof owner["pid"] === "number") {
  try {
    process.kill(owner["pid"], 0);
    return false;
  } catch (error) {
    if (nodeErrorCode(error) !== "ESRCH") {
      if (nodeErrorCode(error) === "EPERM") return false;
      throw error;
    }
  }
} else if (Date.now() - metadata.mtimeMs < INCOMPLETE_CREDENTIAL_LOCK_MILLISECONDS) {
  return false;
}
```

`typeof … === "number"` admits `0`, `-1`, `0.5` and `2 ** 53`. POSIX gives the first two special meanings — `0` signals the caller's own process group and `-1` every process it may signal — so both always succeed and always report a live owner. A fractional or out-of-range value makes `process.kill` throw `ERR_INVALID_ARG_TYPE`, which is neither `ESRCH` nor `EPERM` and is rethrown raw out of a public API.

Note also that the age check is in the `else` branch, so a lock naming any accepted pid is exempt from it no matter how old it is.

## Reproduction

Plant a lock directory with a chosen `owner.json`, age it, and call the public acquire with a 1.5 s abort:

```
dead pid, 1h old (control)                     ACQUIRED in 2ms
no owner.json, 1h old (control)                ACQUIRED in 1ms
reused pid (live), 24h old                     ABORT_ERR: hung 1501ms
pid 1 -> EPERM, 24h old                        ABORT_ERR: hung 1502ms
pid 0 -> process group, 24h old                ABORT_ERR: hung 1500ms
pid -1 -> every process, 24h old               ABORT_ERR: hung 1502ms
pid 0.5 -> ERR_INVALID_ARG_TYPE                ERR_INVALID_ARG_TYPE: hung 1ms
```

Only the abort ended the hang; a 24-hour-old lock is never reclaimed.

The existing test "recovers credential-home locks left by exited processes" only covers a genuinely exited pid, so none of this is pinned.

## Why pid reuse is reachable, especially in the container

The lock is held for the whole scan (acquired in `#run` before runtime initialization, released in the outer `finally`), so it survives as long as the scan does — and outlives it if the process is `SIGKILL`ed by `docker stop`, an OOM kill, or a host reboot.

For the shipped container the collision is close to guaranteed:

- `Dockerfile` sets `CODEX_SECURITY_STATE_DIR=/output/.codex-security-state`
- `compose.yaml` bind-mounts `/output`, so the lock persists across container runs
- `compose.yaml` sets `init: true`, so container pids restart from 1 each run

A leftover `{"pid": 7}` from a killed run therefore names a live process in the next run, and every subsequent scan hangs with no output.

This may be related to #108 ("Run without output directory hangs computer … Randomly success, randomly failure"), whose randomness matches whether the leftover pid happens to collide. I have not confirmed that link.

## Suggested direction

Three separable pieces:

1. **Validate the pid** — only consult `process.kill` for `Number.isSafeInteger(pid) && pid > 0`, otherwise fall through to the existing age check. This is unambiguous and I have opened a PR for it.
2. **Survive pid reuse** — the pid alone cannot prove identity. The usual fix is a heartbeat: have the holder refresh the lock directory's mtime while it works, and reclaim a lock whose mtime has not advanced for some multiple of that interval. A plain absolute age bound is *not* safe here, because the lock is legitimately held for the entire scan and scans can run well over an hour (see #70, #164), so a fixed ceiling would let one scan steal another's lock. This needs a design decision, so I have deliberately left it out of the PR.
3. **Make the wait observable** — right now the user sees nothing at all. Even without (2), reporting which pid holds the lock after a few seconds, and where the lock directory is, would turn an unexplained hang into something actionable.

rohanpoudel2:
## Summary

Severity: P1. Fixes #228.

`recoverStaleCredentialHomeLock` decided whether a credential-home lock was stale from
`process.kill(owner.pid, 0)` alone, so a reused process ID was indistinguishable from the original
owner and the lock was never reclaimed. The 30 s stale-age escape hatch applied only in the `else`
branch of the process-ID check, which exempted a lock naming any accepted live process ID from
expiry no matter how old it was. Acquisition then polled every 25 ms with no timeout.

Merged PR #448 fixed the malformed-process-ID half of #228 and stated that recovery of a genuinely
reused live process ID remained separate work. This is that work.

The lock is held for a whole scan and outlives a killed process, so a leftover owner record can name
an unrelated live process on the next run. A persisted state directory combined with process IDs that
restart from a low number makes the collision easy to hit, and every later scan then waits forever.

## Changes

- Prove ownership by freshness instead of by process-ID liveness. A held lock refreshes its directory
  mtime on an unreferenced 5 s heartbeat, cleared when the lock is released. Heartbeat failures are
  swallowed so they cannot interrupt a scan.
- Apply one 30 s staleness horizon on every path, reusing the existing threshold rather than adding a
  second expiry. A lock whose heartbeat has lapsed is reclaimed regardless of who owns that process ID
  now. This replaces the age check that previously applied only to unidentifiable owners.
- Keep immediate recovery for a definitively exited owner, so an `ESRCH` result still reclaims at once.
- Keep waiting when a live owner is still heartbeating, so scans stay serialized.
- Guard the suspend/wake race: a machine that sleeps mid-scan wakes with a stale-looking mtime while
  its owner is still alive. Recovery of a lock whose process ID looks alive waits one heartbeat
  interval and abandons the reclaim if the mtime advanced. Owners that cannot be identified at all
  skip that wait, so exited and malformed owners recover as quickly as before. The wait honors the
  abort signal.
- Preserve the owner token check on release, the quarantine-then-remove sequence, the non-directory
  lock error, credential-home device and inode pinning, and cancellation.

No public CLI surface changes: no new commands, flags, accepted values, environment variables, or
defaults. The remaining half of #228, that the wait itself is silent, is left for a separate change to
keep this diff reviewable.

A second, test-only commit stops the heartbeat tests from replacing the global timer functions for
every caller. They now capture just the heartbeat registration, delegate all other timer calls to the
real implementation, and assert from the test body rather than from inside a mock. Previously an
unrelated `setInterval` call arriving while those spies were installed could fail an assertion
attributed to another test, or abort the lock wait under test.

## Testing

From `sdk/typescript/`, against this branch on `main` at `f222faf`:

- `pnpm run test` — four unseeded whole-suite runs, each 0 failures.
- `pnpm run test --seed 12345` — 1,576 passed, 29 skipped, 0 failed across 1,605 tests in 87 files.
- `pnpm run types` — passed.
- `pnpm run format` — passed, all matched files use Prettier style.
- `git diff --check` — passed.

New focused coverage in the existing credential-lock suite: a stale lock naming a live process is
reclaimed, a stale lock whose process inspection is denied is reclaimed, a fresh live owner still
blocks and still cancels on abort, a heartbeat keeps a held lock across the stale horizon, and
recovery is abandoned when the owner heartbeat advances. The heartbeat and race tests drive the
captured interval callback and `utimes` explicitly rather than sleeping, and the race test isolates
its module mock in a subprocess. The existing exited-process and malformed-owner tests still pass.

Behavior was also measured directly against the reproduction in #228, on this base before and after
the change:

| planted lock | before | after |
| --- | --- | --- |
| exited process ID, 24 h old | reclaimed in 3 ms | reclaimed in 2 ms |
| process ID `0` and `-1`, 24 h old | reclaimed in 2-4 ms | reclaimed in 2-3 ms |
| live process ID, 24 h old | never reclaimed | reclaimed in 5005 ms |
| live process ID, 30 days old | never reclaimed | reclaimed in 5007 ms |
| process ID `1`, permission denied, 24 h old | never reclaimed | reclaimed in 5004 ms |

Tests use synthetic fixtures and temporary directories. No live scan, network, or credentials were
required. Post-push GitHub Actions checks remain pending.

Two local check failures are unrelated to this branch and reproduce identically on unmodified `main`
in the same environment, which was verified by running them against a clean checkout:

- `audit:prod` reports one pre-existing high advisory. `package.json` and `pnpm-lock.yaml` are
  byte-identical to `main` on this branch, so no dependency changed here.
- `check:package` fails in its installed-package smoke because the packed tarball resolves
  `@openai/codex` at a prerelease version that the public registry does not serve.

Whole-suite runs on this machine also produce occasional 30-second timeouts in I/O-heavy plugin and
publication tests. Unmodified `main` produced the same class of timeout at a comparable rate, in
tests that do not touch credential locking, so it is a local resource limit rather than a regression
here.

## Risk and rollout

A lock whose owner has not heartbeated for 30 s is now reclaimable even when its recorded process ID
is live. That is the intended fix, and it is the only behavior change for a lock that is still being
heartbeated: a healthy owner is never displaced, because its heartbeat keeps the lock fresh and the
suspend/wake re-check backs off when the mtime advances. If two processes ever do overlap, release
still verifies the owner token and refuses to remove a lock it no longer owns.

Recovery of an unidentifiable or exited owner keeps its previous latency. Only the reused or
suspended live-owner case pays one heartbeat interval, replacing a wait that never ended. No
migration, configuration, schema, or public CLI change is required.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/runtime.ts
+++ b/sdk/typescript/src/runtime.ts
@@ -24,6 +24,7 @@ import {
   rm,
   rmdir,
   stat,
+  utimes,
   writeFile,
 } from "node:fs/promises";
 import { homedir, tmpdir } from "node:os";
@@ -74,6 +75,7 @@ const MAX_ZIP_EXPANDED_SIZE = 512 * 1024 * 1024;
 const MODEL_UNSAFE_PATH = /[\u0000-\u001f\u007f-\u009f\u2028\u2029]/u;
 const CREDENTIAL_LOCK_NAME = ".codex-security-scan.lock";
 const CREDENTIAL_LOGOUT_MARKER = ".codex-security-logged-out";
+const CREDENTIAL_LOCK_HEARTBEAT_MILLISECONDS = 5_000;
 const CREDENTIAL_LOCK_POLL_MILLISECONDS = 25;
 const INCOMPLETE_CREDENTIAL_LOCK_MILLISECONDS = 30_000;
 const MAX_PROCESS_ID = 2_147_483_647;
@@ -1065,7 +1067,7 @@ export async function acquireCodexSecurityCredentialHomeLock(
       throw error;
     });
     if (existingLock !== null) {
-      if (await recoverStaleCredentialHomeLock(lock)) continue;
+      if (await recoverStaleCredentialHomeLock(lock, signal)) continue;
       await delay(CREDENTIAL_LOCK_POLL_MILLISECONDS, undefined, { signal });
       continue;
     }
@@ -1078,7 +1080,7 @@ export async function acquireCodexSecurityCredentialHomeLock(
       await mkdir(lock, { mode: 0o700 });
     } catch (error) {
       if (nodeErrorCode(error) !== "EEXIST") throw error;
-      if (await recoverStaleCredentialHomeLock(lock)) continue;
+      if (await recoverStaleCredentialHomeLock(lock, signal)) continue;
       await delay(CREDENTIAL_LOCK_POLL_MILLISECONDS, undefined, { signal });
       continue;
     }
@@ -1094,9 +1096,18 @@ export async function acquireCodexSecurityCredentialHomeLock(
       throw error;
     }
 
+    const heartbeat = setInterval(async () => {
+      try {
+        const now = new Date();
+        await utimes(lock, now, now);
+      } catch {}
+    }, CREDENTIAL_LOCK_HEARTBEAT_MILLISECONDS);
+    heartbeat.unref();
+
     let released = false;
     return async () => {
       if (released) return;
+      clearInterval(heartbeat);
       await requireSecureCredentialHome(codexHome, {
         ...securityOptions,
         expectedDevice,
@@ -1116,7 +1127,10 @@ export async function acquireCodexSecurityCredentialHomeLock(
   }
 }
 
-async function recoverStaleCredentialHomeLock(lock: string): Promise<boolean> {
+async function recoverStaleCredentialHomeLock(
+  lock: string,
+  signal?: AbortSignal,
+): Promise<boolean> {
   const metadata = await lstat(lock).catch((error: unknown) => {
     if (nodeErrorCode(error) === "ENOENT") return null;
     throw error;
@@ -1135,17 +1149,13 @@ async function recoverStaleCredentialHomeLock(lock: string): Promise<boolean> {
     if (nodeErrorCode(error) !== "ENOENT" && !(error instanceof SyntaxError)) {
       throw error;
     }
-    if (
-      Date.now() - metadata.mtimeMs <
-      INCOMPLETE_CREDENTIAL_LOCK_MILLISECONDS
-    ) {
-      return false;
-    }
   }
 
   // Only positive signed-32-bit PIDs identify an owner. Other values can name
   // process groups or fail argument validation, so use the stale-age check.
   const ownerPid = isRecord(owner) ? owner["pid"] : undefined;
+  let ownerExited = false;
+  let ownerIsAlive = false;
   if (
     typeof ownerPid === "number" &&
     Number.isInteger(ownerPid) &&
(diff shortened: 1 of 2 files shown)


mldangelo-oai (approved):
Thank you!

PR #621, opened 2026-08-24
fix(sdk): reject reversed finding line ranges

### LINKED ISSUE #560: TypeScript contract loader accepts reversed finding line ranges
## Summary

The canonical TypeScript contract loader accepts a finding location whose `endLine` is less than its `startLine`, even though the bundled Python producer rejects that same location before sealing.

This creates a producer/consumer validation mismatch: an externally constructed, correctly resealed completed-scan bundle can load through `loadContract()` with a line range that Codex Security itself would never finalize.

## Reproduction / evidence

Current upstream `main` at `37bf87a692fc72d41f7312cc48808d699d204fba` validates ordinary finding locations in two different ways.

The Python finalizer uses `_validate_location()` and requires:

```py
end = location.get("endLine", start)
...
if not isinstance(end, int) or end < start:
    raise ContractError(...)
```

The shared JSON schema only constrains `startLine` and `endLine` independently to positive integers. The TypeScript `validateCanonicalContract()` then validates each location path but does not compare the two line numbers.

A deterministic sealed-contract reproduction is:

1. copy the bundled `examples/completed-scan` fixture;
2. change the first finding location to `startLine: 41, endLine: 40`;
3. recompute the sealed `findings.json` SHA-256 in `scan-manifest.json`;
4. call `loadContract()`.

Current behavior: the bundle loads successfully.

Expected behavior: the SDK loader rejects the same reversed line range the producer rejects.

## Root cause

The relational `endLine >= startLine` invariant lives in Python finalization but was not mirrored in the TypeScript canonical-validation pass. JSON Schema Draft 2020-12 does not express this sibling-value comparison in the existing schema without a nonstandard extension.

## Suggested fix

During `validateCanonicalContract()`, reject any ordinary finding location whose defined `endLine` is less than `startLine`.

Add a focused regression that mutates and reseals the real bundled completed-scan fixture, verifies the reversed range is rejected, and retains a valid multi-line range control.

## Impact

This is contract-integrity correctness. Invalid ranges can reach SDK consumers and downstream source-link/reporting logic even though the canonical producer rejects them, so producer and consumer disagree about what constitutes a valid sealed finding.

v1shay:
## Summary

Reject finding locations whose `endLine` is less than `startLine` during canonical contract validation. This keeps the TypeScript loader aligned with the bundled Python producer, which already rejects reversed ranges.

## Changes

- Validate the relational `endLine >= startLine` invariant for ordinary finding locations.
- Add a regression that mutates and reseals the bundled completed-scan fixture with a reversed range and verifies `loadContract()` rejects it.

Fixes #560

## Testing

- `bun test --timeout 30000 tests-ts/contract.test.ts` — 44 passed
- `pnpm exec prettier --check src/contract.ts tests-ts/contract.test.ts` — passed
- `pnpm run types` — passed
- `git diff --check` — passed

### DIFF
--- a/sdk/typescript/src/contract.ts
+++ b/sdk/typescript/src/contract.ts
@@ -480,6 +480,14 @@ function validateCanonicalContract(
           { cause: error },
         );
       }
+      if (
+        location.endLine !== undefined &&
+        location.endLine < location.startLine
+      ) {
+        throw new ContractValidationError(
+          `${locationContext}.endLine: must be greater than or equal to startLine.`,
+        );
+      }
     }
 
     const fingerprint = `codex-security/v1:sha256:${sha256Text(
--- a/sdk/typescript/tests-ts/contract.test.ts
+++ b/sdk/typescript/tests-ts/contract.test.ts
@@ -1092,6 +1092,21 @@ describe("canonical scan contract", () => {
     }
   });
 
+  test("rejects finding locations with reversed line ranges", async () => {
+    const scanDir = await copyExample();
+    const findingsPath = join(scanDir, "findings.json");
+    const findings = await readJson(findingsPath);
+    findings["findings"][0]["locations"][0]["endLine"] = 40;
+    await writeJson(findingsPath, findings);
+    await reseal(scanDir);
+
+    await expect(
+      loadContract(scanDir, { pluginRoot: PLUGIN_ROOT }),
+    ).rejects.toThrow(
+      "findings.findings[0].locations[0].endLine: must be greater than or equal to startLine.",
+    );
+  });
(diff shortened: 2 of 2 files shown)


mldangelo-oai:
Thanks for putting this together. The Python finalizer already rejects reversed line ranges before a scan is sealed, so this cannot happen in normal scan output. Reproducing it requires manually changing and resealing a bundle. Adding the same check in the SDK would duplicate that validation and could reject previously accepted bundles, so the extra check does not seem worth the compatibility tradeoff. Closing this one, but I appreciate you taking the time to investigate.

PR #623, opened 2026-08-24
fix(plugin): preserve committed diff scan integrity

mldangelo-oai:
## Summary

Keep committed-diff scans tied to the Git objects and replacement view the user selected. Validate a complete snapshot before changing scan state, then recheck it when taking ownership.

## Changes

- Validate batch object framing and content without fetching missing objects. Freeze replacement refs for each inspection, including packed and custom-prefix refs.
- Carry the selected content digest through workspace registration and scan startup; preserve legacy records without silently rebinding an existing selection.
- Keep long inspections outside the database lock, recheck the snapshot inside it, and reuse an already-started matching scan when another caller wins.
- Merge current main. Remove three unused database wrappers, a duplicate SQL fixture helper and the mocked plugin installer; simplify repeated predicates and promise setup. Five functions and 74 net lines removed.
- Set both bundled-plugin markers to 0.1.75 and test real cache upgrades from 0.1.60 and 0.1.74.

## Testing

- Final four focused suites, seed 623: 108 passed. Baseline: 105 passed. Targeted cache and fallback coverage: 9 passed.
- Types, generated models, formatting, build, Ruff, Python 3.10 grammar and decoded runtime syntax passed.
- Fresh 282-entry npm artifact check and full installed smoke passed, including 118 plugin files and the nested Codex worker.
- Full suite, native Windows and alternate Git versions were not rerun. Local Git is 2.54.0. New-head CI is left for a separate pass.

- Final main refresh (`fd98a90`): package 0.1.21 includes the MCP launcher-permission fix; SDK implementation and bundled payload are unchanged. Types/model generation, formatting, build, 28 focused tests, static artifact verification and full installed-package smoke passed, including MCP initialization. CI was not awaited.

## Risk and rollout

This is the base of the restoration, local-diff, budget-attribution and excerpt-scope stack. Keep the stack order and use a distinct newer cache version for each changed payload. Existing credential, path and scan-integrity checks remain. No public CLI surface, package publication or PR merge is included.

## Public disclosure review

Historical commit contact metadata and account-gated links in automated comments remain, so the second attestation is unchecked. This update uses synthetic fixtures and a GitHub noreply commit identity; other authors' comments are unchanged.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

kmbroai (commented):
Reviewed head `4e478501a9283fa0aed4e1e99ab55db00529d134` for correctness, necessity, and simplification.

### Necessity and correctness

This fixes a real integrity boundary: a selected commit/range must remain bound to the bytes inspected, including Git replacement behavior, rather than merely to revision names. The narrow temporary Git view is a reasonable approach because it freezes replacement configuration without copying an entire repository or holding the global SQLite writer lock while streaming changed blobs.

I traced the strict `cat-file` parser, digest/object identity construction, replacement-ref capture, committed-scan registration and completion, and the inspect/reacquire/revalidate transaction path. The parser checks object identity, framing, sizes, missing objects, and unexpected output; changed blobs are streamed through private temporary files. The locked recheck compares the frozen object identity rather than repeating expensive content reads. If lock contention intervenes, external inspection is repeated before the transaction proceeds. These are necessary parts of the fix, not redundant defensive checks.

The compatibility path is also appropriately narrow: recovering a legacy NULL digest does not generally authorize changing an already submitted selection. Current equivalent-workspace joins and state-free inspection retain their separate behavior. I did not reproduce a new correctness blocker at this head.

### Simplification opportunities

This PR combines the object protocol, replacement-view construction, writer-lock scheduling, legacy recovery, and MCP timeout/fallback behavior. Keep the strict parser and immutable inspection result small and independently testable; avoid turning them into a general repository snapshot framework. The transaction helper should own the unlock/inspect/relock/revalidate invariant once, with callers only applying scan-specific updates.

If splitting is still practical, isolate the parser/replacement-view tests from the transaction/legacy compatibility changes while retaining the end-to-end integrity tests. The five-minute timeout belongs only to operations that actually perform the larger work; the implementation's unchanged shorter default is preferable to a blanket timeout increase.

Coordinate this with the restoration/inventory stack (#626 and #630) and assign a fresh integrated plugin version. Several independent branches currently reuse prospective version numbers, which is not a safe installation-cache contract.

### Verification

Ran the focused workbench, compact-diff, and bundled timeout/fallback suites: **105 passed, 0 failed**. These include committed replacement/object handling, stale selections, concurrent/legacy recovery, and actual bundled MCP paths. Execution was on Linux with Git 2.55 and cached dependencies; this is not independent validation of Git 2.39, native Windows, or a live model scan.

mldangelo-oai:
@codex review

PR #635, opened 2026-08-24
fix(windows): keep credential state reusable across scans

faizan-oai:
## Summary

Keep the default Windows sandbox profile from invalidating credential-home ancestry for the next command, and allow credential ACL inspection to complete while ordinary private cache and lock files change.

## Changes

- Remove state-directory-wide sandbox write access. Retain root reads, workspace writes, and explicit read-only credential-home protection; the CLI and workbench continue maintaining state as the current user.
- Replace separate Node and PowerShell descendant counts with one native ACL traversal and a checked completion count. Validate inspected entries instead of requiring two independently timed directory snapshots to match.
- Integrate current main while preserving its targeted three-attempt retry for interrupted enumeration and descriptor-callback draining. Incomplete traversal cannot emit or satisfy the completion marker. Unrelated subprocess failures and descriptor-validation errors are not retried as enumeration interruptions; existing ACL-repair behavior is unchanged.
- Preserve owner, ancestor, ACL, symbolic-link/junction, incomplete-output, and subprocess-failure checks. Keep main's SQLite credential-lock implementation and asynchronous scan-root alias test unchanged.
- Extend existing runtime/configuration regressions and document recovery with fresh private state while preserving old reports. Existing unsafe ancestor ACLs are not automatically rewritten.

## Testing

Latest local verification covers main `f5e98f8` plus the subsequent README-only update `a4413fa`. The final README reconciliation leaves runtime code and tests unchanged.

- Focused retry/callback tests: **5 passed, 0 failed**.
- Five affected suites, seed `12345`: **166 passed, 25 skipped, 0 failed**.
- Package types, formatting, build, and `git diff --check`: passed.
- Seven native controls: baseline and process-local ConstrainedLanguage succeed; one interrupted root or nested enumeration retries successfully; persistent interruption exhausts three attempts; unrelated or missing-root errors fail without retry. No failed traversal emits completion. Faults are injected deterministically into the producer while retaining the actual PowerShell catch and TypeScript consumer. No machine-wide policy was changed on the local test workstation.
- A separate synthetic native Codex/SDK `0.149.1` comparison confirms that the old state-wide write grant still invalidates subsequent credential-ancestry checks. The narrower profile preserves repository reads, workspace writes, credential reuse, and a second native command; pre-existing host ancestor ACLs remain unchanged.
- Full package run, seed `12345`: **1,856 passed, 63 skipped, 5 failed** across 97 files. All failures are in unchanged release-workflow tests: four manual-label reconciliation cases returned missing Bash exit statuses, and one invalid-title case returned null status. An isolated diagnostic reproduced the four launch failures as `EPERM` before a process ID was assigned; the invalid-title case passed in isolation. The cause of the launch failures is not established, and these are not being reported as proven flakes.
- Plain unseeded full package run, reported seed `2882182476`: **1,857 passed, 63 skipped, 4 failed** across 97 files. The same four manual-label reconciliation cases failed; there were no other failures. Both full runs used the package's unchanged timeout and no diagnostic preload. These local results are not passing full runs.
- GitHub CI on `39b942a`: **31 successful checks, 5 skipped, 0 failed or pending**. [Node CI](https://github.com/openai/codex-security/actions/runs/33030228405) passed all 27 jobs. Both the [Windows Node 22](https://github.com/openai/codex-security/actions/runs/33030228405/job/98380836555) and [Windows Node 24](https://github.com/openai/codex-security/actions/runs/33030228405/job/98380836621) shard logs explicitly pass all four locally failing release fixtures. This confirms those cases executed in CI, but does not establish the cause of the local launch failures.

Earlier branch verification included an actual CLI/native Codex localhost replay: repository reads succeeded, credentials remained valid after cleanup, and the next invocation reached a synthetic endpoint. The fixture omitted required scan artifacts, so this proved lifecycle behavior, not a completed security scan. The earlier full seeded run had two failures (an MCP timeout and a null Python subprocess status) and one shutdown follow-on error; isolated reruns passed. Those historical failures remain recorded and are not attributed to the new main integration. CI results for the new head are listed above.

## Risk and rollout

The default profile no longer grants write access to the entire state directory on any platform. Approved requests can still grant additional permissions. Native tests cover workspace writes and credential reuse; CI must cover the remaining platforms. No new CLI flags, dependencies beyond main's existing lockfile, automatic ACL migration, sandbox-identity allowlist, or lock redesign is introduced. Users with an already-invalid state ancestor must follow the documented fresh-state recovery; old reports and scan history are preserved separately.

## Public disclosure review


faizan-oai:
@mldangelo-oai — could you review this credential-lifecycle fix? It narrows the default sandbox write roots and removes the cross-process ACL-count race while retaining the existing credential protections. Native boundary and repeated-start controls pass; the description records the two broader-suite failures and the recovery limits.

@codex review the current head `f033704d97d605d2011976476c3a18fbe4ebec6d`.

mldangelo-oai (dismissed):
I think you are adding too much to the readme and this is hard for me to test, but I trust it solves an issue

faizan-oai:
Condensed the recovery guidance to one short paragraph beside the existing state-directory instructions in `6a58daf`; no runtime or test code changed in that follow-up.

For a focused reproduction check on Windows, from `sdk/typescript` with dependencies installed:

```powershell
bun test --timeout 30000 --seed 12345 tests-ts/runtime.test.ts tests-ts/config.test.ts --test-name-pattern 'Windows credential|Windows credentials|scan helpers to write'
```

This passed 19 tests locally, including the actual sandboxed helper, denied state-directory writes, subsequent credential reuse, and live/broken junction rejection. It uses synthetic fixtures and does not need a model-backed scan.

The initial Windows verification failure was the package smoke test's npm dependency installation timing out before the installed CLI was run. The new head's CI is rechecking installation as well as the tests.

faizan-oai:
@codex review the current head `6a58daff3724aa7bc4961ebdfea41da3280f7242`. The only change since the reviewed runtime commit is the shortened, relocated README recovery paragraph.

kmbroai (commented):
## Critical review

Reviewed `6a58daff3724aa7bc4961ebdfea41da3280f7242`. **The lifecycle fixes are justified, and the changes reduce unnecessary machinery. I found no blocking correctness issue in the reviewed diff.**

### Necessity and correctness

The two changes address different real causes: the broad state-directory write grant can affect credential ancestry, while the old descendant verifier compares two directory enumerations taken at different times. Normal cache/lock churn makes equality of those counts an unreliable success condition.

Removing the [state-wide write root](https://github.com/openai/codex-security/blob/6a58daff3724aa7bc4961ebdfea41da3280f7242/sdk/typescript/src/api.ts#L3344) is more appropriate than adding exceptions to the credential-owner checks. I traced the SDK flow: the model's working directory is the scan workspace, while registration, thread updates, and final persistence are driven by the host-side workbench calls. This distinction is essential; the review is not assuming that a sandboxed process can still write arbitrary state because it runs under the same account.

The [single native inspection](https://github.com/openai/codex-security/blob/6a58daff3724aa7bc4961ebdfea41da3280f7242/sdk/typescript/src/runtime.ts#L694) preserves separate ancestor/home/descendant checks and requires both successful subprocess completion and a matching completion record. It rejects missing/truncated completion and output after completion. It therefore removes the cross-process counting race without simply treating a partial traversal as success. This is an inspection of entries encountered during traversal, not an atomic snapshot of a concurrently changing directory; the code and documentation should not promise the latter.

### Simplification and rollout

Keep the deletion of the Node-side walk and retries. Reintroducing a second enumeration would restore the race and duplicate link/type checks. The completion marker is a small internal protocol; it does not need a general stream-protocol abstraction or another lock redesign.

The default-profile change applies on every platform, not only Windows. Keeping the explicit credential-home read protection and tests for workspace writes/state writes is therefore important. The recovery paragraph also correctly avoids silently rewriting pre-existing ancestor ACLs or pretending a fresh state directory retains the same history. I would keep that explanation beside the state-directory setting, not expand it into a second authentication setup guide.

### Verification and limits

Ran `api-credentials.test.ts`, `api-preflight-config.test.ts`, `config.test.ts`, and `runtime.test.ts` with seed 12345: **170 passed, 13 platform skips, 0 failed**, Bun 1.3.14/Linux with locally available dependencies. Initial runs exposed sandbox ownership mapping and group-writable fixture directories; rerunning outside the sandbox with a normal `022` umask passed without source changes.

This validates the cross-platform code and synthetic ACL-stream controls available on Linux. It does **not** independently validate native Windows ACL inheritance, ConstrainedLanguage behavior, or repeated real Windows scan startup; those remain platform-specific evidence to preserve from the PR's native controls and CI.

mldangelo-oai (approved):

faizan-oai:
@codex review the current head `39b942aa0ec9f98a420d84d64289276a672f9790`.

This integrates main's interrupted-enumeration retry and callback draining while retaining the single-pass ACL completion checks and narrower state permissions. The README follows main's shorter layout with the existing concise recovery guidance. The PR description records the passing focused/native checks, broader local test failures, and checks still in progress.

Please keep feedback self-contained and omit nonpublic report links.

mldangelo-oai (approved):

PR #638, opened 2026-08-24
fix: limit finding comparison input size

ianw-oai:
## Summary

Large finding comparisons can exceed the Codex input limit and leave matching incomplete. Each comparison included all earlier findings in one request. Split large comparisons into smaller requests so the total finding count does not determine the size of a single request.

## Changes

- Compare every earlier/later finding pair in independent batches.
- Join confirmed groups across batches in code. Keep the existing rules for uncertain pairs.
- Keep full finding text. Report an error if one pair exceeds the Codex input limit.
- Add regression tests for large inputs, cross-batch groups, uncertainty, cancellation, and input limits.
- Clarify that component retries create a separate report.

## Testing

- Full suite in two orders: 1,636 passed, 30 skipped, no failures in each run. Used a 90-second per-test timeout.
  - `pnpm run test --seed 12345 --timeout 90000`
  - `pnpm run test --timeout 90000`
- `pnpm run types`, `pnpm run format`, and `pnpm run build` passed.
- 89 focused tests and 100 synthetic group-merging checks passed.

## Risk and rollout

No CLI options, result formats, or dependencies change. Large comparisons make more model requests. Small comparisons still use one request. Component retry behavior is unchanged. This change does not address process shutdown after a completed summary.

## Public disclosure review


sdk/typescript/src/scan-comparison.ts
```diff
@@ -113,25 +115,109 @@ export async function matchScanFindingsInternal(
...
-    throw new CodexSecurityError("Scan comparison returned invalid JSON.", {
-      cause: error,
-    });
+  const comparisons: ScanComparisonResult[] = [];
+  const allowHistoricalUncertainty =
+    options.allowHistoricalUncertainty ?? false;
+  const outputSchema = z.toJSONSchema(comparisonSchema, {
+    target: "openapi-3.0",
+  });
+  for (const batch of comparisonBatches(input)) {
```
mldangelo-oai: [P1] Keep batched post-scan matching inside the requested cost bound

`matchScanFindingsInternal` is also used by ordinary `CodexSecurity.run()` history matching, not just component scans. That path calls `matchCompletedScan` only after `tracker.stop(...)` and `complete-scan` have already persisted the final cost, and it does not pass `maxCostUsd` or any remaining budget into this loop. With four synthetic 300 KB historical findings and one 5 KB new finding, I reproduced four additional model turns after completion; the README currently allows only one extra call when a cost limit is set. Larger histories can therefore generate many untracked, uncapped turns despite `--max-cost`, without appearing in the recorded scan cost. Please either keep batching specific to explicitly unbounded component matching or propagate and enforce the remaining scan budget before starting subsequent batches, and update the single-call documentation accordingly.

mldangelo-oai (commented):
Reviewed current head 596a73a. The batching and group-merging checks pass, but the shared matcher now starts uncapped post-completion model calls for cost-limited ordinary scans; see the inline comment.

kmbroai (approved):

sdk/typescript/src/scan-comparison.ts
ianw-oai: Fixed in 7a85033f. Automatic post-scan history matching now keeps its existing one-call limit, including scans with --max-cost. If the full comparison exceeds the input limit, it reports a warning without starting a model call or changing the completed scan. Explicit comparisons and component matching can still use batches.

Added tests for a large comparison that fits in one call and one that exceeds the input limit. The 206 focused matcher, component-scan, and API tests pass, as do type, format, and build checks.

PR #653, opened 2026-08-26
feat(typescript): add findings service plumbing

kmbroai:
## Summary

Add the first stage of a Docker-hosted findings service in the TypeScript SDK. The service initializes the existing SQLite workbench and exposes two explicitly mocked HTTP endpoints, without adding a CLI command.

## Changes

- Omit the unused bulk-dedupe endpoint from the initial API. The existing unknown-route test verifies HTTP 404 for that path. No provisional deduplication API is exposed.
- Separate server startup, routing, the storage interface, and the SQLite adapter under `src/server/`.
- Add `GET /v1/findings` and `POST /v1/bulk/findings`. Each logs only its route and returns HTTP 501 with `{"error":"not_implemented"}`; request bodies are not processed or logged.
- Reuse the bundled workbench's database initialization and migrations. No second SQLite schema, storage dependency, or unused CRUD implementation is added.
- Add the packaged `start:server` script and a `findings-service` Docker target. Docker invokes the same compiled SDK entrypoint directly so stop signals reach the server.
- Add Compose configuration, non-secret environment defaults, a persistent SQLite volume, documentation, focused tests, and container smoke coverage.
- Keep the existing default scanner image and bulk-scan Compose behavior unchanged.

## Testing

Verified at `620586e4`:

- Focused findings-server suite: 4 passed, including HTTP 404 for the unsupported bulk-dedupe path.
- Full SDK suites: 1,895 passed, 28 skipped, 0 failed in each run (seeds `12345` and `1566945902`).
- `pnpm run types`, `pnpm run format`, changed workflow formatting, and `git diff --check`: passed.
- CI passed, including Linux, macOS, Windows, both container targets, the updated two-route service smoke, and package/installed-consumer validation.
- Existing server runtime files remain packaged; no provisional dedupe module, new dependency, or release gate is introduced. Model calls in checks use synthetic fixtures.

## Risk and rollout

- This is plumbing only. Finding insertion, embeddings, pagination, and deduplication are not implemented; HTTP 501 must not be treated as successful processing.
- No existing CLI commands, arguments, flags, or defaults change. The new SDK package script is `pnpm run start:server`; server-specific environment settings are `HOST` (local default `127.0.0.1`) and `PORT` (default `3000`). Existing state-directory and Python settings are reused.
- Docker binds inside the container on `0.0.0.0:3000`; Compose publishes only on the host loopback interface. The preview has no API authentication and should not be exposed to an untrusted network without an authenticated proxy.
- Startup applies the existing workbench migrations before listening. The named volume preserves SQLite state; deleting that volume deletes the stored data.
- API behavior and the deduplication workflow remain separate follow-up changes.

## Public disclosure review


.github/workflows/container-ci.yml
```diff
@@ -99,6 +101,61 @@ jobs:
           docker run --rm codex-security:ci bulk-scan --help
           docker run --rm codex-security:ci info --json
 
+      - name: Verify findings service and persistent SQLite storage
```
kmbroai, line 106: simple test. this is extracted to a script in the next PR

soyeon-oai (approved):

PR #667, opened 2026-08-27
feat: publish custom findings and persist dedupe groups

kmbroai:
## Summary

Stacked on #661. Complete the local scan-to-findings-service flow: publish completed findings from the CLI, review candidates locally, and persist the accepted duplicate relationships in SQLite.

## Changes

- Merge the updated #661 base, carrying group storage and migration 35 into `plugins/codex-security/scripts/` and inheriting the canonical generated-bundle workflow. Preserve the parent's two-stage, transitive deduplication behavior and update the write-back fixture to remove its obsolete whole-group reviewer.
- Inherit the earlier removal of the unused provisional API. Drop obsolete transition wording and a redundant smoke assertion; production code, unit tests, and the supported local workflow are unchanged.
- Translate the mounted SQLite smoke fixture directly to TypeScript using Node built-ins, with the existing SQL and assertions retained at this stack layer. Include it explicitly in TypeScript checking; no runtime dependency or public contract changes.
- Add `publish scan --to custom --findings-url http://localhost:3000` and the SDK's `publishScanToCustom`. The URL is an explicit service base URL with no default; publication posts complete sealed findings and their repository ID to `/v1/bulk/findings`. Existing scan selection and `--dry-run` work without changing Linear or Cloud publication.
- Add an append-only migration for `finding_dedupe_groups` and `finding_dedupe_group_members`. Findings can belong to multiple overlapping reviewed groups. Membership identifies a group idempotently; overlapping groups are not automatically united.
- Add `POST /v1/dedupe-groups` for atomic group batches and `GET /v1/finding/{id}/dedupe-groups` for readback. Missing members reject the entire batch.
- Have `dedupe` persist accepted groups after all reviews and fail if write-back fails. Preserve its existing result shape, required review fields, original findings, embeddings, and sealed scan artifacts.
- Document the contracts and extend CLI, SDK, SQLite, installed-package, and Docker coverage. Update the npm runtime-module allowlist; test and Docker fixtures remain excluded.

### Canonical plugin source update (`3de1f167`)

- `build:plugin` and `check:plugin-source` passed on this branch: 119 generated plugin files, including `workbench_findings.py`, and no tracked files under `sdk/typescript/_bundled_plugin`.
- Focused service, embedding, plugin-build, deduplication, transport, CLI, and custom-publication tests: 48 passed.
- Both full SDK suites passed: seeds `12345` and `4011281236`, each with 1,978 passed, 29 skipped, and zero failures.
[...]

PR #673, opened 2026-08-27
feat: persist resumable local findings workflows

kmbroai:
## Summary

Make the local scan → custom publication → dedupe workflow resumable between completed steps, using the existing workbench SQLite database.

Stacked on #667, with individual review checkpoints in #674. This update includes the current parent, canonical plugin source layout, and screening/pair deduplication. Neither PR is merged into its base.

## Changes

- Add the approved optional `--workflow-id ID` to `scan`, `publish scan --to custom`, and `dedupe`, plus corresponding SDK `workflowId` options. Without it, existing behavior is unchanged. Publication and dedupe can resolve the saved scan from the workflow instead of requiring `--scan`.
- Record scan ID and artifact directory atomically during registration. Recover completed scans even when their workflow receipt was not saved; retry unfinished stages without repeating acknowledged work.
- Persist immutable scan/artifact identity, destination and scope, stage status, receipts, errors, and exact dedupe results, including successful empty results. Workflow dedupe completes missing publication first.
- Use explicit SQLite columns for metadata and retain JSON only for receipts and result/write payloads. Existing append-only migrations preserve records, timestamps, references, and atomic updates.
- Keep workflow metadata outside sealed artifacts and retain existing output shapes and source authorization.
- Move the helper into the canonical plugin source, retain the runtime package contract, and support the existing helper `--help` check. Align canonical Python schema tests with this branch's existing migrations; SQL is unchanged.
- This integration adds no public CLI syntax beyond the previously approved workflow option.

## Testing

Current integration at `f6b60100`:

- Focused integration tests: 230 passed, zero failures.
- Full SDK suites passed with fixed seed `12345` and randomized seed `44458130`: each had 1,993 passed, 29 skipped, and zero failures.
- Canonical plugin generation/source ownership, generated-model consistency, TypeScript checks including MCP, formatting, and `git diff --check` passed.
- Canonical Python migration/setup, scan indexing, coordinator, and helper-help checks: 64 passed. Verified that all existing SQL migration tuples and helper functions are unchanged by this integration.
- Docker smoke passed from a tracked source snapshot with no pre-generated host plugin: host publication, embeddings/SQLite storage, workflow resume, dedupe/group writes, repeated invocation without additional reviews, restart persistence, and clean shutdown.
- The Docker-built npm archive passed package and installed-consumer checks: 364 entries, including 120 plugin files. Verified public imports, NodeNext types, CLI, credential locking, MCP initialization, bundled Codex, and nested-worker startup without global Codex. Test fixtures are excluded.
- Scoped workflow lint passed.
- Hosted CI is tracked by the checks attached to the current PR commit; these results describe local verification.

## Risk and rollout

Opt-in behavior with append-only SQLite migrations. The column migration retains existing IDs and payloads without rebuilding the table or disabling foreign keys. Use updated workflow/checkpoint code after upgrading the schema; old `state_json` readers are incompatible with the migrated table.

Use one coordinating process per workflow. Reusing an ID with a different scan, destination, or bound repository scope fails; use another ID for another workflow. Unfinished scans retain existing output-directory and archive safeguards. This PR resumes between stages, not inside an unfinished scan or dedupe review. Lost publication acknowledgements retry the existing idempotent upsert. #674 checkpoints individual reviews and pending group writes.

Local verification uses Linux and synthetic provider responses. Native Windows execution and live model reviews were not run locally. No release, image publication, or PR merge is performed by this update.

## Public disclosure review


sdk/typescript/src/api.ts
```diff
@@ -433,8 +446,113 @@ export class CodexSecurity {
...
+    );
+    const workflow = new FindingWorkflow(
+      workflowId,
+      this.#dependencies.environment,
+      this.#dependencies.runWorkbench,
+      this.config.pythonPath,
     );
+    if (options.outputDir !== undefined)
+      await workflow.protectArtifacts(options.outputDir);
+    const state = await workflow.bind({
```
soyeon-oai, line 478: **[P1] Bind resumed scans to the scanned source**

This binding covers only the repository path and scan options. If the checkout at that path changes after the scan completes, reusing the workflow ID returns the previous sealed findings below, while a later dedupe snapshots and reviews the new checkout. That can publish findings or persist duplicate groups against the wrong source. Please bind the workflow to the scan target revision/snapshot digest and reject reuse when the current source does not match, requiring a new workflow ID.

soyeon-oai (approved):

sdk/typescript/src/custom-publish.ts
```diff
@@ -34,31 +41,55 @@ export async function publishScanToCustom(
...
   };
+  if (options.workflowId === undefined || options.dryRun)
+    return await publish();
+  const workflow = new FindingWorkflow(
+    options.workflowId,
+    dependencies.environment,
+    dependencies.runWorkbench,
+  );
+  await workflow.protectArtifacts(canonicalDirectory);
+  await workflow.bind({
```
zcrab-oai, line 87: **[P1] Bind the scanned repository when publication initializes a workflow**

Publication can create a workflow here with `scanId`, `scanDir`, and `artifactDigest` but no `repositoryPath`, then mark its scan stage completed. A subsequent `run(repositoryB, { workflowId })` fills in the missing path with repository B and returns repository A's already-completed contract without checking that the saved scan belongs to B. This differs from the existing same-path/source-revision review comment: it returns findings from an entirely different requested repository. Bind and validate the original scanned repository before allowing publication-created workflows to be resumed.

sdk/typescript/src/api.ts
```diff
@@ -433,8 +446,113 @@ export class CodexSecurity {
...
+      let completed = state.stages.scan.status === "completed";
+      if (!completed) {
+        const scan = await workflow.registeredScan(state.scanId);
+        completed =
+          (scan["progress"] as JsonObject | undefined)?.["status"] ===
+          "complete";
+        if (completed)
+          metadata = {
+            threadId: (scan["continuationThreadId"] as string) ?? "",
+            turnResult: { status: "completed" },
```
zcrab-oai, line 513: **[P2] Recover the complete scan result when the workflow receipt is missing**

If scan registration is complete but the workflow completion receipt was lost, recovery reconstructs only `threadId` and a completed turn status, then persists that incomplete result. Model usage/cost, SARIF metadata, and `repositoryFindings` are consequently lost permanently even though the scan itself completed successfully. The receipt-loss regression case does not compare the recovered `ScanResult`. Persist enough result metadata atomically with scan completion, or reconstruct all fields before writing the workflow receipt, and cover receipt-loss recovery in the result assertions.

zcrab-oai (commented):
Two validated findings introduced in this PR: cross-repository workflow reuse and incomplete scan recovery metadata.

sdk/typescript/src/api.ts
kmbroai: That is actually what we want to happen. It's not a bug.

PR #676, opened 2026-08-27
fix(sdk): prefer exact POSIX PATH for trusted executables

Hughhhhcoder:
## Summary

On POSIX systems, `resolveTrustedExecutable` can select a case-variant `path` environment entry instead of the exact `PATH` entry. That can cause trusted executable resolution to inspect the wrong search path.

Reproduction:

- Provide both `path=/tmp/decoy` and `PATH=/tmp/trusted` in the environment.
- Resolve an executable that exists only under `/tmp/trusted`.

Expected: the exact POSIX `PATH` value is used.

Actual: the first case-insensitive match can select `path` instead.

## Changes

- Prefer the exact `PATH` key before the existing case-insensitive fallback.
- Add a POSIX regression test that uses distinct decoy and trusted directories.

## Testing

- `bun test --timeout 30000 -t 'prefers the exact POSIX PATH' tests-ts/trusted-executable.test.ts` (1 passed, 9 skipped)
- `git diff --check`

## Risk and rollout

This is a narrow POSIX environment-selection fix. It does not change the public API or the Windows fallback behavior. The exact `PATH` value is now preferred when present; case-insensitive fallback remains available for environments that do not provide it.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/trusted-executable.ts
+++ b/sdk/typescript/src/trusted-executable.ts
@@ -23,9 +23,11 @@ export async function resolveTrustedExecutable(
   const root = await realpath(protectedRoot).catch(() =>
     resolve(protectedRoot),
   );
-  const path = Object.entries(environment).find(
-    ([name]) => name.toUpperCase() === "PATH",
-  )?.[1];
+  const path =
+    environment["PATH"] ??
+    Object.entries(environment).find(
+      ([name]) => name.toUpperCase() === "PATH",
+    )?.[1];
   const entries: string[] = [];
   for (let entry of path?.split(delimiter) ?? []) {
     if (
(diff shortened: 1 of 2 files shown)


Hughhhhcoder:
Focused regression tests and repository formatting checks pass on the current head (`1dc921d`). @codex review

mldangelo-oai (approved):
Thanks for the focused fix and regression test! Preferring the exact PATH key is the right behavior here, and this keeps the change nicely scoped.

I verified the before/after behavior and ran the relevant executable-resolution, target, and GitHub tests: 48 passed, with four platform-specific skips. Typechecking and formatting also passed.

PR #677, opened 2026-08-27
fix(sdk): preserve caller cancellation in cloud publication

Hughhhhcoder:
## Summary

Caller cancellation during cloud publication is currently caught and rethrown as a generic `CodexSecurityError`. This loses the caller's cancellation identity and can make cancellation look like a publication failure.

Reproduction:

- Start a cloud publication with an `AbortSignal`.
- Abort the signal while the publication request is in flight, or while the acceptance receipt is being decoded.

Expected: the original cancellation error is propagated unchanged.

Actual: the generic cloud-publication error wrapper is thrown instead.

## Changes

- Re-throw an aborted caller signal before wrapping publication request failures.
- Re-throw an aborted caller signal before wrapping acceptance receipt decoding failures.
- Add regression tests for both cancellation points.

## Testing

- `bun test --timeout 30000 ./tests-ts/cloud-publish.test.ts` (35 passed)
- `pnpm run types`
- `pnpm run format`
- `pnpm run test:ci` (1,917 passed, 30 platform-specific skips, 0 failed)
- `pnpm pack --pack-destination ../../dist`
- `pnpm run check:package /tmp/codex-security-verify/dist/openai-codex-security-0.1.20.tgz`
- `pnpm run test:package`
- `git diff --check`

The package archive inspection and installed-package smoke test validated the public import, NodeNext types, CLI, credential locking, bundled plugin files, bundled Codex version, and a nested worker without a global `codex` executable.

## Risk and rollout

This preserves the existing cancellation contract without changing the public API or successful publication behavior. Non-cancellation failures continue through the existing `CodexSecurityError` handling.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/cloud-publish.ts
+++ b/sdk/typescript/src/cloud-publish.ts
@@ -455,6 +455,7 @@ async function publishCloudPayload(
       signal,
     });
   } catch {
+    dependencies.signal?.throwIfAborted();
     // A lost response does not establish whether the server accepted the POST.
     throw new CodexSecurityError(
       "Cloud publication was not confirmed. The request was not retried; check whether it was accepted before submitting again.",
@@ -475,7 +476,10 @@ async function publishCloudPayload(
     );
   }
   const receipt = receiptSchema.safeParse(
-    await response.json().catch(() => undefined),
+    await response.json().catch(() => {
+      dependencies.signal?.throwIfAborted();
+      return undefined;
+    }),
   );
   // Cloud assigns opaque IDs in request order, so they cannot be compared to
   // local finding IDs. The authenticated response must still preserve the
--- a/sdk/typescript/tests-ts/cloud-publish.test.ts
+++ b/sdk/typescript/tests-ts/cloud-publish.test.ts
@@ -684,6 +684,58 @@ describe("Cloud publication", () => {
     expect(requests).toBe(1);
   });
 
+  test("preserves caller cancellation during the publication request", async () => {
+    const { scan, environment } = await fixture();
+    const controller = new AbortController();
+    const cancellation = new Error("publication cancelled");
+
+    await expect(
+      publishScanToCloud(scan, {
+        environment,
+        signal: controller.signal,
+        fetch: async (_url, options) => {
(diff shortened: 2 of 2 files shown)


Hughhhhcoder:
Focused regression tests and repository formatting checks pass on the current head (`9fb4482`). @codex review

mldangelo-oai (approved):
Thanks for keeping this small and covering both cancellation points! I ran the Cloud publisher and CLI suites: all 66 tests passed, along with typechecking.

The added cancellation test needs a small Prettier cleanup before CI can pass: run `pnpm exec prettier --write tests-ts/cloud-publish.test.ts` from `sdk/typescript`. The required adjustment is format-only.

Approving the cancellation-identity change. One follow-up worth retaining: an interrupted upload may already have been accepted, so the CLI should keep the check-before-retrying guidance and a useful batch error instead of only SIGINT/SIGTERM. The existing CLI tests mock the publisher, so a test connecting the real publisher to the CLI would help cover that interaction.

mldangelo-oai:
Thanks again for the focused cancellation fix! I carried the change into #680 with the original author attribution, the test-formatting cleanup, and coverage connecting the real publisher to the CLI. That follow-up preserves the caller cancellation reason while keeping the check-before-retrying guidance and useful batch failure output.

Hughhhhcoder:
Thank you for carrying this into #680 and preserving the original author attribution. I confirmed that #680 includes the cancellation propagation plus the CLI-level recovery guidance and integration coverage you described.

I also pushed the requested Prettier-only cleanup to this branch. On the current head, the focused Cloud publication file passes (35 tests), the full CI suite passes (1,917 passed, 30 platform-specific skips, 0 failures), and the packed/installed-package checks pass. Since #680 supersedes this branch, I will avoid duplicating its follow-up changes and leave closure to the maintainers.

mldangelo-oai (approved):
Thanks for the formatting follow-up and the additional validation! I verified that this head is identical to the locally validated formatting fix, with no production changes beyond the original cancellation patch. Approving this exact head. The CLI recovery-guidance follow-up is tracked in #680, including cancellation before and during an upload.

PR #720, opened 2026-08-28
feat(sdk): support renewable model authentication

gero-oai:
## Summary

Allow SDK callers to configure the embeddings endpoint and supply renewable credentials. Preserve an explicitly selected command-auth provider in `CODEX_HOME` when preparing reviews.

## Changes

- Add embedder options for the API base URL, headers, and a static key or per-batch credential callback. Preserve the original constructor, OpenAI defaults, batching, and vector validation.
- Export the embedder, findings server, SQLite store, and core types through `@openai/codex-security/server` without starting a server on import.
- Preserve selected command authentication, including profile selection, without replacing it with managed login or static OpenAI keys. Keep other authentication behavior and review isolation unchanged.
- Select the first nonempty `OPENAI_API_KEY` or `CODEX_API_KEY`, including Windows environment-name casing.
- Start the review host in its existing private temporary directory while keeping the review thread in the source checkout. Resolve relative `CODEX_HOME` paths before spawning, so checkout files cannot supply an implicitly relative auth helper.
- For CLI comparisons, default command authentication to the supplied home via native `auth.cwd`. Preserve explicit auth directories and provider overrides, keep the thread in its original directory, and leave the home configuration unchanged.
- Extend existing deterministic and installed-package tests. No new provider smoke framework; README changes are limited to one sentence.

## Testing

Validated with Bun 1.3.14 and Node 22:

- Focused comparison, component planning, and review tests: **73 passed, 7 skipped, 0 failed**. Covers default and explicit auth directories, relative homes, provider overrides, and unchanged thread restrictions.
- Full shared SDK suite through all three existing CI shards, seed `12345`: **2,138 passed, 42 skipped, 0 failed**.
- `pnpm run types`, `pnpm run format`, `pnpm run build`, and `git diff --check`: passed.
- `pnpm pack` and `pnpm run check:package`: passed, including the server import, strict NodeNext consumer, installed CLI, credential locking, bundled runtime, and nested-worker checks.

Direct local verification with the pinned Codex 0.149.1 and a localhost provider confirmed that the checkout-relative helper was not executed, failures remained safe, and a trusted absolute helper completed both review models with a relative `CODEX_HOME`. A further native CLI comparison check confirmed that relative helpers resolve from the supplied home, including a provider ID containing a dot. When that helper was missing, the comparison failed without executing the checkout helper or using the competing ambient key. Native Codex attempted unauthenticated requests, which the synthetic provider rejected. No new provider test framework is included. No live provider credentials were used; this does not establish compatibility with a live service.

## Risk and rollout

The SDK options and server subpath are additive. No new CLI surface, dependencies, models, or database migrations. Embedding renewal failures retain the existing safe error boundary without ambient fallback. Review threads retain the same target and isolation settings, but host-side relative helper paths no longer resolve from that checkout. Explicit caller configuration remains authoritative. CLI comparisons now resolve implicit auth commands and relative arguments from the supplied home rather than the caller’s working directory; callers needing another directory can use native `auth.cwd`.

Callers own token acquisition and private-home lifecycle; Codex owns review-token renewal. Release through the normal package process after review and CI, and validate the caller integration against an authorized provider before rollout.

## Public disclosure review

### DIFF
--- a/docker/fixtures/mock-embeddings.mjs
+++ b/docker/fixtures/mock-embeddings.mjs
@@ -2,7 +2,10 @@ import assert from "node:assert/strict";
 
 globalThis.fetch = async (url, init) => {
   assert.equal(url, "https://api.openai.com/v1/embeddings");
-  assert.equal(init.headers.Authorization, "Bearer synthetic-container-key");
+  assert.equal(
+    new Headers(init.headers).get("Authorization"),
+    "Bearer synthetic-container-key",
+  );
   const { input, model, dimensions } = JSON.parse(init.body);
   assert.equal(model, "text-embedding-3-large");
   assert.equal(dimensions, 1536);
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -1357,6 +1357,8 @@ their vectors are combined by token weight and normalized. Requests respect
 the provider's 8,192-token input, 300,000-token request, and 2,048-input limits.
 See the [embedding API contract](https://developers.openai.com/api/reference/resources/embeddings/methods/create).
 
+`@openai/codex-security/server` exposes the findings server and an embedder with configurable `baseUrl`, headers, and an API-key callback resolved before each batch.
+
 Storage initializes before the server listens. The SQLite adapter reuses the
 bundled workbench's schema and migrations at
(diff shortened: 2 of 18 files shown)


gero-oai:
Superseeded by https://github.com/openai/codex-security/pull/779 and others

PR #733, opened 2026-08-29
test: separate workflow orchestration from database rules

mldangelo-oai:
## Summary

Workflow orchestration and index-query tests repeatedly launched Python and rebuilt database state for each permutation. Separate those rules from the process and storage boundaries so the tests exercise the owning implementation with less setup.

## Changes

- Record fake workbench commands and inputs before parsing, then verify them after error handling instead of keeping a separate call counter. A swallowed persistence error can no longer hide a mismatched request.
- Test TypeScript retry, resume, checkpoint invalidation, dry-run, and error propagation through small, strict workbench fakes using the existing injection points.
- Test Python workflow transitions, bindings, rollback, and review ownership directly against fresh real SQLite databases. Build an empty schema once and copy it into an isolated in-memory connection for each case, with foreign keys enabled.
- Move four native-index filter and pagination cases to direct production functions with real SQLite query fixtures. Keep scan completion and index registration in the existing integration tests.
- Retain real-process workflow termination/resume, publication replay across a database migration, old checkpoint upgrades, source snapshots, and sealed-artifact path protection. Keep the column migration on a real file that is closed and reopened.
- Document the division between orchestration, database rules, and boundary tests.

## Testing

- Follow-up full Bun seed 12345: 2,059 passed, 36 skipped, no failures. The final focused workflow and publication tests passed 19 cases.
- A deliberate request mismatch passed unnoticed before the fake fix and is rejected afterward. The retained regression failed before the fix and now passes; missing, extra, reordered, and invalid-command requests were also checked. The invalid-command regression also failed before its fix and passes afterward.
- Full Bun suite with seeds 12345 and 67890: each passed 2,056 tests with 36 platform/live-integration skips and no failures.
- Full Python suite plus source-compatibility tests: 1,066 passed, 5 skipped, and 104 subtests passed.
- `pnpm run types`, `pnpm run format`, Ruff checks, Ruff formatting, and `git diff --check` passed.
- Local focused workflow Bun run: 65.1 seconds before, 30.7 seconds after, including retained integration cases. The direct Python workflow rules also passed.
- Four native-index query cases: 13.1 seconds before and 0.28 seconds after, including pytest startup. These are group-level measurements on an uncontrolled local machine, not a whole-CI speedup claim.

## Risk and rollout

No production code changes. The strict fakes cover TypeScript decisions; real Python functions own the SQL assertions. File-backed migration, crash/restart, concurrency, signals, permissions, symlinks, and installed-package coverage remain enabled in the integration suites. Each in-memory connection and temporary directory belongs to one test. Hosted Windows and Linux validation and the current-head review remain required.

## Public disclosure review


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

kmbroai (approved):

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

PR #742, opened 2026-08-29
feat: share scan settings across the CLI and SDK

mldangelo-oai:
## Summary

Reusing scan settings currently requires repeating CLI options or translating them into SDK calls. This adds explicitly selected YAML and JSON configuration shared by the CLI and SDK, with one input schema, validation path, and set of Deep Scan defaults.

## Changes

- Add `-c FILE` / `--config FILE` to `scan`, `bulk-scan`, `scan-components`, and `info`. `CODEX_SECURITY_PROJECT_CONFIG` is an optional fallback; an explicit flag wins. Files are never discovered automatically, and SDK runs and saved reruns do not read this variable.
- Add `init [file]`, defaulting to `codex-security.yaml`, with YAML and JSON support and no overwrites. Ship an editor schema and expose `loadProjectConfig()` and `resolveProjectConfig()` for SDK callers. `info --config FILE --json` inspects resolved settings and their sources without launching a scan.
- Keep repository selection explicit. Scope paths are relative to each repository; prompt, context, and output paths are relative to the configuration file. CLI overrides replace configured lists and scope variants. Bulk rows and component plans keep their existing target precedence.
- Share all six Deep Scan defaults with the plugin, preserve unrelated legacy TOML settings, and save effective settings and authentication in launch recipes. Reruns accept replacement instructions through `--scan-prompt-file`; interrupted Deep Scans and bulk recovery retain their saved instructions and settings.
- Apply configured native overrides, prompts, per-attempt cost limits, and severity policy to single, bulk, and component scans. Preserve policy outcomes when completed bulk work is reused or interrupted work is recovered.
- Merge `main` at `5851c8e41ee3b9920c6b3ddb16fb39c125ce3308`, retaining its policy generation, recovery, authentication, native package, and release-title changes. Reuse shared option helpers in recovery and keep workflow preparation from reading prompt files twice. Verify configuration isolation between concurrent scans and forwarding through fresh and resumed discovery and reducer workers.

## Testing

- Focused configuration, prompt, SDK, CLI, and recovery suites: 217 passed, 1 platform skip. The additional recovery-policy cases passed after correcting their interrupted-directory fixture.
- Complete randomized SDK run (`--seed=537208730`): 2,956 passed, 57 skipped, 2 failed. The import error-message regression was fixed afterward and all 22 import tests passed; the unchanged pagination test timed out once, then passed in isolation (and also passed in the fixed-seed run).
- The complete fixed-seed run (`--seed=12345`) recorded 2,951 passed, 59 skipped, and 6 failures. Its recovery-fixture and import-message failures were corrected and retested; its two environment-dependent publication checks passed with isolated state and process inspection enabled. Both complete runs began before the final corrections, so neither is claimed as a clean run of the final commit.
- After integrating the latest release-title change from `main`, both affected test files passed: 360 tests, 0 failures.
- Release build and installed package validation passed: public imports, NodeNext types, configuration CLI flows, SDK lifecycle, credential locking, all 139 bundled plugin files, MCP initialization, and a nested Codex worker without a global installation.
- SDK types and generated-file checks, formatting, portable source compilation, plugin compatibility checks and 9 checker tests, Ruff lint/format, worker-launch tests, and 61 focused Python tests passed. `git diff --check` passed.

## Risk and rollout

Project files are trusted operator configuration with the same authority as CLI options and native Codex overrides. They can select model-service destinations and configured MCP processes. Keep scanner configuration outside untrusted checkouts. Existing credential, output-path, and scan-integrity protections remain in place.

The unreleased project-file format uses `snake_case`; existing SDK options keep camelCase and existing CLI flags keep their spelling. The new command, flags, and environment variable are optional. Files cannot select a repository, enable automatic patching or publication, or choose a plugin or Python installation. Bulk snapshots support repository and path scopes; unsupported diff scopes fail before work starts.

Deep-only options require active Deep Scan mode, and custom validation remains limited to Standard scans. Cost limits apply per attempt. Configured severity thresholds return exit code 1, with operational failures taking precedence. Saved reruns use current source and context files and require replacement instructions when the original scan used nonblank instructions.

Local validation uses synthetic fixtures on macOS arm64. Other platforms and Node versions require CI. No live inference, release, or deployment was performed.

## Public disclosure review

Historical commit contact metadata and automated review links prevent the second attestation. This update preserves existing history and uses a GitHub noreply author and committer identity.


mldangelo-oai:
@codex review

Please review the current head, `e934a27a04ca703e856022956bb489f034fafa76`.

docs/project-configuration.md
```diff
@@ -0,0 +1,178 @@
...
+Use `scan [repository] -c FILE` or `scan [repository] --config FILE` to load one
+YAML or JSON file:
+
+```sh
+codex-security scan . -c codex-security.yaml --dry-run --json
+codex-security scan . -c codex-security.json --model gpt-5.6-terra
+```
+
+The supported extensions are `.yaml`, `.yml`, and `.json`. Without `-c`, no file
+is loaded, even if `codex-security.yaml` exists. Other commands and SDK `run()`
```
kmbroai: Should we actually load it by default?

mldangelo-oai:
@codex review

Please review the current head, `0ea24644ce1d3d9dbdc5fa6088eb72654bd956a1`.

mldangelo-oai:
@codex review

Please review the current head, `1e33329e85ae3d58c0d30beb8bcbbf23cadd29b5`.

mldangelo-oai:
@codex review

Please review the current head, `3219d7f782e587f03e2e798f9e4dc8a9a04ed016`.

mldangelo-oai:
@codex review

Please review the current head, `52da4db7d0c9e1ad96d74ab1ad88e46dad0aef8f`.

mldangelo-oai:
@codex review

Please review the current head, `d1a465e43a81eb14b1fdeca98a41f5b48a0bf316`.

mldangelo-oai:
@codex review

Please review the current head, `4633f70c1855659ed919a6803a8d35cd37de0462`.

mldangelo-oai:
@codex review

Please review the current head, `6eb26ae6f58ef7ed591ef22d3ce8b563f9c32467`.

mldangelo-oai:
@codex review

Please review the current head, `99aec8696668fc11cc503138b2a83e122acc6e32`.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,274 @@
...
+    );
+  }
+  let value: unknown;
+  try {
+    if (extension === ".json") {
+      value = JSON.parse(text);
+    } else {
+      const document = parseDocument(text, { prettyErrors: false });
+      if (document.errors.length > 0) throw document.errors[0];
+      value = document.toJS({ maxAliasCount: -1 });
```
mldangelo: **`maxAliasCount: -1` disables the billion-laughs guard.** This came in with `99aec869`, and `-1` removes the limit rather than raising it.

Verified through this exact code path:

```
source bytes: 439
DEFAULT (cap 100) -> rejected: "Excessive alias count indicates a resource exhaustion attack"
THIS CODE (-1)    -> expanded until JSON.stringify threw RangeError: Invalid string length
```

And through `readProjectConfig` itself, a 6-level bomb (531k nodes) **resolved successfully in 79 ms** — because expansion happens here, at line 100, *before* `requireProjectConfig` at line 106, and `codex` is `.catchall(z.json())`, so the expanded blob then passes validation and flows on into `mergeCodexOverrides` -> `cloneJson` -> `deepMerge`. Strict schema offers no protection.

The test that motivated this uses 150 aliases; the library default is 100. Calibrating:

| config | 100 | 1000 | 10000 |
|---|---|---|---|
| the 150-alias test below | rejected | ok | ok |
| 500 aliases of a 20-key table | rejected | ok | ok |
| alias bomb | rejected | rejected | rejected |

```suggestion
      value = document.toJS({ maxAliasCount: 10_000 });
```

Severity note so this is not over-read: `-c` is explicit and the file's `codex` block still passes the full `validateOverrides` chain, so this grants nothing beyond `--codex`. The reason to fix it is that this format's whole value is being inert data, and this is the one line that quietly breaks that.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+}
+
+export function resolveScanSettings(
+  project: ProjectConfigSource | undefined,
+  overrides: Partial<ScanSettings> & { codexOverrides?: JsonObject },
+  directory: string,
+): ResolvedProjectConfig {
+  const file = project?.input;
+  const sources: Record<string, ConfigurationSource> = {
```
mldangelo: `sources` seeds four settings to `"default"` (`auth`, `scan.mode`, `scan.scope`, `scan.knowledge_base`) but `scan.instructions_file`, `scan.validation_file`, `output.directory`, `policy.fail_on_severity`, and `limits.max_cost_usd_per_scan` only appear once `choose()` picks them.

Since this map is a public `--dry-run --json` output, a consumer cannot distinguish "unset" from "not reported". Seed all nine or none.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,274 @@
...
+      sources[key] = "cli";
+      return explicit;
+    }
+    if (configured !== undefined) {
+      sources[key] = "project";
+      return configured;
+    }
+    return undefined;
+  };
+  const filePath = (value: string | undefined): string | undefined =>
```
mldangelo: `resolve(project!.directory, ...)` — the non-null assertion is safe only because every argument passed to `filePath` is derived from `file`, which is `project?.input`. Nothing enforces that coupling; a future `filePath(overrides.something)` would throw at runtime with no type error.

Narrowing on `project` once at the top, or having `filePath` close over a resolved `directory` captured only when `project !== undefined`, removes the assertion.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,274 @@
...
+  const cliPath = (value: string | undefined): string | undefined =>
+    value === undefined ? undefined : resolve(directory, expandHome(value));
+
+  const mode =
+    choose("scan.mode", file?.scan?.mode, overrides.mode) ?? DEFAULT_SCAN_MODE;
+  if (
+    mode !== "deep" &&
+    DEEP_SCAN_SETTINGS.some(([name]) => overrides[name] !== undefined)
+  ) {
+    throw new ConfigurationError("Deep scan settings require --mode deep.");
```
mldangelo: This message uses CLI flag syntax, but `resolveScanSettings` is reachable from the public `resolveProjectConfig()` / `loadProjectConfig()` SDK entry points, where `--mode deep` means nothing.

There is also a second copy of this same rule with different wording at `deep-config.ts:38` ("Deep scan settings require deep mode."), and `api.ts:2343` calls `deepScanOptions(options)` purely for its throw. One rule, three sites, two strings — worth collapsing to one that phrases itself from the input layer.

mldangelo (commented):
## Summary

I read this end to end, ran the suite, and probed several behaviours directly rather than reasoning from the diff. The design is strong: one `ScanSettings` source of truth shared by CLI/SDK/file, `snake_case` matching native Codex, explicit `-c` with no discovery, no executable config, no interpolation, no remote include, no multi-file merge, strict schema, hard failure on invalid input, and unusually thorough docs. The `sources` provenance in `--dry-run` is a genuinely good idea. Several of these choices are worth defending explicitly, because the obvious comparables in this space get them wrong — a permissive "warn and continue" loader would turn a typo'd `fail_on_severity` into a silently report-only gate, and auto-discovery would mean a scanned repository could configure its own scanner.

Two findings I would resolve before merge, one free fix, and a set of architectural notes that get more expensive the longer this surface grows.

### Verification

- `pnpm run types`: pass. `prettier --check`: pass.
- `pnpm test`: two full runs, different seeds, both `2198 pass / 45 skip / 22 fail`. All 22 failures live in `test-shards.test.ts` (16), `codex-review.test.ts` (4), `skeleton.test.ts` (1), `release-automation.test.ts` (1). I ran all four against `origin/main` in a clean worktree and they fail identically there — local environment (bun 1.3.3 `Bun.YAML.parse`, absent API keys), not this branch. **No regressions introduced.**

---

## 1. Blocking — the workflow digest is no longer stable across releases

`api.ts:471-475` spreads the fully resolved deep settings into `options` before `workflowDigest` consumes it at `:486`. `local.deepScanConfiguration.settings` is `Required<DeepScanOptions>` — all six values, resolved from ambient TOML plus shipped defaults — and `workflowDigest` is `JSON.stringify` + sha256, so those numbers enter the hash.

`workbench_finding_workflows.py:90` treats a changed binding as fatal:

```python
if field in state and state[field] != value:
    raise SystemExit(f"Workflow {state['id']} is already bound to a different {field}. Use another --workflow-id.")
```

Measured with the real function:

```
pre-742 (identical command) : 1279abaf39c95e87
post-742                    : 95941751b4d0133c DIFFERS
after a default bump        : d64f24a4ab2c6b35 DIFFERS
after an ambient TOML edit  : 7e5e9486b43e733e DIFFERS
```

So bumping any deep default in a release, or a user editing `~/.codex/codex-security/config.toml`, makes every in-flight deep workflow unresumable — with an error telling them to abandon the ID. The same command no longer yields the same digest.

This is the mirror image of the care taken for recipes ("Complete saved settings do not depend on today's ambient configuration"). The digest should hash the user's *request*, not its resolution: hash `scanSettings(options)` before the spread, or exclude the resolved deep block explicitly.

## 2. Blocking — `rerun` becomes unreachable for any scan that used instructions

`cli.ts:4418` refuses the rerun outright, but the `rerun` command (`cli.ts:1805-1817`) exposes only `--validation-prompt-file` and `--verbose`. Confirmed against `origin/main`: `scanArgumentsFromRecipe` previously had no scan-prompt handling at all, so these reruns worked and silently dropped the instructions.

Refusing to omit silently is correct. But custom validation gets the recoverable treatment — blocked, then re-openable by supplying the file — and this does not. Adding `--scan-prompt-file` (and plausibly `-c`) to `rerun` makes the two symmetric and keeps the capability.

## 3. Free fix — the YAML alias guard is disabled

`project-config.ts:100`. Detail in the inline comment; the short version is that a 439-byte file expands past V8's maximum string length, expansion happens *before* schema validation, and `codex` is `.catchall(z.json())` so the expanded blob validates cleanly. I calibrated the replacement:

| config | cap 100 | 1000 | 10000 |
|---|---|---|---|
| this PR's own 150-alias test | rejected | ok | ok |
| 500 aliases of a 20-key table | rejected | ok | ok |
| alias bomb | rejected | rejected | rejected |

`maxAliasCount: 10_000` passes the test that motivated the change plus a config far fatter than anything plausible, and still rejects the bomb.

I want to be precise about severity: since `-c` is explicit and the file's `codex` block still flows through the full `validateOverrides` chain, this grants no more than `--codex` already does, so it is not a security blocker. The reason to fix it is that this format's entire value proposition is *inert data* — no exec, no network, no interpolation — and this is the one line that quietly breaks that promise.

---

## Architecture

**Provenance is a stringly-typed mutable map with six write sites across two modules.** `Record<string, ConfigurationSource>` keyed by hand-built dotted strings, written by: the four-key seed (`project-config.ts:138`), the `choose()` closure, `recordNativeSources` (recursive, with `delete`s), `resolveCliScope`'s separate map, `Object.assign(provenance.sources, scope.sources)`, and the dry-run patch that mutates `arguments_.projectConfig.sources`. No exhaustiveness checking, nothing linking a key to the setting it describes, and `scan.deep.${field}` built by concatenation with a hardcoded exception in two files. This is a public output contract assembled by mutation. A `ProvenanceKey` union plus a single builder that returns a frozen map would make it checkable and remove the ordering dependency between `resolveScanSettings` and `resolveCliScope`.

**Path-resolution invariants are enforced by convention, not types.** The genuinely hard thing here is that a path means different things per layer — config-dir-relative, cwd-relative, repo-relative. `ResolvedScanSettings` is exactly where that should be encoded, and it types `scanPromptFile?: string` identically to raw input. Two consequences: `resolveScanPrompts` re-resolves already-absolute paths (harmless only by accident), and `scanArgumentsFromRecipe` constructs the same `ResolvedScanSettings` type with *different* invariants — its `validationPromptFile` is raw CLI input. A branded `AbsolutePath` would make the subtlest part of this feature machine-checked, and would have caught the recipe asymmetry for free.

**One rule, three enforcement sites, two messages.** The mode/deep-settings constraint is enforced at `project-config.ts:172`, again at `deep-config.ts:38`, and `api.ts:2343` calls `deepScanOptions(options)` purely for its throw. One of those puts CLI flag syntax (`--mode deep`) into a path reachable from the public `resolveProjectConfig()`.

**Type-only import cycle; `ScanMode` has no canonical home.** `scan-settings.ts:4` imports `ScanTarget` from `targets.ts`; `targets.ts:10` imports `ScanMode` back and re-exports it. Erased at runtime, so a smell rather than a bug — but `deep-config.ts` now reaches `ScanMode` through `targets.js` while taking `DEEP_SCAN_SETTINGS` from `scan-settings.js`, i.e. two import paths to one module. Either move `ScanTarget` to a leaf types module or leave `ScanMode` in `targets.ts`.

**Four vocabularies, one hand-maintained map, one hardcoded exception.** TOML `stop_after_no_new` / project `stop_after_no_new` / SDK `stopAfterNoNew` / CLI `--stop-after-no-new`. `DEEP_SCAN_SETTINGS` carries two of four; the third is derived by `name === "subagents" ? "subagents_per_worker" : key` in two files. Errors currently pick one vocabulary regardless of which the user typed — a bad ambient TOML yields `Deep scan stopAfterNoNew must be a positive integer.` for a key the user wrote as `stop_after_no_new`, with no file named, while the adjacent "unknown key" and "must be a TOML table" errors *do* name the file. Putting all four spellings in the tuple and threading the source layer into validation fixes the whole class.

**`#validateLocalInputs` no longer only validates.** It reads prompt files and parses ambient TOML, and is called from four sites (`api.ts:466, 586, 683, 793`), twice per workflow scan. The second call avoids re-reading ambient TOML only because `#runWorkflow` spreads the resolved settings back into `options`, making `explicit` complete and short-circuiting `readDeepScanDocument`. That is a load-bearing side effect of a line that reads like a convenience — and it is the same line causing finding 1. Worth a comment at minimum, and ideally memoisation on the inputs struct instead.

---

## UX and DX

**`-c` landed on the wrong command first.** `config: optionValue("--config")` appears once, in `scan`. Not on `bulk-scan` (`cli.ts:3374`), not on `scan-components` (`:3125`). "Reusing scan settings means repeating command-line options" is a far bigger problem across many repositories than for a single scan — `bulk-scan` is the strongest case for this feature and currently cannot use it. Cheaper to add now while `resolveScanSettings` is fresh than to re-derive precedence per command later.

**There is no `init`.** This ships a new format, a generated schema, an npm export path for it, and two doc examples, with no way to produce a starting file. The `$schema` line users need is `./node_modules/@openai/codex-security/schemas/project-config.schema.json`, which nobody types from memory — and the two files in `docs/examples/` use a repo-relative path that is wrong to copy into a real project. `codex-security init` emitting a commented starter with the correct `$schema` and current defaults as comments would make the feature discoverable. The command list has no equivalent today.

**The repetition problem is half solved.** Six flags became one flag, still typed every invocation. Refusing discovery is right — this tool reads untrusted repositories, and auto-loading from the working directory would let a scanned repo configure its scanner. But an *operator-set* environment variable preserves that property exactly, since the operator controls the environment and repo content does not, and it removes the friction for both CI and local iteration. Note `CODEX_SECURITY_CONFIG_PATH` is already taken by the runtime config path passed to the plugin, so this needs a distinct name.

**No cheap way to ask "is my config right?"** Today that requires `scan . -c f.yaml --dry-run --json`, which needs a git repository and a positional argument and returns a large preflight payload with provenance nested inside. The `projectConfig` block is also emitted *only* when `-c` is passed, so "what settings will this scan use" is unanswerable for a plain invocation. The data already exists — this is an entry point and presentation problem. Either `-c` on `info`, or a `--print-config` that resolves, prints the settings table with sources, and stops. That is the command people will actually reach for, and it is what makes the provenance work in this PR pay off.

**Docs gap: the trust model is never stated.** The format is explicitly opt-in, which implies the file is trusted like the command line — but it is a repo-resident YAML file that reads like `.eslintrc`, and someone will wire `-c codex-security.yaml` into a workflow that runs against fork branches, at which point contributors control the scanner's model, provider, and native Codex config. That is inherent to the feature, not a defect in it. One sentence in `docs/project-configuration.md` saying the file is trusted like the command line, and not to point `-c` inside untrusted repository content, closes it.

---

## Simplification opportunities

- **`resolveCliScope` (`cli.ts:7360`) reimplements scope selection** that `projectScopeTarget` and `resolveScanSettings` also perform, and discards its own `scope` when nothing changed. Mutual exclusion of `--path/--diff/--working-tree` is additionally checked twice — once in the zod `.refine`, once here. One of these will drift.
- **`DEEP_SCAN_SETTINGS`'s third tuple element duplicates bounds the schema owns.** `["subagents", "subagents", 0]` exists only to choose between "a non-negative" and "a positive" in an error string, while `DeepScanSettingsSchema.shape.subagents` is the real validator. Derive the wording from the schema, or move the bound out of the schema — but not both.
- **`scanRecipe` now takes 11 positional parameters**, four optional and adjacent (`failOnSeverity`, `knowledgeBasePaths`, `maxCostUsd`, `deepScan`, `auth`). This PR added two. An options object is overdue and would make the next addition free.
- **`scanSettings()` is a filter that shares its name with the `ScanSettings` type.** `scanSettings(arguments_)` reads like construction; it is "pick the defined schema keys". `pickScanSettings` or `definedScanSettings` would read correctly at every call site.
- **`sources` seeds four of nine settings** (`project-config.ts:138`), so consumers cannot distinguish "unset" from "not reported". Seed all nine or none.

---

## Smaller notes

- `generate-deep-defaults.mjs` has `--check` wired into `pnpm run types`; `generate-project-config-schema.mjs` has none. Asymmetric, though `project-config.test.ts:151` covers the drift, so this is cosmetic.
- `skeleton.test.ts:360` was loosened from an exact build-command match to `not.toMatch(/\b(?:pnpm|npm|bun)\b/u)`, dropping the guarantee that `clean` runs first and that the dashboard build is included. Reasonable to avoid churn; worth being deliberate about what was given up.
- `smoke-package.mjs` exercising `loadProjectConfig`/`resolveProjectConfig` equivalence against the installed tarball is a good addition and worth keeping as the pattern for future public helpers.
- The ambient write-back at `deep-config.ts:110` materialises all six values into a user's config where the old code wrote only the overrides. Reachability is narrow — it needs `CODEX_SECURITY_STATE_DIR` or a link to make source and destination canonicalise equal — so this is a note rather than a finding, but the risk section mentions preserving unrelated TOML sections and not this.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,274 @@
...
+    choose(
+      "scan.scope",
+      projectScopeTarget(file?.scan?.scope),
+      overrides.target,
+    ) ?? "repository";
+  const configuredDeep = file?.scan?.deep;
+  const deep: DeepScanOptions = {};
+  if (mode === "deep") {
+    for (const [name, key] of DEEP_SCAN_SETTINGS) {
+      const field = name === "subagents" ? "subagents_per_worker" : key;
```
mldangelo: `name === "subagents" ? "subagents_per_worker" : key` is the one place the project-file spelling diverges from the TOML key, and it is hardcoded here and again at `cli.ts:6789`.

`DEEP_SCAN_SETTINGS` already carries a per-setting tuple. Adding the project-file key (and the CLI flag) as further elements removes both special cases and makes the four vocabularies — TOML `stop_after_no_new`, project `stop_after_no_new`, SDK `stopAfterNoNew`, CLI `--stop-after-no-new` — enumerable rather than implied.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,274 @@
...
+        recordNativeSources(item, source, path);
+      } else {
+        for (const existing of Object.keys(sources)) {
+          if (existing.startsWith(`${path}.`)) delete sources[existing];
+        }
+        sources[path] = source;
+      }
+    }
+  };
+  recordNativeSources(DEFAULT_CODEX_CONFIG, "default");
```
mldangelo, line 320: Architectural note anchored here rather than repeated per line.

`sources` is a `Record<string, ConfigurationSource>` keyed by hand-built dotted strings, with six write sites across two modules: the seed at :138, the `choose()` closure, `recordNativeSources` here (recursive, with `delete`s), `resolveCliScope`'s separate map, `Object.assign(provenance.sources, scope.sources)` in cli.ts, and the dry-run patch at `cli.ts:6787` that mutates `arguments_.projectConfig.sources` in place.

The result is a public output contract assembled by mutation, with no exhaustiveness checking, nothing linking a key to the setting it describes, and an ordering dependency between `resolveScanSettings` and `resolveCliScope` that is invisible at both call sites. A `ProvenanceKey` union plus a single builder returning a frozen map would make it checkable and let the deep-source reconciliation happen in one place instead of three.

sdk/typescript/src/scan-settings.ts
```diff
@@ -0,0 +1,121 @@
+import { z } from "zod";
+import { DEFAULT_DEEP_SCAN_SETTINGS } from "./deep-scan-defaults.js";
+import type { Finding, SeverityLevel } from "./models.js";
+import type { ScanTarget } from "./targets.js";
```
mldangelo, line 7: Type-only import cycle: this imports `ScanTarget` from `targets.ts`, and `targets.ts:10` imports `ScanMode` back from here and re-exports it.

Erased at runtime so there is no real cycle, but `ScanMode` now has no canonical home — `deep-config.ts:13` reaches it through `targets.js` while taking `DEEP_SCAN_SETTINGS` and `DeepScanSettingsSchema` from `scan-settings.js`, i.e. two import paths to one module. Either move `ScanTarget` into a leaf types module, or leave `ScanMode` where it was.

sdk/typescript/src/scan-settings.ts
```diff
@@ -0,0 +1,121 @@
...
+  "low",
+] as const;
+export const SCAN_SEVERITIES = [
+  ...REPORTABLE_SEVERITIES,
+  "informational",
+] as const;
+export const FailureSeveritySchema = z.enum(REPORTABLE_SEVERITIES);
+export type FailureSeverity = z.infer<typeof FailureSeveritySchema>;
+
+export const DEEP_SCAN_SETTINGS = [
```
mldangelo, line 27: The third tuple element duplicates a bound the schema already owns. `["subagents", "subagents", 0]` exists only so `deep-config.ts:44` can pick between "a non-negative" and "a positive", while `DeepScanSettingsSchema.shape.subagents` is the actual validator.

Change `.nonnegative()` to `.positive()` and this tuple silently lies. Derive the wording from the schema, or keep the bound here and out of the schema — but not both.

sdk/typescript/src/scan-settings.ts
```diff
@@ -0,0 +1,121 @@
...
+  ScanSettings,
+  | "scanPrompt"
+  | "scanPromptFile"
+  | "validationPrompt"
+  | "validationPromptFile"
+  | "postScanPrompt"
+  | "postScanPromptFile"
+>;
+
+export function scanSettings(settings: ScanSettings): ScanSettings {
```
mldangelo: `scanSettings()` shares its name with the `ScanSettings` type, so `scanSettings(arguments_)` reads like construction when it is really "pick the keys that are defined". `pickScanSettings` or `definedScanSettings` would read correctly at every call site.

It also derives its key list from `Object.keys(ScanSettingsSchema.shape)` at runtime, which couples the picker to zod internals and means anything absent from the schema is silently dropped rather than rejected.

sdk/typescript/src/scan-settings.ts
```diff
@@ -0,0 +1,121 @@
...
+    ...Object.keys(ScanSettingsSchema.shape),
+  ] as (keyof ScanSettings)[];
+  return Object.fromEntries(
+    keys
+      .filter((key) => settings[key] !== undefined)
+      .map((key) => [key, settings[key]]),
+  ) as ScanSettings;
+}
+
+export function meetsSeverity(
```
mldangelo, line 155: An unrecognised `threshold` makes `SCAN_SEVERITIES.indexOf` return `-1`, so this returns `false` for every finding — a policy that silently passes everything.

This is now reachable from the public `ScanResult.hasFindingsAtOrAbove()`, where the argument is `SeverityLevel` rather than the four reportable levels the CLI constrains itself to. Worth throwing on an unknown threshold instead.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,157 @@
...
+import { writeCodexConfig, type JsonObject } from "./config.js";
+import { DEFAULT_DEEP_SCAN_SETTINGS } from "./deep-scan-defaults.js";
+import { CodexSecurityError } from "./errors.js";
+import {
+  DEFAULT_SCAN_MODE,
+  DEEP_SCAN_SETTINGS,
+  DeepScanSettingsSchema,
+  type DeepScanOptions,
+} from "./scan-settings.js";
+import type { ScanMode } from "./targets.js";
```
mldangelo: `ScanMode` imported via the `targets.ts` re-export while everything else in this file comes from `scan-settings.js` directly — see the note on `scan-settings.ts:4`.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,157 @@
...
+      throw new CodexSecurityError("Deep scan settings require deep mode.");
+    }
+    if (!DeepScanSettingsSchema.shape[name].safeParse(value).success) {
+      if (name === "maxTimeHours") {
+        throw new CodexSecurityError(
+          "Deep scan maxTimeHours must be a positive number no greater than 96.",
+        );
+      }
+      throw new CodexSecurityError(
+        `Deep scan ${name} must be ${minimum === 0 ? "a non-negative" : "a positive"} integer.`,
```
mldangelo: This message names the SDK spelling regardless of which input layer produced the value. Confirmed by probing `resolveDeepScanConfig` with a bad ambient file:

```
MESSAGE: Deep scan stopAfterNoNew must be a positive integer.
MESSAGE: Deep scan maxTimeHours must be a positive number no greater than 96.
```

The user wrote `stop_after_no_new` in `~/.codex/codex-security/config.toml`, and no file is named — while the adjacent errors in this same function ("Unknown Codex Security Deep Scan configuration ... in ${source}", "[deep_scan] at ${source} must be a TOML table") do name it, and `source` is in scope right here.

This matters more now than before the PR, because deep preflight validates applicable legacy TOML earlier, so users hit these messages more often. The same string also serves project-file and CLI input, so it is wrong for at least two of three callers whatever it says. Threading the source layer plus the spelling through from `DEEP_SCAN_SETTINGS` fixes the class.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,157 @@
...
+  return {
+    settings,
+    sources,
+    source,
+    document,
+    hasOverrides: Object.keys(explicit).length > 0,
+  };
+}
+
+export async function writeDeepScanConfig(
```
mldangelo, line 119: When `source` and `destination` canonicalise to the same path and there are overrides, this replaces the whole `[deep_scan]` table with all six resolved values. The previous `prepareDeepScanConfig` wrote `{...existing, ...overrides}`.

Reproduced:

```
before:  [deep_scan]        after resolve({workers: 8}) + write:
         workers = 7            workers = 8
         [other]                subagents = 3
         keep = true            stop_after_no_new = 4
                                stop_after_consecutive_errors = 3
                                max_discovery_runs = 40
                                max_time_hours = 96
                                [other]
                                keep = true
```

`[other]` is preserved, which is the fix noted in the description. But five values the user never set are now pinned into their file, so future default changes will not reach them.

Reachability is narrow — the default layout puts the destination at `$CODEX_HOME/state/plugins/codex-security/codex-home/codex-security/config.toml` and the source at `$CODEX_HOME/codex-security/config.toml`, so this needs `CODEX_SECURITY_STATE_DIR` or a link. Flagging as a note rather than a finding, but the risk section covers preserving unrelated sections and not this.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+  await writeCodexConfig(destination, {
+    ...document,
+    deep_scan: Object.fromEntries(
+      DEEP_SCAN_SETTINGS.map(([name, key]) => [key, resolved.settings[name]]),
+    ),
+  } as JsonObject);
+}
+
+async function canonicalConfigPath(path: string): Promise<string> {
```
mldangelo: Unguarded recursion, one frame per path segment. `dirname` reaching a fixpoint terminates it, so this is not a live bug, but a loop would be both cheaper and obviously bounded.

Separately: both `canonicalConfigPath` calls in `writeDeepScanConfig` are `.catch(() => null)`, so a non-ENOENT failure (EACCES on the ambient directory, say) silently degrades to "these are different paths" and takes the overwrite branch. The `source !== null` guard stops `null === null` from matching, which is right, but the swallowed error means a permissions problem presents as a surprising write rather than a diagnostic.

sdk/typescript/src/prompt-files.ts
```diff
@@ -0,0 +1,103 @@
...
+    scanPrompt:
+      options.scanPrompt !== undefined || scanPrompt?.trim()
+        ? scanPrompt
+        : undefined,
+    validationPrompt,
+    postScanPrompt:
+      options.postScanPrompt !== undefined || postScanPrompt?.trim()
+        ? postScanPrompt
+        : undefined,
+    scanPromptFile: undefined,
```
mldangelo, line 62: Contract change worth a comment in the code: the previous `readPromptFiles` *omitted* blank keys, this returns them as explicit `undefined`.

That difference is load-bearing. Every `{...options, ...prompts}` spread now actively clears `scanPromptFile`/`validationPromptFile`/`postScanPromptFile`, which is what stops `#validateLocalInputs` re-reading the files on its second invocation per workflow scan. It is correct everywhere today, and completely invisible to anyone editing either side later.

sdk/typescript/src/api.ts
```diff
@@ -479,6 +468,11 @@ export class CodexSecurity {
       { ...options, outputDir: undefined, archiveExisting: false },
       signal,
     );
+    options = {
+      ...options,
+      ...local.prompts,
+      ...local.deepScanConfiguration?.settings,
```
mldangelo: **Blocking.** This spreads the fully resolved deep settings into `options`, and `options` is what `workflowDigest` hashes twelve lines down at :486. `local.deepScanConfiguration.settings` is `Required<DeepScanOptions>` — all six values resolved from ambient TOML plus shipped defaults — and `workflowDigest` is `JSON.stringify` + sha256, so `undefined` keys drop out but numbers do not.

`workbench_finding_workflows.py:90` treats a changed binding as fatal:

```python
if field in state and state[field] != value:
    raise SystemExit(f"Workflow {state['id']} is already bound to a different {field}. Use another --workflow-id.")
```

Measured with the real function:

```
pre-742 (identical command) : 1279abaf39c95e87
post-742                    : 95941751b4d0133c DIFFERS
after a default bump        : d64f24a4ab2c6b35 DIFFERS
after an ambient TOML edit  : 7e5e9486b43e733e DIFFERS
```

So bumping any deep default in a release, or a user editing `~/.codex/codex-security/config.toml`, makes every in-flight deep workflow unresumable — with an error telling them to abandon the ID. The same command stops producing the same digest.

This is the mirror image of the care taken for recipes ("Complete saved settings do not depend on today's ambient configuration"). The digest should hash the request, not its resolution — hash `scanSettings(options)` captured before this spread, or exclude the resolved deep block at the `workflowDigest` call.

Worth a regression test: same command, mutated `DEFAULT_DEEP_SCAN_SETTINGS`, digest unchanged.

sdk/typescript/src/api.ts
```diff
@@ -791,7 +788,10 @@ export class CodexSecurity {
         outputDir: requestedOutput,
         protectedRoot,
         stateDirectory,
+        deepScanConfiguration,
+        prompts,
       } = await this.#validateLocalInputs(repository, options, signal);
+      options = { ...options, ...prompts };
```
mldangelo: Reassigning the `options` parameter (also at :471) means `options` denotes different things above and below one statement, inside a 3,700-line file. A distinct `resolved` binding would keep the original request addressable — which finding 1 needs anyway, since the digest wants the pre-resolution value.

sdk/typescript/src/api.ts
```diff
@@ -1042,8 +1040,10 @@ export class CodexSecurity {
         options.failureSeverity,
         knowledgeBase?.sources,
         options.maxCostUsd,
-        deepScanOptions(options),
+        deepScanConfiguration?.settings,
```
mldangelo: `scanRecipe` is now 11 positional parameters with four optional and adjacent (`failOnSeverity`, `knowledgeBasePaths`, `maxCostUsd`, `deepScan`, `auth`); this PR added two of them. A transposition here is a silent, persisted corruption of the recipe rather than a type error. An options object would make the next addition free and this call site self-describing.

sdk/typescript/src/api.ts
```diff
@@ -2421,6 +2422,26 @@ export class CodexSecurity {
       outputDir: requestedOutput,
       protectedRoot,
       stateDirectory,
+      prompts,
```
mldangelo, line 3395: `#validateLocalInputs` now performs I/O — reading prompt files above, parsing ambient TOML here — and is called from four sites (:466, :586, :683, :793), twice per workflow scan.

The second call avoids re-reading the ambient file only because `#runWorkflow` spread the resolved settings back into `options` at :471, which makes `explicit` complete and short-circuits `readDeepScanDocument`. That is a subtle and undocumented dependency between two distant functions, and it is the same line that destabilises the workflow digest. Memoising the resolved configuration on the inputs struct would decouple the two, so fixing the digest does not silently reintroduce a second ambient read.

The name is also now wrong — this validates *and* loads. `#prepareLocalInputs` would set the right expectation.

sdk/typescript/src/cli.ts
```diff
@@ -2804,17 +2690,19 @@ export async function main(
       }),
       options: z
         .object({
-          workflowId: optionValue("--workflow-id")
+          config: optionValue("--config")
```
mldangelo: This is the only `--config` in the file. `bulk-scan` (:3374) and `scan-components` (:3125) do not accept it.

"Reusing scan settings means repeating command-line options" is a much bigger problem across many repositories than for a single scan, so `bulk-scan` is the strongest case for this feature and currently cannot use it. Adding it now, while `ScanArguments`/`resolveScanSettings` are fresh, is far cheaper than re-deriving the precedence rules per command later.

Two related gaps while this surface is open:

- **No `init`.** This PR ships a format, a generated schema, an npm export path for it, and two doc examples, with no way to produce a starting file. The `$schema` line users need is `./node_modules/@openai/codex-security/schemas/project-config.schema.json`, which nobody types from memory — and the files in `docs/examples/` use a repo-relative path that is wrong to copy into a real project.
- **Still one flag per invocation.** Refusing discovery is right, since this tool reads untrusted repositories. But an operator-set environment variable keeps that property exactly — the operator controls the environment, repository content does not — and removes the repetition for CI and local iteration alike. `CODEX_SECURITY_CONFIG_PATH` is already taken by the runtime config path, so it would need a distinct name.

sdk/typescript/src/cli.ts
```diff
@@ -4507,6 +4415,11 @@ function scanArgumentsFromRecipe(
       "This scan does not have a saved launch recipe.",
     );
   }
+  if (recipe["requiresScanPrompt"] === true) {
```
mldangelo: **Blocking.** This refuses the rerun outright, but the `rerun` command at :1805 exposes only `--validation-prompt-file` and `--verbose`. There is no `--scan-prompt-file` and no `-c`, so the operation is unreachable, not merely gated.

Confirmed against `origin/main`: `scanArgumentsFromRecipe` previously had no scan-prompt handling at all, so these reruns worked and silently dropped the instructions. Refusing to omit silently is the right call — but custom validation two lines below gets the recoverable treatment (blocked, then re-openable by supplying the file) and this does not.

Adding `--scan-prompt-file` to `rerun` makes the two symmetric and keeps the capability. The message already points at `--scan-prompt-file`, which reads like it is available here.

sdk/typescript/src/cli.ts
```diff
@@ -6884,7 +6784,29 @@ async function executeScan(
       verified: effectivePreflight.authentication.verified,
     });
     progress?.stopTimer();
-    return { exitCode: 0, data: { dryRun: true, ...effectivePreflight } };
+    if (arguments_.projectConfig !== undefined) {
+      for (const [name, key] of DEEP_SCAN_SETTINGS) {
+        const source = preflight.deepScanSources?.[name];
```
mldangelo: This mutates `arguments_.projectConfig.sources` in place to fold `preflight.deepScanSources` into a map built by `project-config.ts`. Three modules now co-own one mutable object through stringly-typed keys, and `scan.deep.${field}` repeats the `subagents_per_worker` special case from `project-config.ts:184`.

It also only runs on the dry-run branch, so the reconciliation exists in exactly one code path. Returning a new merged map from a single provenance builder would remove the mutation, the duplication, and the path-dependence together.

sdk/typescript/src/cli.ts
```diff
@@ -7430,25 +7357,70 @@ function quoteCliPath(path: string): string {
...
-  if (arguments_.paths.length > 0) return arguments_.paths;
-  if (arguments_.diff !== undefined) {
-    return DiffTarget.refs({
-      base: arguments_.diff,
-      head: arguments_.head ?? "HEAD",
-    });
-  }
-  if (arguments_.workingTree) {
-    return DiffTarget.workingTree({ base: arguments_.base ?? "HEAD" });
+function resolveCliScope(
```
mldangelo, line 8832: `resolveCliScope` reimplements scope selection that `projectScopeTarget` and `resolveScanSettings` also perform, then discards its own `scope` whenever `changed` is false and lets `resolveScanSettings` recompute the same value.

Mutual exclusion of `--path/--diff/--working-tree` is likewise checked twice — once in the zod `.refine` above, once here at :7372. Two copies of one rule in one file will drift.

The ordering here is subtle and correct, and worth a comment saying so: `sources["scan.scope"] = "cli"` is assigned *before* the `--head`/`--base` blocks, which is what lets `--head` alone refine a project-file diff scope while leaving `scan.scope` attributed to `"project"` and only `scan.scope.diff.head` to `"cli"`. That is exactly right and exactly the kind of thing a later edit reorders by accident.

sdk/typescript/src/cli.ts
```diff
@@ -7523,7 +7495,7 @@ export function parseCodexOverrides(
   }
   if (
     (isExternalModelProvider(provider) || provider === "amazon-bedrock") &&
-    !("model" in result)
+    scanModel(mergeCodexOverrides(defaults ?? {}, result)) === undefined
```
mldangelo: Good fix — a file-supplied or profile-supplied model now satisfies the provider requirement, which the previous `!("model" in result)` could not see.

One edge: `scanModel` returns `config["model"]` unnarrowed, so a non-string (`model: 123`) or empty string is `!== undefined` and passes here, then fails later in `scanModelConfiguration` with "The configured Codex model must be a nonempty string." Checking for a non-empty string here would keep the error attached to the flag the user actually typed.

docs/project-configuration.md
mldangelo, line 1: Two additions worth making while this document is being written.

**State the trust model.** The format is explicitly opt-in with no discovery, which implies the file is trusted like the command line — but nothing says so. It is a repo-resident YAML file that reads like `.eslintrc`, and someone will wire `-c codex-security.yaml` into a workflow that runs against fork branches, at which point contributors control the scanner's model, provider, and native Codex config. That is inherent to the feature rather than a defect in it, and one sentence closes it: the file is trusted like the command line, so do not point `-c` inside untrusted repository content.

**Call out the one path-anchoring seam.** The table is accurate, but `scan.scope.paths` being repository-relative while `knowledge_base`, `instructions_file`, `validation_file`, and `output.directory` are config-file-relative is the single place a reader can get it wrong. The `--path` symmetry is a real justification; it is worth writing that justification down next to the table rather than leaving the split to be inferred.

Also missing from the docs: there is no way to check a config without a scan target. `scan . -c f.yaml --dry-run --json` needs a git repository and a positional argument and returns a large preflight payload with provenance nested inside, and `projectConfig` is emitted only when `-c` is passed. The data already exists — `-c` on `info`, or a `--print-config` that resolves and stops, would be what people actually reach for and would make this PR's provenance work visible.

docs/examples/codex-security.yaml
mldangelo, line 1: The `$schema` hint here is `../../sdk/typescript/schemas/project-config.schema.json`, which is correct in-repo and wrong the moment anyone copies this file into their own project — and copying is exactly what an example is for.

The docs body has the right form (`./node_modules/@openai/codex-security/schemas/project-config.schema.json`). Either use the published path in the examples with a comment noting the in-repo variant, or add a short header line saying these two files are repository-local and pointing at the docs form. Same applies to `codex-security.json`.

sdk/typescript/scripts/generate-project-config-schema.mjs
mldangelo, line 1: `generate-deep-defaults.mjs` has a `--check` mode wired into `pnpm run types`; this generator has none, so `schemas/project-config.schema.json` is only refreshed by a full `pnpm build`.

In practice `project-config.test.ts:151` compares the committed file against `projectConfigJsonSchema()`, so drift is caught — this is cosmetic. Mirroring the `--check` flag would make the two generators consistent and would fail with a message naming the fix rather than as a schema diff.

sdk/typescript/src/project-config-schema.ts
```diff
@@ -0,0 +1,120 @@
...
+        ),
+    })
+    .optional(),
+  codex: z
+    .object({
+      model: nonempty.optional(),
+      model_reasoning_effort: nonempty.optional(),
+      model_provider: nonempty.optional(),
+    })
+    .catchall(z.json())
```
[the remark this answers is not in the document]
mldangelo-oai: @codex security dismiss false-positive This format loads operator-selected configuration, with the same authority and override checks as --codex and SDK codexOverrides. The existing SECURITY.md threat model treats chosen tools and configuration as running with the operator's local permissions. Project files are not discovered from scan targets, and the enforced scan permission profile is unchanged. The documentation now explicitly warns that native settings can launch configured MCP processes or select model destinations, and that CI configuration must remain outside untrusted checkouts. See the configuration documentation in commit c847acf3f70a3510f232d357d6924889ce5fd707.

sdk/typescript/src/cli.ts
```diff
@@ -7521,7 +7536,8 @@ export function parseCodexOverrides(
   }
   if (
     (isExternalModelProvider(provider) || provider === "amazon-bedrock") &&
-    !("model" in result)
+    !("model" in result) &&
+    defaults?.["model"] === undefined
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Provider selection uses the effective model from the selected native profile. This update also rejects blank or non-string profile models at the provider flag; the file/profile model regressions remain covered.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export async function resolveDeepScanConfig(
+  options: DeepScanOptions,
+  source: string,
+  signal?: AbortSignal,
+): Promise<ResolvedDeepScanConfig> {
+  const explicit = deepScanOptions({ ...options, mode: "deep" });
+  let document: TomlTable = {};
+  // Complete saved settings do not depend on today's ambient configuration.
+  if (!DEEP_SCAN_SETTINGS.every(([name]) => explicit[name] !== undefined)) {
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Complete snapshots preserve unrelated TOML sections. Partial updates to the ambient file now write only explicit overrides, so inherited defaults remain unpinned. Coverage includes direct paths, directory links, and stale deep keys.

sdk/typescript/src/api.ts
```diff
@@ -1039,8 +1037,10 @@ export class CodexSecurity {
         options.failureSeverity,
         knowledgeBase?.sources,
         options.maxCostUsd,
-        deepScanOptions(options),
+        deepScanConfiguration?.settings,
+        options.auth,
       );
+      if (options.scanPrompt !== undefined) recipe["requiresScanPrompt"] = true;
```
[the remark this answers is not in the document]
mldangelo-oai: Verified at `c847acf3f70a3510f232d357d6924889ce5fd707`. The saved requirement is based on the effective prompt being nonblank. Empty and whitespace-only SDK prompts remain replayable; that regression is retained.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,145 @@
...
+  destination: string,
+  resolved: ResolvedDeepScanConfig,
+): Promise<void> {
+  const [source, target] = await Promise.all([
+    realpath(resolved.source).catch(() => null),
+    realpath(destination).catch(() => null),
+  ]);
+  let document = resolved.document;
+  if (source !== null && source === target) {
+    if (!resolved.hasOverrides) return;
```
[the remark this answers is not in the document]
mldangelo-oai: Verified at `c847acf3f70a3510f232d357d6924889ce5fd707`. A default deep scan leaves a missing ambient configuration untouched. Path identity is resolved through its nearest existing ancestor, including missing files and directories.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,146 @@
...
+    document,
+    hasOverrides: Object.keys(explicit).length > 0,
+  };
+}
+
+export async function writeDeepScanConfig(
+  destination: string,
+  resolved: ResolvedDeepScanConfig,
+): Promise<void> {
+  if (destination === resolved.source && !resolved.hasOverrides) return;
```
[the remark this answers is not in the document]
mldangelo-oai: Verified at `c847acf3f70a3510f232d357d6924889ce5fd707`. Path identity resolves the absolute path and existing ancestors before comparing missing destinations. Relative spellings and directory-link aliases retain the missing-file behavior.

sdk/typescript/src/deep-config.ts
```diff
@@ -0,0 +1,151 @@
...
+  };
+}
+
+export async function writeDeepScanConfig(
+  destination: string,
+  resolved: ResolvedDeepScanConfig,
+): Promise<void> {
+  if (
+    !resolved.hasOverrides &&
+    resolve(destination) === resolve(resolved.source)
```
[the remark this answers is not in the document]
mldangelo-oai: Verified at `c847acf3f70a3510f232d357d6924889ce5fd707`. The existing Windows casing regressions are retained. Identity uses the canonical existing ancestor even when the configuration file or directory is missing; it does not require the final file to exist.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,274 @@
...
+    );
+  }
+  let value: unknown;
+  try {
+    if (extension === ".json") {
+      value = JSON.parse(text);
+    } else {
+      const document = parseDocument(text, { prettyErrors: false });
+      if (document.errors.length > 0) throw document.errors[0];
+      value = document.toJS();
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. The later review supplied a concrete nested-alias expansion failure, so unlimited expansion has been replaced with the YAML parser guard at 10,000. The 150-profile case and 500 profiles sharing a 20-field table still pass. The new nested-expansion case is rejected before configuration resolution.

sdk/typescript/src/project-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Restored the YAML expansion guard at 10,000 and added a nested-alias regression. The existing 150-profile configuration and a 500-profile, 20-field case continue to load.

sdk/typescript/src/project-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. The source builder now includes all nine wrapper settings. Unset values are explicitly attributed to default, and info reports them without requiring a selected file.

sdk/typescript/src/project-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Removed the project non-null assertion by capturing the path base explicitly. Resolved context, prompt, and output paths now carry AbsolutePath, including paths reconstructed for reruns.

sdk/typescript/src/project-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. The rule now lives in deepScanOptions and uses the same neutral deep-mode wording for the CLI and SDK. The API uses its returned settings during preparation instead of calling it only for a throw.

sdk/typescript/src/project-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. DEEP_SCAN_SETTINGS now lists SDK, legacy TOML, project-file, and CLI spellings. Runtime mapping and provenance use the project key directly, and CLI deep-option schemas derive from entries that expose a flag.

sdk/typescript/src/project-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Added a typed ProvenanceKey union and one immutable source-map builder. Scope sources are an explicit resolver input; preflight merges return a new map. Both SDK helpers expose sources, including default and unset settings.

sdk/typescript/src/scan-settings.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Scan modes now have a leaf module, removing the type dependency back from targets to scan-settings. The existing targets export remains compatible.

sdk/typescript/src/scan-settings.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Removed the numeric bounds from the metadata tuple. Validation and its error text now come from the same Zod schemas, including the existing 96-hour maximum.

sdk/typescript/src/scan-settings.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Renamed it to pickScanSettings and documented the filtering contract. It intentionally selects scan keys from broader objects containing workflow controls and callbacks; the shared schema remains the key source instead of adding a second manual list.

sdk/typescript/src/scan-settings.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Unknown thresholds now throw, including on an empty result. Added regressions for empty and nonempty findings, while keeping informational support for existing SDK callers.

sdk/typescript/src/deep-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. deep-config now imports ScanMode directly from the new leaf scan-modes module.

sdk/typescript/src/deep-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Legacy-value errors now name the actual TOML key and source file. Validation wording comes from the same schema that enforces the bound; a regression checks stop_after_no_new with its source path.

sdk/typescript/src/deep-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Same-file partial updates now merge only explicitly supplied values, preserving inherited defaults and unrelated sections. Complete explicit snapshots still replace the deep table so obsolete keys do not contaminate a replay.

sdk/typescript/src/deep-config.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Replaced recursion with an ancestor loop and stopped swallowing source and directory-resolution errors. Retained one existing runtime behavior: a cyclic link at the destination file can be replaced by the atomic write. Source or parent errors still stop the write; both paths have regression coverage.

sdk/typescript/src/prompt-files.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Documented the explicit undefined fields and gave resolved prompts a type that guarantees those file inputs are cleared. Workflows also pass prepared inputs directly to avoid relying on another preparation pass.

sdk/typescript/src/api.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Workflow identity now hashes the explicit request and effective prompt contents before adding ambient/default deep settings. Regressions resume after both an ambient edit and a shipped-default change, while rejecting a changed explicit request.

sdk/typescript/src/api.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. The original options parameter is preserved. Execution uses a separate resolvedOptions binding, and workflow identity retains access to the original request.

sdk/typescript/src/api.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Converted scanRecipe to one typed options object with named fields. The call now makes the saved policy, context, cost, deep settings, and authentication choice explicit.

sdk/typescript/src/api.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Renamed the method to prepareLocalInputs. Workflow execution reuses that prepared input object, including prompts and deep settings; it separately checks the output only when starting new work, so completed workflows can still resume their existing artifacts.

sdk/typescript/src/cli.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Added -c/--config to bulk-scan and scan-components through the shared resolver, including paths, prompts, per-scan limits, and severity policy. CSV mode/scope and component plans retain their target precedence. Added init [file] and optional CODEX_SECURITY_PROJECT_CONFIG, with an explicit -c taking precedence and no discovery. Documentation and command tests cover these paths, including bulk resume and policy exit status.

sdk/typescript/src/cli.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Added scans rerun --scan-prompt-file. Required replacements must be nonempty, and both replacement prompt paths resolve from the invocation directory. Reruns retain saved settings and do not reload project files or the project-config environment variable.

sdk/typescript/src/cli.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Preflight source reconciliation now uses the shared builder and returns a new map. The deep project-key mapping comes from DEEP_SCAN_SETTINGS, and info uses the same source-building path.

sdk/typescript/src/cli.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Removed the duplicate Zod scope-exclusion check and skipped scope reconstruction when no CLI scope override exists. Added the ref-only provenance explanation and retained coverage for head/base refinements and scope replacement.

sdk/typescript/src/cli.ts
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. The provider check now rejects empty, whitespace-only, and non-string effective models while retaining the existing missing-model message. Added selected-profile cases that fail before client construction.

docs/project-configuration.md
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Documented the operator trust boundary and why scope paths are repository-relative while support files are config-relative. Added info -c FILE for inspection without a target or runtime; it reports settings, source labels, and effective model details without dumping raw native values.

docs/examples/codex-security.yaml
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Both example hints now use the installed package path. The YAML header and documentation explain the project-root/local-install assumption; init emits that hint automatically.

sdk/typescript/scripts/generate-project-config-schema.mjs
mldangelo-oai: Addressed in `c847acf3f70a3510f232d357d6924889ce5fd707`. Added --check and wired it into types. The generator reads the source schema through the existing esbuild dependency, so checking does not require a previous SDK build or Bun. Drift errors name pnpm build as the repair.

mldangelo-oai:
@codex review

Please review the current head, `c847acf3f70a3510f232d357d6924889ce5fd707`.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,363 @@
...
+    ) ?? [];
+  const settings: ResolvedScanSettings = {
+    ...pickScanSettings(overrides),
+    auth: choose("auth", file?.auth, overrides.auth) ?? DEFAULT_SCAN_AUTH,
+    mode,
+    target,
+    knowledgeBasePaths,
+    scanPromptFile: choose(
+      "scan.instructions_file",
+      filePath(file?.scan?.instructions_file),
```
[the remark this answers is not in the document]
mldangelo-oai: @codex security dismiss false-positive These support paths are explicitly chosen operator inputs. The existing SECURITY.md threat model gives chosen configuration the operator's local permissions, and the documented contract permits support files outside the target, just like --scan-prompt-file and --knowledge-base. The reported flow assumes an untrusted repository controls the selected configuration; that violates this feature's explicit trust prerequisite. CI documentation requires operator-controlled configuration outside the assessed checkout. There is no target-based discovery. Existing regular-file, file-identity, and actual-repository directory-link checks remain in place. Confinement to the target would break the supported shared-context workflow.

sdk/typescript/src/cli.ts
```diff
@@ -3571,13 +3514,48 @@ export async function main(
...
+            overrides,
+            currentDirectory,
+          );
+          const settings = resolved.options;
+          const prompts = await resolveScanPrompts(
+            settings,
             currentDirectory,
-            options.scanPromptFile,
-            options.postScanPromptFile,
             currentDirectory,
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in `ce186091afda7677a2e177ed134ee03acdd4d961`. Shared bulk prompts no longer use the invocation directory as a scan-repository boundary. Regular-file and file-identity checks remain active, and SDK scans still enforce directory-link checks against their actual repository. The regression reads linked operator prompts outside the invocation directory across both CSV modes and a resumed campaign.

sdk/typescript/src/multiscan.ts
```diff
@@ -252,10 +265,15 @@ async function runCampaign(
...
+          const scanSettings = options.scanOptionsByMode?.[task.mode];
+          const scanPrompt = [
+            options.scanPrompt?.trim() ?? scanSettings?.scanPrompt?.trim(),
+            task.prompt,
+          ]
             .filter(Boolean)
             .join("\n\n");
           const result = await security.run(checkout, {
+            ...scanSettings,
             ...(task.scope === undefined ? {} : { target: [task.scope] }),
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in `ce186091afda7677a2e177ed134ee03acdd4d961`. Bulk scans reject a diff or working-tree scope only when a row would inherit it, before creating output state or a client. Explicit CSV path scopes still take precedence. Tests cover both rejected variants and a CSV path overriding a configured working-tree scope; the limitation is documented.

sdk/typescript/src/project-config.ts
```diff
@@ -0,0 +1,363 @@
...
+    throw new ConfigurationError(
+      "Project configuration must be a .yaml, .yml, or .json file.",
+    );
+  }
+  return extension;
+}
+
+export function projectConfigStarter(path: string): string {
+  const schema =
+    "./node_modules/@openai/codex-security/schemas/project-config.schema.json";
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in `ce186091afda7677a2e177ed134ee03acdd4d961`. Starter hints now locate the invocation directory's local package relative to the chosen configuration file. The default root hint stays unchanged. Nested YAML and JSON cases are covered, and the installed Node smoke check resolves a nested starter's hint to the shipped schema.

mldangelo-oai:
@codex review

Please review the current head, `ce186091afda7677a2e177ed134ee03acdd4d961`.

sdk/typescript/src/cli.ts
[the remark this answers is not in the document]
mldangelo-oai: Follow-up in `ddfeb21a5be23cf5f56557e4c67f4e33b008a1c4`: prompt files are now prepared after reading the CSV, using its actual local repository paths for the existing directory-link checks. Operator-owned links outside those repositories still work. Preparation happens once before the campaign, so every row and the manifest use the same prompt contents; missing source repositories remain per-row failures.

mldangelo-oai:
@codex review

Please review the current head, `ddfeb21a5be23cf5f56557e4c67f4e33b008a1c4`.

mldangelo-oai:
@codex review

Please review current head `eee616956460f2744b0287699975ed2d60c78d2d`, including the base integration and verified CI follow-up. Current validation is recorded in the PR description.

mldangelo (commented):
## Re-review

Re-read against `eee61695`. Every point from the previous round is addressed, both blocking items are fixed, and the two architectural notes I expected to be deferred were actually taken. Details below, then four small notes on the newly added `init`.

### Verification

- `pnpm run types`: pass — and `generate-project-config-schema.mjs --check` is now wired into it alongside the deep-defaults check, so schema drift fails typechecking rather than relying on the test alone.
- `pnpm test`: `2245 pass / 45 skip / 25 fail`. All 25 failures sit in `test-shards.test.ts` (16), `codex-review.test.ts` (7), `skeleton.test.ts` (1), and `release-automation.test.ts` (1). Running those four files against current `origin/main` in a clean worktree gives `25 fail` as well — same count, same files — so these remain local environment artifacts (bun 1.3.3, absent credentials) and this branch introduces no regressions. The count moved from 22 to 25 only because the `main` merge added `codex-review` cases.

### Blocking items

**Workflow digest.** `#runWorkflow` no longer reassigns `options`, and the digest now spreads `...options, ...local.prompts` with no resolved deep block — so it hashes the request rather than its resolution. The regression test at `api.test.ts:209` is better than what I suggested: it covers both axes (`"ambient settings"` and `"shipped defaults"`, the latter by mutating `DEFAULT_DEEP_SCAN_SETTINGS` in place), asserts the workflow still resumes, and asserts that a genuinely different request (`workers: 2`) still rejects. That last assertion is the part that keeps the fix from being over-applied later.

**`rerun`.** `--scan-prompt-file` added, resolved through `resolveCliPath` alongside `--validation-prompt-file`, and `prepareScanArgumentsFromRecipe` now takes both. The two prompt kinds are symmetric.

### Architecture

Both of the structural notes landed rather than being deferred, which materially changes the maintenance story for this surface:

- **`ProvenanceKey` + `configurationSources()`.** Provenance went from a stringly-typed map with six mutation sites to one builder returning `Object.freeze(...)` over a typed key union, seeded from `PROJECT_SETTING_KEYS` so coverage is complete rather than partial. The dry-run path now constructs a new object instead of patching `arguments_.projectConfig.sources` in place, and the deep reconciliation happens in the builder using the tuple's project key — so the `subagents_per_worker` special case is gone from both files.
- **`AbsolutePath` in `config-path.ts`.** `ResolvedScanSettings` now narrows `knowledgeBasePaths`, `scanPromptFile`, `validationPromptFile`, `postScanPromptFile`, and `outputDir` to the branded type. The one genuinely subtle part of this feature — which directory a path is anchored to — is now machine-checked.

Also resolved: `scan-modes.ts` breaks the `scan-settings` ↔ `targets` cycle and gives `ScanMode` a single home; `DEEP_SCAN_SETTINGS` carries all four spellings so error labels can name the layer the user actually typed (`${key} in ${source}` for legacy TOML); bound wording now derives from the schema's own `error` options instead of a duplicated tuple element; `pickScanSettings` reads as the filter it is; `severityThresholdRank` throws instead of silently matching nothing; `scanRecipe` takes an options object; `#validateLocalInputs` became `#prepareLocalInputs` and its result is reused via `preparedInputs` rather than recomputed; the duplicated mutual-exclusion `.refine` is gone and `resolveCliScope` early-exits when no scope flag was given; `canonicalConfigPath` is a loop, and `runtimeConfigPath` narrows the previously swallowed `catch` to `ELOOP` on an actual symlink.

The ambient write-back is now the right shape — same-file writes persist only `resolved.overrides`, with the comment explaining why an ambient file keeps inherited defaults unset. That preserves the property I was worried about (future releases can still move defaults) without giving up the complete snapshot an isolated runtime needs.

### UX

`-c` now covers `scan`, `bulk-scan`, `scan-components`, and `info`; `CODEX_SECURITY_PROJECT_CONFIG` gives operators a set-once path with `-c` taking precedence and reruns explicitly excluded; `init` scaffolds a starter; and `info -c --json` answers "what settings will this use" without a repository or runtime. That closes the whole DX cluster from the last round.

The trust paragraph is sharper than what I asked for — naming MCP server processes and model-service destinations is a more accurate statement of the authority a config file carries than "trusted like the command line" on its own. The new **Batch and component scans** section documenting CSV-row precedence, the diff/working-tree rejection for bulk checkouts, and the `--workers` versus `scan.deep.workers` distinction covers the ambiguity that adding `-c` to those commands would otherwise have introduced.

### Notes on `init` (new since the last round, non-blocking)

1. **The JSON starter carries no guidance.** `projectConfigStarter` returns `{"$schema": "..."}` and nothing else for a `.json` destination, against twenty commented lines for YAML. JSON genuinely cannot carry the comments, so this may be as good as it gets — but `codex-security init config.json` currently produces a file that teaches the user nothing. Worth either printing the settings summary to stderr on a JSON init, or having the command output point at `docs/project-configuration.md`.

2. **A second `init` prints an errno.** `flag: "wx"` plus the generic `errorMessage(error)` yields:

   ```
   codex-security: EEXIST: file already exists, open '/path/codex-security.yaml'
   ```

   Refusing to overwrite is the command's advertised behaviour, so this is the most likely non-success path, not an edge case. A mapped message naming the file and pointing at `-c` to use the existing one would read better.

3. **`mode: 0o600`** on a file the docs describe as credential-free and clearly intend to be committed. Correct for a secret-bearing file, surprising for this one — a shared checkout or a container running as a different user gets an unreadable config. Default permissions subject to umask would match what `init` normally does for tracked configuration.

4. **The `$schema` hint assumes `node_modules` directly under the invocation directory**, so running `init` inside a workspace package with hoisted dependencies emits a path that does not resolve — verified against a `packages/app` layout with `node_modules` at the root. The docs state this limitation explicitly, so this is a suggestion rather than a defect: probing for the file and falling back to the resolved module location would make the hint work in the monorepo case without changing the common one.

mldangelo-oai:
@codex review

Please review head `0d97fe345536adb15bf566095fb7814131d6576d` against `main` at `6750642c94edce8d60112f816631de05dc10efd6`. Preserve project-config settings and prompt preparation while integrating main's mock scans, live budget increases, and severity-classification exports.

Frozen dependency installation, SDK TypeScript check, generated project-schema freshness, plugin bundle generation, changed-file formatting, all three portable plugin source checks, and git diff --check passed. Focused project configuration, CLI/SDK configuration, and mock-scan tests: 108 passed. The full suite is left to CI.

kmbroai (approved):
Reviewed this commit; no blocking code issues found. Focused validation passed after correcting one fixture's permission setting: 381 tests passed, 5 platform tests skipped; TypeScript and generated schema/default checks passed. Current CI checks pass. Merge conflicts with main still need resolution.

PR #753, opened 2026-08-30
ci: maintain a rolling draft release PR

mldangelo-oai:
## Summary

Release proposals are currently assembled by hand, so later merges can leave the proposed version and notes out of date. Maintain one draft release PR per cycle while preserving maintainer edits and the human release decision.

## Changes

- Recompute the proposal from the commit where the current package version first reached `main`. Before 1.0, ordinary changes, including features, bump the patch; explicitly marked breaking changes bump the minor. Release PRs use the repository's squash merge flow, and a new empty draft can open after the preceding version bump merges.
- Append commits without force-pushing and retry concurrent branch changes. Pause successfully for another release PR on main; a closed, retargeted, or ready proposal; and unrelated file or package edits. Recheck before advancing the branch and before requesting review.
- Preserve edited or deleted note sections and the PR description. Incorporate new main commits independently of proposal content, so unchanged release files do not trigger another Codex review. Still post new suggestions for human-owned notes, and recover missed review comments on a rerun.
- Update only the top-level package version while preserving JSON formatting and nested values. Use a shared stable-version check, paginate using GitHub's next-page header, build the plan once, and reuse the pre-write PR snapshot when updating an existing proposal.
- Keep the workflow read-only by default. An explicit manual write run uses a repository-scoped GitHub App; automatic writes after pushes require a separate opt-in. Document rollout, note ownership, and final-review controls.

## Testing

- Focused updater suite: 57 tests passed. Coverage includes temporary Git repositories, concurrent commits, note ownership, successful holds, duplicate prevention, ready/draft transitions, retargeting, orphan recovery, proposal review deduplication, comment recovery, workflow controls, serialized HTTP requests, and pagination.
- SDK types and formatting checks, Node syntax, and whitespace checks passed.
- Live read-only preview selected the expected next patch and paused for the existing manual release PR without changing it.
- Full SDK suite on `eef0e82c`, seed `12345`: 2,152 passed, 43 skipped, zero failures.
- Codex code review on `eef0e82c` reported no major issues; the security review also completed.

## Risk and rollout

This does not change the public CLI, enable automated writes, merge a release PR, or publish a package. Existing publication gates remain in place. Breaking-change detection uses commit/PR metadata, and generated notes still need maintainer review. Ready proposals are frozen; if main advances before merge, return the proposal to draft and rerun the updater. Before marking it ready again, check CI and request a final Codex review if the last review targets an older head.

Missing note-ownership metadata deliberately continues to preserve existing text. An explicit section reset resumes generation; absence of state does not authorize overwriting manual notes. If GitHub has not exposed an updated PR head, a manual rerun can post deferred feedback after the head catches up.

The GitHub App write path has been exercised with serialized API fixtures, but has not been run on a hosted workflow. After merging, configure the repository-scoped App, run a manual write pilot with automatic updates disabled, and verify its draft, CI, and Codex review before enabling push-triggered writes. The existing manual release PR is left untouched.

## Public disclosure review

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -41,6 +41,102 @@ Use `release` and `test` only for changes that do not affect package users. A
 maintainer can apply `skip-release-notes` to exclude another internal change.
 That manual label takes precedence over the title category.
 
+## Version policy before 1.0
+
+While the package is on `0.x`, ordinary changes, including features, use a
+patch release. Breaking changes use a minor release and reset the patch to
+zero. For example, changes after `0.1.23` propose `0.1.24`, or `0.2.0` if any
+included change is breaking. The category of a feature remains **Features**;
+the category does not imply a minor version bump.
+
+The release PR updater recognizes breaking changes from a `!` in a
+Conventional Commit title, a `BREAKING CHANGE:` or `BREAKING-CHANGE:` footer,
+or the `breaking-change` label. `skip-release-notes` affects visibility, not
+version impact. All merges and direct commits after the release boundary
+count toward the next version, including documentation and internal changes.
+Review this policy before enabling automation for `1.x`.
+
+## Rolling release PRs
+
+`node-release-pr` runs after pushes to `main` and can be dispatched manually.
+It defaults to a read-only preview. It does not merge, tag, publish, or change
+the existing publication gates.
+
+The updater keeps one draft proposal on `release/next-<base-version>` and
+recomputes its version from all changes since the current package version
+first reached `main`. A later breaking change changes the version on the
+same PR. Each update incorporates the latest `main` and appends a commit;
+the updater never force-pushes. A concurrent commit causes it to reread and
+retry. It requests Codex review on each updated head.
+
+When the release version merges, the next draft can open immediately, even
+while publication is still running. Until another change reaches `main`,
+that draft leaves the package version unchanged. Do not mark an empty draft
+ready or merge it. Publication of the previous version still has to complete
+and pass the verification steps below.
+
+The updater leaves another open `release:` PR, including a manually prepared
+release, untouched and does not open a duplicate. Finish or close that PR
+before enabling the new flow. Closing an automated proposal pauses its
+cycle; reopen it to resume. Changes to other files on the release branch,
+or changes to package fields other than the version, also pause the updater
+so those edits cannot be lost.
+
+### Editing the draft notes
+
+The committed `.github/release-notes.md` is authoritative. The updater
+drafts highlights from merged titles and lists marked breaking changes for
+migration review; it does not infer migration instructions from source code.
+Review and refine these suggestions before releasing.
+
+- Edit the highlights or upgrade notes on the release PR branch. Keep the
+  surrounding `release-section` comments if you want to retain section
+  boundaries. The bot refreshes a section only while it matches the last
+  generated draft. An edit or deletion makes that section human-owned, and
+  later updates preserve it. Custom prose outside the sections is preserved.
+- Later suggestions appear in a new bot comment on the same PR. They do not
+  replace human-owned notes. The bot updates the version header and PR title
+  but never rewrites the PR description. Review version-specific links in
+  human-owned prose when the proposed version changes.
+- `.github/release-pr-state.json` records the cycle and section ownership.
+  To explicitly regenerate a section, add `"reset": true` to that section's
+  state entry. The next run consumes the reset and resumes automatic updates.
+  Restore any damaged section markers before resetting. Do not delete the
+  state file to reset ownership.
+- Deleting the notes file is preserved too. Restore reviewed notes, or
+  explicitly reset the sections, before merging; publication requires the
+  versioned notes file.
+
+### Preview and enable
+
+With an authenticated `gh` CLI, a local checkout can preview the plan with:
+
+```bash
+RELEASE_PR_DRY_RUN=true node sdk/typescript/scripts/release-pr.mjs
+```
+
+The preview may fetch missing Git objects locally, but performs no GitHub
+writes. Its JSON output includes the proposed files and any reason the
+updater would pause. After the workflow is on `main`, use its **Run workflow**
+form with **dry_run** enabled to test the hosted read-only path.
+
+To test writes, configure a GitHub App installed on this repository
+with **Contents: write** and **Pull requests: write**, set the
+`RELEASE_APP_CLIENT_ID` repository variable and `RELEASE_APP_PRIVATE_KEY`
+secret, then manually dispatch the workflow with **dry_run** disabled.
+This permits a single write run while automatic updates remain disabled.
+Review the resulting draft, its hosted CI, and the Codex review request.
+The workflow requests an App token scoped to this repository so CI on the
+bot's PR does not need the approval required for PR events created by
+`GITHUB_TOKEN`. See [GitHub's token documentation](https://docs.github.com/en/actions/concepts/security/github_token).
+
+After the manual write run is verified, set the `RELEASE_PR_ENABLED`
+repository variable to `true` to allow updates after pushes to `main`.
+Remove it or set it to `false` to return push-triggered runs to previews.
+Manual runs always honor their **dry_run** input, which defaults to a preview;
+disabling automatic updates does not prevent an explicit manual write run.
+Generated PRs leave the disclosure attestations unchecked for maintainer review.
+
 ## Prepare a release
 
 1. Choose the next stable version and update `sdk/typescript/package.json`.
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -69,7 +69,16 @@ repository's enabled merge method, so the whole proposal lands as one
 release boundary commit. A later breaking change changes the version on the
 same PR. Each update incorporates the latest `main` and appends a commit;
 the updater never force-pushes. A concurrent commit causes it to reread and
-retry. It requests Codex review on each updated head.
+retry. It requests Codex review when the proposal files change. Updates that
+only incorporate `main` keep CI current without repeating the same proposal
+review. New suggestions for human-owned notes still appear in a comment.
+Before marking the proposal ready, check CI and request a final Codex review
+if the last review targets an older head.
+
+If a run reports that GitHub has not exposed the updated PR head, manually
+rerun the updater with **dry_run** disabled after the PR catches up. This
+allows any deferred review request or note suggestions to be posted. Verify
+review on the current head before marking the proposal ready.
 
 When the release version merges, the next draft can open immediately, even
 while publication is still running. Until another change reaches `main`,
@@ -86,8 +95,10 @@ ready pauses updates, preserving the reviewed version and notes. To resume,
 convert it back to a draft and rerun the updater. Do this before merging if
 `main` has advanced, then review the updated proposal. The updater rechecks
 these conditions before advancing the branch. Changes to other files on the
-release branch, or to package fields other than the version, also pause the updater
-so those edits cannot be lost.
+release branch, or to package fields other than the version, also pause the
+updater so those edits cannot be lost. These intentional pauses return
+`action: "held"` and leave the workflow successful. Preserve or merge the
+additional changes, then rerun the updater to resume.
 
 ### Editing the draft notes
 
@@ -109,7 +120,9 @@ Review and refine these suggestions before releasing.
   To explicitly regenerate a section, add `"reset": true` to that section's
   state entry. The next run consumes the reset and resumes automatic updates.
   Restore any damaged section markers before resetting. Do not delete the
-  state file to reset ownership.
+  state file to reset ownership. A section with missing ownership metadata
+  is preserved until explicitly reset; missing state does not authorize
+  replacing manual notes.
 - Deleting the notes file is preserved too. Restore reviewed notes, or
   explicitly reset the sections, before merging; publication requires the
   versioned notes file.
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -64,7 +64,9 @@ the existing publication gates.
 
 The updater keeps one draft proposal on `release/next-<base-version>` and
 recomputes its version from all changes since the current package version
-first reached `main`. A later breaking change changes the version on the
+first reached `main`. Squash-merge release PRs, as required by this
+repository's enabled merge method, so the whole proposal lands as one
+release boundary commit. A later breaking change changes the version on the
 same PR. Each update incorporates the latest `main` and appends a commit;
 the updater never force-pushes. A concurrent commit causes it to reread and
 retry. It requests Codex review on each updated head.
@@ -78,7 +80,10 @@ and pass the verification steps below.
 The updater leaves another open `release:` PR, including a manually prepared
 release, untouched and does not open a duplicate. Finish or close that PR
 before enabling the new flow. Closing an automated proposal pauses its
-cycle; reopen it to resume. Changes to other files on the release branch,
+cycle; reopen it to resume. Marking the proposal ready also pauses updates,
+preserving the reviewed version and notes. To resume, convert it back to a
+draft and rerun the updater. The updater rechecks these conditions before
+advancing the branch. Changes to other files on the release branch,
 or changes to package fields other than the version, also pause the updater
 so those edits cannot be lost.
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -77,14 +77,16 @@ that draft leaves the package version unchanged. Do not mark an empty draft
 ready or merge it. Publication of the previous version still has to complete
 and pass the verification steps below.
 
-The updater leaves another open `release:` PR, including a manually prepared
-release, untouched and does not open a duplicate. Finish or close that PR
-before enabling the new flow. Closing an automated proposal pauses its
-cycle; reopen it to resume. Marking the proposal ready also pauses updates,
-preserving the reviewed version and notes. To resume, convert it back to a
-draft and rerun the updater. The updater rechecks these conditions before
-advancing the branch. Changes to other files on the release branch,
-or changes to package fields other than the version, also pause the updater
+The updater leaves another open `release:` PR targeting `main`, including a
+manually prepared release, untouched and does not open a duplicate. Finish
+or close that PR before enabling the new flow. Closing an automated proposal
+pauses its cycle; reopen it to resume. Retargeting it away from `main` also
+pauses updates; restore its `main` base before resuming. Marking the proposal
+ready pauses updates, preserving the reviewed version and notes. To resume,
+convert it back to a draft and rerun the updater. Do this before merging if
+`main` has advanced, then review the updated proposal. The updater rechecks
+these conditions before advancing the branch. Changes to other files on the
+release branch, or to package fields other than the version, also pause the updater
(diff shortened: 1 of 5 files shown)


mldangelo-oai:
@codex review the current head 982ece706dc21702019c998a08a3a37c1188afc7. Please focus on release-cycle boundaries, preservation of human notes, concurrent GitHub updates, and workflow write controls.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,601 @@
...
+  const packageText = repo.readFile(mainSha, packagePath);
+  const baseVersion = releaseVersion(JSON.parse(packageText));
+  let baseCommit = mainSha;
+  for (const sha of repo
+    .git("log", "--first-parent", "--format=%H", mainSha, "--", packagePath)
+    .trim()
+    .split("\n")) {
+    const packageJson = JSON.parse(repo.readFile(sha, packagePath));
+    if (packageJson.version !== baseVersion) break;
+    baseCommit = sha;
```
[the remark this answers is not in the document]
mldangelo-oai: Clarified in 7475203b. The repository currently allows only squash merges, and main requires a pull request. RELEASING.md now explicitly requires that release flow. The rollover fixture now includes multiple updater commits and verifies that squashing their final tree opens an empty next cycle. Replaying a release branch through rebase or cherry-pick is outside the documented release flow.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,601 @@
...
+    const history = await readReleaseHistory(repo, mainSha, github);
+    let plan = createReleasePlan(history);
+    const openPulls = await github.list(
+      "pulls?state=open&base=main&per_page=100",
+    );
+    const pull = openPulls.find(
+      (candidate) =>
+        candidate.head.ref === plan.branch &&
+        candidate.head.repo?.full_name === github.repository,
+    );
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in 7475203b. Ready proposals now pause before writes, and the draft state is rechecked before advancing the branch and before requesting review. Regression tests cover freezing the reviewed version, title, and body; resuming after conversion back to draft; and transitions to ready during preparation and final checks.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,601 @@
...
+      });
+      const commit = await github.request("POST", "git/commits", {
+        message: plan.title,
+        tree: tree.sha,
+        parents: [...new Set([headSha, mainSha].filter(Boolean))],
+      });
+      if ((await branchHead(github, "main")) !== mainSha) continue;
+      try {
+        if (headSha) {
+          await github.request("PATCH", `git/refs/heads/${plan.branch}`, {
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in 7475203b. The updater rereads open release PRs immediately before the Git ref write and checks for competing proposals for both new and existing automated PRs. Regression tests open a manual proposal after commit preparation and verify that the automated branch is neither created nor advanced.

mldangelo-oai:
@codex review the current head 7475203b0303e2fcfe0f309bfdcb4aee80566743. The follow-up freezes ready proposals, rechecks competing releases before the branch write, and documents the existing squash-only release flow. The focused suite now has 44 passing tests.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,623 @@
...
+  dryRun = true,
+}) {
+  // Re-read only when main or the proposal actually moves; API failures are not retried blindly.
+  for (;;) {
+    const mainSha = await branchHead(github, "main");
+    const history = await readReleaseHistory(repo, mainSha, github);
+    let plan = createReleasePlan(history);
+    const openPulls = await github.list(
+      "pulls?state=open&base=main&per_page=100",
+    );
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in 2a69e60a. The cycle lookups now include all bases, and closed-cycle checks no longer filter to main. A matching proposal targeting another base pauses the updater before branch writes or PR creation; unrelated maintenance releases remain outside the main release check. Regression tests cover open and closed retargeted proposals across runs and an orphan branch that gains a non-main PR during recovery.

mldangelo-oai:
@codex review the current head 2a69e60a4fae536a03e839eaef08137759d679ba. The latest fix preserves retargeted proposals across workflow runs and orphan-branch recovery, with 48 focused tests passing.

mldangelo-oai:
Final validation for `2a69e60a4fae536a03e839eaef08137759d679ba`: both full SDK runs passed with 2,143 tests and 43 skips each (seeds `12345` and `29381448`). The 48 focused updater tests also pass. Current CI is green, and Codex reported no major issues.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+  const mergeBase = repo.git("merge-base", mainSha, headSha).trim();
+  const paths = repo
+    .git("diff", "--no-renames", "--name-only", "-z", mergeBase, headSha)
+    .split("\0")
+    .filter(Boolean);
+  if (
+    paths.some((path) => ![packagePath, notesPath, statePath].includes(path))
+  ) {
+    throw new Error(
+      "The release branch has other file edits. Preserve or merge them before running the updater.",
```
mldangelo: **A documented "pause" is implemented as a permanently red job on `main`.**

This throw and the one below it (package edits beyond the version) escape `reconcileReleasePullRequest` and fail the job on *every subsequent push to main*. I edited a file on the release branch and then pushed three changes — three failed runs, no self-healing.

Every other pause condition (`ready`, `closed`, `retargeted`, another release PR) returns `action: "held"` and exits 0. RELEASING.md groups this one with them, and the test that covers it is titled "pauses instead of losing unrelated human edits".

Nothing is lost — `writes` stays flat, so it fails safe. But routing both guards through `releaseHoldReason`'s `held` result would match the docs, match the test name, and stop main's CI going red for a legitimate maintainer action (editing a note by hand, pinning a dependency as part of the release).

RELEASING.md
```diff
@@ -41,6 +41,109 @@ Use `release` and `test` only for changes that do not affect package users. A
...
+The updater leaves another open `release:` PR targeting `main`, including a
+manually prepared release, untouched and does not open a duplicate. Finish
+or close that PR before enabling the new flow. Closing an automated proposal
+pauses its cycle; reopen it to resume. Retargeting it away from `main` also
+pauses updates; restore its `main` base before resuming. Marking the proposal
+ready pauses updates, preserving the reviewed version and notes. To resume,
+convert it back to a draft and rerun the updater. Do this before merging if
+`main` has advanced, then review the updated proposal. The updater rechecks
+these conditions before advancing the branch. Changes to other files on the
+release branch, or to package fields other than the version, also pause the updater
```
mldangelo: This describes the branch-edit guard as pausing the updater, alongside the `ready` / `closed` / `retargeted` conditions. Those four return `action: "held"` and exit 0; this one throws and fails the workflow on every subsequent push to `main` until a maintainer intervenes. Either the wording or the implementation should move — see the comment on `readReleaseBranch`.

sdk/typescript/tests-ts/release-pr.test.ts
```diff
@@ -0,0 +1,1221 @@
...
+    fixture.merge("fix: another fix");
+    const next = await fixture.run();
+    expect(fixture.repo.readFile(next.headSha!, notesPath)).toBeNull();
+    expect(next.plan.humanOwned).toEqual(["highlights", "upgrades"]);
+  });
+
+  test.each([
+    { "sdk/typescript/src/example.ts": "Human implementation changes.\n" },
+    { [packagePath]: packageText("0.1.24", { example: "1.0.0" }) },
+  ])("pauses instead of losing unrelated human edits", async (files) => {
```
mldangelo, line 1173: The name says "pauses", the assertion is `rejects.toThrow("edits")`. Those aren't the same outcome: every other pause in this file asserts `action: "held"`. Worth renaming to match what it actually pins, or changing the behavior so the name becomes true.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+  }
+  const version = nextReleaseVersion(baseVersion, changes);
+  const generated = generateNoteSections(changes);
+  const { notes, sections } = updateReleaseNotes(
+    version,
+    generated,
+    previous?.notes ?? null,
+    previous?.state.sections,
+  );
+  const state = { baseVersion, baseCommit, mainSha, sections };
```
mldangelo: **`mainSha` is written into the committed state but never read back** — `createReleasePlan` only reads `baseVersion` and `baseCommit` from `previous.state`.

Its one effect is making `changed` true whenever main moves, which appends a commit, which produces a new `commitSha`, which posts a fresh `<!-- release-pr-head: … -->` comment ending in `@codex review the current head …`.

Verified with the fixture: merging `test: add coverage` (filtered out of the notes, no version change) gives `action: "updated"` with `package.json` and `release-notes.md` **byte-identical** across the two heads — and a second Codex review request. Over a cycle with N merges that's N reviews of a draft whose reviewable content changed maybe twice.

Keeping the branch on top of main is worth doing so PR CI stays meaningful, but it shouldn't be driven by a write-only field. Suggest dropping `mainSha` from the state, advancing the branch on an explicit "first parent ≠ mainSha" check, and gating the review comment on `plan.files` actually differing.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+export async function reconcileReleasePullRequest({
+  repo,
+  github,
+  dryRun = true,
+}) {
+  // Re-read only when main or the proposal actually moves; API failures are not retried blindly.
+  for (;;) {
+    const mainSha = await branchHead(github, "main");
+    const history = await readReleaseHistory(repo, mainSha, github);
+    let plan = createReleasePlan(history);
```
mldangelo: This plan is built only to read `plan.branch`, then discarded and rebuilt at line 489 — which is also why `plan` has to be a `let`.

The branch name is a pure function of `history.baseVersion`, so hoisting it (`const branch = ...`) makes the data flow linear and lets the plan be built once, after `readReleaseBranch`. It also collapses the `release/next-` literal that's currently duplicated between `isReleasePull` (line 17) and `createReleasePlan` (line 189).

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+    execFileSync("gh", ["auth", "token"], {
+      encoding: "utf8",
+      stdio: ["pipe", "pipe", "pipe"],
+    }).trim();
+  const result = await reconcileReleasePullRequest({
+    repo: createGitRepository(directory),
+    github: createGitHubClient(repository, token),
+    dryRun: process.env.RELEASE_PR_DRY_RUN !== "false",
+  });
+  // Do not copy PR bodies into logs or the generated proposal.
```
mldangelo: Dead code. Change objects are built at lines 257-263 as `{sha, number, title, labels, url}` and never carry `body` — it only exists as a local inside `readReleaseHistory`, passed transiently to `isBreakingChange`. So the map copies every change for nothing, and the comment warns about a hazard the code can't produce.

Delete both. If the invariant is worth recording, a one-liner where the change is constructed ("bodies are deliberately not retained") documents it where it's actually enforced.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+    request,
+    async list(path) {
+      const items = [];
+      for (let page = 1; ; page += 1) {
+        const batch = await request(
+          "GET",
+          `${path}${path.includes("?") ? "&" : "?"}page=${page}`,
+        );
+        items.push(...batch);
+        if (batch.length < 100) return items;
```
mldangelo: The `100` here is decoupled from the `per_page` each caller writes into its path. I confirmed `list("pulls?state=open&per_page=30")` stops after one page and silently returns a truncated list.

All four current callers pass 100, so it's latent — but it's a foot-gun in a shared helper. `src/github.ts:169` already paginates off the `Link` `rel="next"` header, which plain `fetch` exposes via `response.headers.get("link")`. Failing that, have `list` append its own `per_page` constant rather than trusting the caller to keep them in sync.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+export function isBreakingChange(change) {
+  return (
+    conventionalTitle.exec(change.title)?.[2] === "!" ||
+    /^(?:BREAKING CHANGE|BREAKING-CHANGE):\s+\S/mu.test(change.body ?? "") ||
+    (change.labels ?? []).includes("breaking-change")
+  );
+}
+
+export function nextReleaseVersion(version, changes) {
+  releaseVersion({ name: "@openai/codex-security", version });
```
mldangelo: This validates a version string by fabricating a package object and discarding the result, and re-hardcodes the package name that `release-automation.mjs` already keeps in `packageName`. Exporting the `stableVersion` check (or a small `assertStableVersion`) would say what it means and drop the duplicated literal.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+    title:
+      changes.length > 0
+        ? `release: bump Codex Security to ${version}`
+        : "release: prepare the next Codex Security release",
+    changes,
+    generated,
+    humanOwned: sectionIds.filter((id) => sections[id].humanOwned),
+    files: {
+      [packagePath]: packageText.replace(
+        /("version"\s*:\s*")[^"]+(")/u,
```
mldangelo: This takes the *first* `"version":` in the file. I built a `package.json` with a nested `config.version` ahead of the top-level key: the nested value gets bumped, the real version silently stays at the old value, and nothing throws — so the PR title and notes header would advertise a version the package doesn't have, surfacing only later at tag time.

Can't happen with today's `sdk/typescript/package.json` (version is line 3), so this is hardening rather than a live bug. Anchoring (`/^(\s*"version"\s*:\s*")[^"]+(")/mu`) plus re-parsing to confirm the result is a two-line fix for a silent-wrong failure mode.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+    notes = /^<!-- release-version: [^\r\n]* -->/u.test(notes)
+      ? notes.replace(/^<!-- release-version: [^\r\n]* -->/u, header)
+      : `${header}\n\n${notes}`;
+  }
+  for (const id of sectionIds) {
+    const previous = previousSections[id];
+    const block = notes === null ? null : findSection(notes, id);
+    const humanOwned =
+      previous?.reset !== true &&
+      (previous?.humanOwned !== false ||
```
mldangelo, line 144: `previous?.humanOwned !== false` treats *absent* the same as *human-owned* — and then writes that back — so a section id missing from a saved state is locked human-owned forever. Verified it survives reruns.

Only bites if `sectionIds` gains an entry mid-cycle, in which case the new section would never be generated on any existing branch. `previous === undefined` probably wants to fall through to the generate path rather than join the sticky one.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+    (candidate) =>
+      candidate.number !== pullNumber &&
+      candidate.base.ref === "main" &&
+      isReleasePull(candidate),
+  );
+  if (other)
+    return `Another release PR is open: #${other.number}. It has not been changed.`;
+  const current = openPulls.find(
+    (candidate) => candidate.number === pullNumber,
+  );
```
mldangelo, line 418: "Not a draft" is expressed twice with the same sentence and opposite outcomes — here it's a `held` result, in `assertReleasePull` (line 370) it throws.

Which one a maintainer hits depends on where in the run they click Ready: I reproduced a red run for a mid-write click, with the next push reporting `held`. It self-heals, so this is minor, but the duplicated message strings should collapse into one source of truth.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+  if (current && !current.draft)
+    return `Release PR #${current.number} is ready for review. Convert it back to a draft to resume automatic updates.`;
+  return null;
+}
+
+async function ensurePullRequest(github, plan, pull, template, headSha) {
+  const holdReason = releaseHoldReason(
```
mldangelo, line 430: This re-lists all open PRs immediately after the identical call at line 525, with only the ref write in between. Post-write it adds only "another release PR appeared" — the closed / retargeted / draft checks for *this* PR are already covered by the two `assertReleasePull` calls below.

Passing `currentPulls` down (or making this conditional on the unchanged path, where it does earn its keep) would drop a full paginated list per update run.

sdk/typescript/scripts/release-pr.mjs
```diff
@@ -0,0 +1,642 @@
...
+      title: plan.title,
+      head: plan.branch,
+      base: "main",
+      draft: true,
+      body: initialPullBody(template),
+    });
+  }
+  const current = await github.request("GET", `pulls/${pull.number}`);
+  assertReleasePull(current, plan);
+  if (current.head.sha !== headSha) {
```
mldangelo, line 471: `pulls/N.head.sha` is a denormalized field that GitHub updates asynchronously, so it can lag the ref you just wrote and verified at line 561. When it does, the run returns `reviewRequested: false` and relies on "a subsequent run".

With only a `push: main` trigger, if no further push arrives, that final head never gets its Codex review — and the proposal that eventually merges is the one that wasn't reviewed. Worth either retrying the read briefly or calling the gap out in RELEASING.md so a maintainer knows to dispatch manually.

mldangelo (commented):
Reviewed the full diff, ran the suite (48 pass), prettier and actionlint clean, and probed runtime behavior with this PR's own test harness.

The harness is unusually strong — real temporary git repos plus a fake GitHub that actually rejects non-fast-forwards and asserts `force: false`, so the append-only claim is genuinely tested. `markdownText` holds up against a hostile PR title, the `RELEASE_PR_DRY_RUN` expression is correct in all four event/var combinations and fails safe, and comparing branch `package.json` against the merge base minus the version key is a clean way to detect human edits.

Two things I'd want resolved before setting `RELEASE_PR_ENABLED`: the branch-edit guard turns `node-release-pr` permanently red on `main` for a documented maintainer action, and every push to `main` triggers a fresh `@codex review` even when the proposal is byte-identical. The rest are simplifications and small hardening notes. Details inline.

mldangelo-oai:
@codex review the current head eef0e82cfdd9de5decbc8b775fbd94c98b2d294d. Please focus on successful release holds, review deduplication, preservation of human notes, GitHub pagination, and top-level package version updates. Missing ownership metadata deliberately continues to preserve existing text until an explicit reset.

sdk/typescript/tests-ts/release-pr.test.ts
kmbroai, line 1: might want to ask codex to simplify these tests. seem a bit overkill for what we actually need

kmbroai (approved):

PR #756, opened 2026-08-30
feat: add Azure OpenAI as a scan provider

mldangelo:
## What

Adds `azure` to `--provider`, so scans can run against an Azure OpenAI resource.

## Public CLI change

- **User need.** Azure OpenAI is where many organizations are permitted to send
  source code at all; today reaching it requires hand-writing a
  `model_providers` table through `--codex`.
- **Syntax.** `codex-security scan . --provider azure --model <deployment>`.
  The endpoint comes from `AZURE_OPENAI_BASE_URL` or `AZURE_OPENAI_ENDPOINT`
  and the credential from `AZURE_OPENAI_API_KEY`.
- **Defaults.** `--provider` still defaults to `openai`. Nothing changes for
  any existing invocation.
- **Why existing behavior is insufficient.** `--codex` works but leaves the
  provider outside authentication reporting, credential scoping, worker
  environment passthrough, and saved recipes.
- **Compatibility.** Additive. One new enum value, no new flags, no changed
  defaults.

## Why this shape

Azure's v1 surface (`/openai/v1`) takes no `api-version`, and it accepts a
bearer token for **both** a resource key and a Microsoft Entra token. I
confirmed all three combinations return 200 against a live resource, so the
existing `env_key` mechanism covers Azure with no `query_params`,
`env_http_headers`, or other new auth machinery. The only structural difference
from the existing external providers is that the endpoint is per-resource, so
`base_url` is resolved from the environment and normalized to the v1 path. The
resource endpoint Azure prints works as-is; `/openai/v1` is appended when
absent, which covers the `openai.azure.com`, `services.ai.azure.com`, and
`cognitiveservices.azure.com` domains Azure hands out.

Provider tables are now built in one place, so each provider's endpoint,
credential, environment scrubbing, and recipe fields derive from the registry
instead of parallel literal lists. Saved recipes keep the resolved endpoint so a
replay targets the same resource; provider secrets are still stripped.

## Drive-by fix

Auth failures advised `--auth chatgpt`, which cannot reach a third-party
endpoint. That guidance was already wrong for `openrouter` and `fireworks`; it
is now suppressed for any external provider.

## Testing

`azure-provider.test.ts` adds 23 tests covering endpoint normalization and
rejection, the three Azure domains, override construction, the deployment-name
requirement, authentication selection, cross-provider credential scrubbing, and
recipe preservation.

Verified end to end against a live Azure OpenAI resource: a scan of a fixture
repository with a command-injection bug completed with `COVERAGE complete` and
reported it as CWE-78, high severity and high confidence, at the correct line.
Cost reporting worked, since the deployed model is in the cost table.

Four earlier runs on other deployments each failed outside this change and are
worth recording, because two became documentation:

| Deployment | Result |
| --- | --- |
| Reasoning model, 150 TPM | Sustained 429s — quota too small for scan prompts |
| Non-reasoning model, high TPM | `Unsupported parameter: 'reasoning.effort'` |
| Older reasoning model | All phases ran; agent wrote no artifacts |
| Older general model | All phases ran; artifacts failed schema validation |

Only the model differed between the last of those and the passing run, so the
provider path is not implicated. The README now states that an Azure deployment
must serve a reasoning model and needs enough tokens-per-minute.

## Known limitation

`--max-cost` works for deployments whose model is already priced in
`MODEL_PRICING_NANODOLLARS` and errors clearly for others. Adding Azure-specific
pricing is left out, since it varies by region and tier.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RTv1bYexiubXSVTz2rVt6b


sylvesterkaczmarek (commented):
`azureBaseUrl()` accepts `http:` even though this provider sends `AZURE_OPENAI_API_KEY` (resource key or Entra token) to that endpoint. A misconfigured plaintext URL can expose credentials in transit. Please require HTTPS, or explicitly restrict HTTP to loopback-only test endpoints, and add a rejection test.

PR #767, opened 2026-08-31
fix(deep-scan): remove coverage from aggregation

daneschneider-oai:
## Summary

Deep scans combine several independent reviews. A scan can reach its stopping point and still be labeled partial because one worker left notes about unfinished work. That makes the final status confusing: it reflects one worker's notes instead of whether the Deep scan finished.

Completion follows the scan's stopping rules, and reducers focus on combining findings. The report includes the configured directories and exclusions, and the host supplies `coverage.json` for existing readers.

## Changes

- Remove `coverage` from reducer tool inputs, returned inputs, saved results, checkpoints, and prompts. Ignore coverage in older saved reducer results when resuming.
- Successful Deep compatibility files contain the configured paths, host-derived completeness, empty `surfaces`, `explicitExclusions`, and `deferred` arrays, and no `openQuestions`. Standard workers keep their original coverage.
- Publish the accepted aggregate directly without rereading workers, recovering checkpoints, repairing findings, or downgrading coverage during successful completion.
- Cancel unfinished workers at saturation; their late findings and failures cannot change the accepted result. Preserve distinct findings with colliding identities using numeric suffixes.
- Support stopped-scan recovery from coverage-free reducer results, including scoped scans without a parent draft and repeated recovery.

## Testing

- MCP typecheck and focused reducer, validation, coordinator, template, and source/bundled tool-contract tests passed.
- Python: 20 schema tests plus 44 subtests, 11 publication/recovery tests, 5 stop-condition tests, and 11 existing recovery tests passed.
- SDK build, types, format, and 92 focused tests passed; the final bundled reducer suite also passed.
- End-to-end coordinator → native workbench → finalizer → SDK loading verified configured scope in the report, empty compatibility observations, and unchanged worker and legacy artifacts.
- Ruff, portable source checks, and diff checks passed. Full SDK suite completion remains unverified locally.

## Risk and rollout

The reducer contract changes; canonical output schemas stay compatible. Existing readers still use the host-generated `coverage.json`, including completeness for status and comparisons. Successful Deep reports omit worker/reducer coverage observations.

Configured scope exclusions, Standard/diff behavior, stop thresholds, no-review time-limit diagnostics, and stopped-scan findings recovery remain supported. No version bump.

## Public disclosure review


daneschneider-oai:
@codex review

sylvesterkaczmarek (commented):
This can report a Deep scan as complete even when the accepted workers never actually completed repository inspection.

The reducer input now strips every worker's `coverage`, and the final coordinator then synthesizes `coverage.completeness = "complete"` with empty surfaces/deferred entries whenever it has an accepted reducer result. Saturation also suppresses late discovery failures. As a result, evidence such as partial/deferred worker coverage or an inspection failure no longer reaches the final completeness decision.

The PR's own Scanbench note appears to demonstrate the consequence: three scans were blocked before repository inspection by sandbox namespace errors but still reported completion.

Reducer completion/saturation is not the same property as coverage completeness. Could the host retain an independent execution/inspection signal from accepted workers and refuse to emit `complete` when configured scope was not successfully inspected? A regression where workers return partial/deferred coverage (or cannot inspect the target), the reducer returns no findings, and final Deep output must remain partial/failed would pin this boundary.

daneschneider-oai:
@sylvesterkaczmarek Thanks for [the review](https://github.com/openai/codex-security/pull/767#pullrequestreview-5088878202). This is intentional: Deep runs many independent workers, so individual worker coverage judgments do not determine the aggregate status. `complete` means the overall Deep run finished normally; `partial` is for the overall run being stopped or cut off (e.g. by time or budget). It does not certify exhaustive repository coverage. The reducer merges findings only; adding a worker-coverage gate would reintroduce the behavior this PR removes.

faizan-oai (approved):

PR #768, opened 2026-08-31
refactor(ci): port source and test report checks to TypeScript

kmbroai:
## Summary

Replace the portable-source checker and JUnit inventory comparator with typed TypeScript utilities. Compile them with the existing TypeScript compiler and run the emitted JavaScript on Node 22.13.0, removing these utilities' Python dependency without experimental type stripping.

## Changes

- Replace the two Python utilities and source-checker test with three typed `.mts` files while preserving source limits, Markdown checks, tracked-file handling, JUnit outcomes, duplicate detection, and shard comparison.
- Add `build:ci` to emit adjacent ignored `.mjs` files. Compile before running the utilities in CI and update contributor commands, regression tests, and development dependencies.
- Give parameterized cost, budget, release, and matching tests unique JUnit identities. The four boolean-title corrections inherited from current main are in `sdk/typescript/tests-ts/finding-catalogue.test.ts`, `sdk/typescript/tests-ts/cli-workbench.test.ts`, and `sdk/typescript/tests-ts/cli.test.ts`. Only their `%s` title placeholders change to `%p`; fixtures, assertions, comparator behavior, and product code are unchanged.

## Testing

- Frozen SDK and MCP dependency installation, CI-tool compilation, plugin build, and formatting of the three corrected test files passed.
- On Bun 1.3.13, all eight affected boolean cases passed with 50 assertions in each of baseline, isolated, and parallel modes. Each JUnit report has 229 unique identities, including the 221 filtered cases, and records distinct `false` and `true` identities for all four corrected pairs.
- The unchanged comparator accepted both focused mode comparisons.
- Replayed all 13 existing hosted runner reports using only the four known title corrections. Outcomes and test counts were preserved, no additional duplicate groups were found, and all five inventory comparisons passed, including the combined seven Windows shards. This projection verifies the naming correction; fresh hosted execution must still validate the revised head.
- The five existing commits were rebased onto current main without patch changes, verified by `git range-diff`.

## Risk and rollout

The affected CI jobs install SDK development dependencies and compile the utilities before running them. Generated JavaScript is ignored and excluded from the published SDK. Test fixtures compile isolated copies so parallel runs do not share generated files. The comparator continues to reject duplicate identities; corrected titles expose existing parameter values without changing test behavior.

## Public disclosure review


mldangelo-oai (approved):
TypeScript compilation, the source checker tests, and the report/workflow regression tests passed locally. No actionable findings in the port.

The current inventory-comparison failure reproduces with the original Python comparator on the same report and comes from duplicate test names already on main.

sdk/typescript/scripts/compare-test-reports.mts
```diff
@@ -0,0 +1,252 @@
...
+  attributes: SaxesTagNS["attributes"];
+  status: TestStatus;
+};
+type TestReport = {
+  cases: Map<string, number>;
+  duration: number;
+  failed: boolean;
+};
+
+function matchingReports(pattern: string): string[] {
```
mldangelo-oai, line 21: **[P3] Let the workflow shell expand the shard glob**

`matchingReports` is 54 lines plus `minimatch` and glob-specific tests for one fixed call site: an Ubuntu Bash step. Unquote `reports/runner-windows-latest-shard-*.xml` there and have this script read each candidate path directly. Bash expands existing shards; with no match it passes the literal path and `readFileSync` still fails, so aggregation and failure behavior stay intact. This removes roughly 98 net lines and the dependency metadata.

sdk/typescript/package.json
```diff
@@ -49,10 +49,11 @@
     "audit:prod": "pnpm audit --prod --audit-level high",
     "clean": "node -e \"require('node:fs').rmSync('dist',{recursive:true,force:true})\"",
     "build": "node --run clean && tsc -p tsconfig.build.json && node scripts/build-dashboard.mjs",
+    "build:ci": "tsc -p tsconfig.ci.json",
```
mldangelo-oai, line 52: **[P3] Keep these CI utilities as checked-in JavaScript**

These three standalone CI utilities need to run directly, but `.mts` adds an emit-only `build:ci`, a second tsconfig, generated-file ignores, workflow/instruction compile steps, and a compile/symlink harness in the tests. Keep checked-in `.mjs` files, add `// @ts-check` and JSDoc where inference needs help, and include them in the existing strict no-emit config. That preserves type checking while deleting about 74 lines of build plumbing.

mldangelo-oai (commented):
The port remains behaviorally sound. These two non-blocking simplifications remove custom filesystem/build plumbing and keep the checks closer to native Node and shell behavior.

sdk/typescript/scripts/compare-test-reports.mts
kmbroai: I am retaining glob expansion to preserve the existing command contract. The original comparator advertises candidate files or glob patterns and expands them itself. I verified that passing an unexpanded candidate-?.xml argument directly, without a shell, succeeds in both the original Python helper and this TypeScript port. Shell expansion preserves this Ubuntu step but removes that accepted input behavior for direct process calls and shells that do not expand it. The migration keeps that compatibility, so this change is not applied.

sdk/typescript/package.json
kmbroai: This migration requires TypeScript as the authored source, compiled with the existing tsc, with the generated JavaScript executed on Node 22.13. Keeping .mts plus the emit step meets that requirement without experimental type stripping. Checked-in JavaScript with JSDoc would change the source-language requirement. The same build:ci step is also reused by the native TypeScript proofs later in the stack, so I am retaining this build path.

PR #773, opened 2026-09-01
fix(plugin): stream tracked diffs while hashing

### LINKED ISSUE #249: Content digests buffer entire git diff --binary patches in memory
## Summary

The content digest buffers a whole `git diff --binary` patch in memory before hashing it. A repository containing a large changed binary can therefore exhaust the workbench process during setup inspection, before any review begins.

Correction to an earlier version of this issue: I first wrote that *both* content-digest functions do this. On `main` there is only one `git diff --binary` call site, in `worktree_content_digest_for_context`. A committed base..head equivalent exists only in the unmerged PR #241, so if that lands it should adopt the same streaming helper.

## Environment

- macOS 15 (Darwin 25.5.0), Python 3.14.5

## Where it is

`sdk/typescript/_bundled_plugin/scripts/workbench_target.py` — `worktree_content_digest_for_context` obtains the tracked diff through `git_bytes`, which is `subprocess.run(..., capture_output=True)`, so `completed.stdout` holds the entire patch. The digest is then computed from that single value.

Binary patches are larger than the underlying files: measured on a 20 MiB incompressible file, `git diff --binary` produced a 27 MB patch, about 1.29x.

Measured peak RSS in the Python process on that fixture: **145.8 MiB**, against a 25.2 MiB floor for a clean worktree. Git's own peak was 165.6 MiB, which is inherent to producing the patch and not affected by this issue.

This runs on every setup inspection, so it is not confined to scan registration.

## Why this is not a one-line change

`update_digest_field` frames each value with an 8-byte big-endian length prefix, so the total byte count must be known before any content is hashed. Streaming therefore needs either a second full `git diff` pass to learn the length — non-atomic, and double the work — or spooling the patch to a temporary file and hashing from it.

Whichever shape is chosen, the resulting digest must stay byte-identical to today's, because recorded digests are compared against freshly computed ones when a saved selection is revalidated.

## Suggested direction

Add a streaming helper next to `git_bytes` that spools stdout to a temporary file, then feeds the framed length and the file contents into the digest. Apply it to both digest functions rather than only one, since the working-tree path dominates. A fixture asserting old and new produce the same digest for the same input would pin the compatibility requirement.

Hughhhhcoder:
## Summary

Avoid buffering the complete `git diff --binary` patch in the Python workbench while computing a working-tree content digest. Large changed binaries now use temporary-file storage instead of growing the workbench heap.

Fixes #249.

## Changes

- Allow the shared Git runner to direct stdout to a binary file while preserving captured stderr and existing failure behavior.
- Stream the tracked diff into an automatically removed temporary file, frame its exact byte length, and hash it in 1 MiB chunks.
- Add a binary-patch regression that verifies the streaming path and proves the resulting digest is byte-identical to the legacy buffered calculation.

## Testing

- `python -m pytest -q plugins/codex-security/tests`: 1079 passed, 5 skipped, 104 subtests passed.
- Portable Ruff check: passed.
- Portable Ruff format check: 137 files already formatted.
- `python .github/scripts/check_plugin_source_compatibility.py`: passed.
- `git diff --check`: passed.

## Risk and rollout

The `codex-security-snapshot/v1` framing and digest bytes are unchanged. The workbench trades peak Python heap usage for transient local temporary-file storage; the file is removed automatically when hashing finishes. Git process memory usage is unchanged. No CLI, configuration, schema, database, or network behavior changes.

## Public disclosure review

### DIFF
--- a/plugins/codex-security/scripts/workbench_target.py
+++ b/plugins/codex-security/scripts/workbench_target.py
@@ -10,8 +10,9 @@ import sqlite3
 import stat
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
-from typing import Any
+from typing import Any, BinaryIO
 
 # Some plugin hosts launch Python with safe-path isolation enabled.
 sys.path.insert(0, str(Path(__file__).resolve().parent))
@@ -127,6 +128,7 @@ def git_command(
     input_data: str | bytes | None = None,
     git_dir: Path | None = None,
     work_tree: Path | None = None,
+    stdout_file: BinaryIO | None = None,
 ) -> subprocess.CompletedProcess[str] | subprocess.CompletedProcess[bytes]:
     if (git_dir is None) != (work_tree is None):
         raise ValueError("git_dir and work_tree must be provided together")
@@ -148,15 +150,20 @@ def git_command(
         command.extend(["--git-dir", str(git_dir), "--work-tree", str(work_tree)])
     full_command = [*command, *args]
     try:
+        output_options = (
+            {"capture_output": True}
+            if stdout_file is None
+            else {"stdout": stdout_file, "stderr": subprocess.PIPE}
+        )
         return subprocess.run(
             full_command,
             check=False,
-            capture_output=True,
             env=environment,
             text=text,
             encoding="utf-8" if text else None,
             errors="surrogateescape" if text else None,
             input=input_data,
+            **output_options,
         )
     except FileNotFoundError:
         # Git is optional for Codebase scans. Treat an unavailable executable like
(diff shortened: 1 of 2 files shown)


Hughhhhcoder:
@codex review

mldangelo-oai (approved):
Thanks for the focused fix! I verified that the temporary-file path preserves the existing digest bytes, including a binary patch spanning multiple reads, CRLF text, Unicode paths, linked worktrees, and recovery copies. The Windows, Linux, and macOS test suites all passed. I will carry the larger regression fixture in a small follow-up.

PR #782, opened 2026-09-01
fix: restore batching for automatic finding matching

ianw-oai:
## Summary

Restore bounded batching for automatic post-scan history matching. The single-call restriction added in #638 can skip matching for a large history.

## Changes

- Remove the `allowBatching` switch and its two single-call tests.
- Keep the existing batch size limit, group matching, and cancellation behavior.
- Correct the README's one-call claim and state where cost tracking stops.

## Testing

- Focused matcher, component-scan, and API tests: 235 passed, 5 platform-specific tests skipped.
- `pnpm run types`, `pnpm run format`, and `pnpm run build` passed.
- `git diff --check` passed.

## Risk and rollout

Automatic history matching can make several model calls. These calls remain outside the scan's recorded cost and `maxCostUsd` limit. This PR restores batching; it does not add post-scan budget accounting. No CLI flags or result formats change.

## Public disclosure review

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -39,8 +39,9 @@ try {
 ```
 
 `result.findings` contains this scan's findings; `repositoryFindings` also
-includes earlier open findings when available. Matching earlier findings can
-make one extra model call, even with a cost limit.
+includes earlier open findings when available. Matching earlier findings uses
+additional model calls, with large inputs split into batches. These calls are
+outside the scan's recorded cost and `maxCostUsd` limit.
 
 Keep results outside the repository and restrict access: reports can contain
 source code, vulnerability details, and reproduction steps.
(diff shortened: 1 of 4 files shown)


ianw-oai:
Local validation for c269641 passed:

- Full suite in two orders (seeds 12345 and 2164852511): 2,208 passed, 43 skipped, no failures in each run.
- Focused matcher, component-scan, and API tests: 235 passed, 5 skipped.
- Type, format, build, and diff checks passed.

sylvesterkaczmarek (commented):
The batching path looks sound. Large comparisons are recursively split on one side, so every before/after pair is still covered once; each batch is validated before `combineComparisons`, then the combined result is validated again against the full input. Cancellation still happens between calls, and the README now makes the out-of-budget model calls explicit. I don't see a blocker here.

zcrab-oai (approved):

soyeon-oai (approved):

PR #784, opened 2026-09-02
fix(sdk): reject control characters in remote targets

### LINKED ISSUE #231: scan.target.remote accepts embedded tab/newline/CR because new URL() strips them before validating

Hughhhhcoder:
## Summary

The canonical contract validator accepts tab, line-feed, and carriage-return characters inside `scan.target.remote`. The WHATWG URL parser removes those characters before validation, while the original unsanitized value remains in the loaded manifest.

This can make a validated remote target contain embedded line or control characters and contradict the SDK contract that the value is sanitized.

Fixes #231.

## Changes

- Move the existing model-unsafe control and line-separator character pattern into a shared internal module.
- Reuse that pattern for both output-directory and canonical remote validation.
- Reject control and line-separator characters in `scan.target.remote` before URL parsing.
- Add regression coverage for tab, line-feed, and carriage-return remote values.

## Testing

- `bun test --timeout 30000 ./tests-ts/contract.test.ts -t "rejects schema-valid but canonically invalid contract data"`: 1 passed, 40 skipped, 15 expect calls
- `pnpm run check:plugin-source`: passed
- `pnpm run types`: passed
- `pnpm run lint`: passed
- `pnpm run format`: passed
- `git diff --check`: passed

The local Node runtime is 25.4.0 and emits the repository engine warning; the package requires Node 22, 24, or 26.

## Risk and rollout

Low and limited to invalid contract input. Valid URL handling, manifest schemas, output-directory behavior, and the public API shape remain unchanged. The shared pattern preserves the existing output-directory validation behavior.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/contract.ts
+++ b/sdk/typescript/src/contract.ts
@@ -19,6 +19,7 @@ import {
   requirePrivateOutputDirectory,
   requireSecureOutputAncestry,
 } from "./runtime.js";
+import { MODEL_UNSAFE_STRING } from "./string-safety.js";
 import type { NormalizedTarget, ScanMode } from "./targets.js";
 import { isWindowsUnsafePathComponent } from "./windows-path.js";
 
@@ -428,7 +429,11 @@ function validateCanonicalContract(
     const authority = /^[A-Za-z][A-Za-z0-9+.-]*:\/\/([^/?#]+)/.exec(
       remote,
     )?.[1];
-    if (remote.includes("\\") || authority === undefined) {
+    if (
+      MODEL_UNSAFE_STRING.test(remote) ||
+      remote.includes("\\") ||
+      authority === undefined
+    ) {
       throw new ContractValidationError(
         "scan.target.remote: expected a sanitized canonical absolute URL.",
       );
--- a/sdk/typescript/src/runtime.ts
+++ b/sdk/typescript/src/runtime.ts
@@ -63,6 +63,7 @@ import {
   isWindowsUnsafePathComponent,
   windowsUnsafePathComponent,
 } from "./windows-path.js";
+import { MODEL_UNSAFE_STRING } from "./string-safety.js";
 
 const execFile = promisify(execFileCallback);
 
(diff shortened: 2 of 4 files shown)


Hughhhhcoder:
@codex review

Hughhhhcoder:
Gentle follow-up on this focused fix for #231. The issue remains open, the affected validation path is still present on current main, and this PR is mergeable. Please let me know if any scope change or additional validation would help.

PR #788, opened 2026-09-02
fix(plugin): resolve Windows Node to an absolute executable

xinoai:
## Summary

The Windows MCP launcher should select one Node executable and launch it consistently. Resolve `node.exe` from `PATH` once, then invoke its absolute path.

## Changes

- Replace the separate `where` and bare `node` commands with the shell's built-in executable lookup.
- Keep the existing working-directory change and managed runtime priority.
- Add a native Windows regression test covering command extensions, caller-directory candidates, quoted paths, argument forwarding, exit codes, and a missing runtime.

## Testing

- `bun test --timeout 30000 tests-ts/mcp-launcher.test.ts`: passed the MCP smoke test on Linux; native Windows case skipped.
- `pnpm run types`: passed.
- `pnpm run format`: passed.
- All three required portable plugin source checks: passed.
- `pnpm run test --seed 12345` and `pnpm run test`: both passed on Linux (2,241 passed, 42 skipped in each run).
- [Native Windows regression and full CI test matrix](https://github.com/openai/codex-security/actions/runs/33635889874): passed. An unrelated subprocess test passed on retry.

## Risk and rollout

The PATH fallback requires a native `node.exe`. Installations using script-only PATH shims can use the existing `CODEX_MCP_NODE_PATH` override. No commands, flags, or public environment variables are added. The native Windows regression passed in CI.

## Public disclosure review


sylvesterkaczmarek (commented):
Checked the Windows fallback path. The managed-runtime branches still win first; the PATH fallback now resolves one native `node.exe` through cmd's `%~$ENV:I` lookup and invokes that exact executable instead of doing a second command search. The native regression also covers quoted PATH entries, caller-directory shims/candidates, argument forwarding, exit status and no-runtime behavior. Looks sound.

ianw-oai (approved):

PR #789, opened 2026-09-02
fix(cli): keep pre-commit hooks on external installations

xinoai:
## Summary

Generated pre-commit hooks need an installation that stays stable as the repository changes. Require the CLI installation and runtime to live outside the target worktree, and let users refresh previously generated hooks from that installation.

## Changes

- Resolve the Git worktree root and reject hook installation when the runtime, CLI entrypoint, or package root is inside it.
- Mark generated hooks and recognize the complete previous generated format for migration, including quoted paths.
- Preserve custom hooks and the existing severity threshold behavior.
- Update command help and setup guidance, with regression coverage for local npm installations, local runtimes, and hook refresh across a branch checkout.

## Testing

- `bun test --timeout 30000 tests-ts/cli.test.ts tests-ts/cli-launcher.test.ts`: 159 passed.
- `pnpm run types`: passed.
- `pnpm run format`: passed.
- `pnpm run test --seed 12345` and `pnpm run test`: both passed on Linux (2,243 passed, 41 skipped in each run).
- [Full CI matrix, including native Windows tests](https://github.com/openai/codex-security/actions/runs/33635219880): passed.

## Risk and rollout

`install-hook` now returns exit code 2 for repository-local installations. Install the CLI and runtime outside the repository, then rerun `install-hook` with the existing severity threshold to refresh an old generated hook. Package updates alone do not rewrite hooks. No commands, flags, or defaults are added.

## Public disclosure review

### DIFF
--- a/sdk/typescript/README.md
+++ b/sdk/typescript/README.md
@@ -893,7 +893,13 @@ reject `--json`.
 
 `install-hook` scans staged and unstaged changes before each commit. It blocks
 on high-severity findings or failed scans, respects `core.hooksPath`, and leaves
-existing hooks alone. Change the threshold with `--fail-on-severity`.
+custom hooks alone. Change the threshold with `--fail-on-severity`.
+The Codex Security installation and Node runtime must be outside the scanned
+repository. Install the CLI globally, then run `codex-security install-hook`
+from the repository. Repository-local installations cannot install a hook.
+To refresh an existing generated hook, rerun `install-hook` from the external
+installation with the same severity threshold. Updating the package alone does
+not update installed hooks.
 
 ### Import alerts from the CLI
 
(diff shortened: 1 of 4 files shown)


sylvesterkaczmarek (commented):
The marker-less migration check is broad enough to overwrite a custom hook. Since the managed comment is optional and both captured paths only need to be absolute, a hook like `exec '/abs/node' '/abs/custom-script' scan . --working-tree --fail-on-severity high` is treated as generated and replaced. That changes the previous “leave custom hooks alone” behavior. Can we require stronger provenance for the marker-less case, e.g. identify the old Codex Security CLI/package path before overwriting?

ianw-oai (approved):

PR #792, opened 2026-09-02
refactor(examples): port custom validation to TypeScript

kmbroai:
## Summary

Remove Python from the custom-validation demo. Compile the typed app and HTTP validator through the existing SDK build command, then execute the generated JavaScript on the supported Node.js runtime.

## Changes

- Replace the two Python example files with typed `.mts` sources and keep generated JavaScript untracked.
- Reuse `node --run build:examples -- --outDir <target>` in the demo runner and its test, with the SDK as the build working directory.
- Preserve the synthetic invoice ownership flaw, authentication controls, proof JSON, server cleanup, and compilation into the temporary scan target.
- Update validation prompts and documentation and include the example in strict type and formatting checks.

## Testing

- The focused generated HTTP-proof test passed with five assertions on Node 22.13.0.
- SDK/MCP type checks, SDK formatting, runner formatting, the documented SDK build, and runner help from outside the checkout passed.
- Default and custom output paths, including a path containing spaces, produced identical compiled files. Compiler diagnostics and the stop-before-scan failure guard were preserved.
- Earlier comparisons verified the original Python proof JSON and failure-path server cleanup. A live model-backed scan was not run.

## Risk and rollout

Stacked on #768. The demo uses existing SDK development dependencies, synthetic data, and an ephemeral loopback HTTP server. Build failures now use the Node script runner's exit status: the checked compilation failure returns 1 instead of the compiler's 2, with the same diagnostics and no scan start. This PR removes two Python files.

## Public disclosure review


mldangelo-oai (approved):
The compiled example preserves the authentication controls and proof output, and the HTTP validation and server cleanup test passes locally. No actionable findings in the example migration.

mldangelo-oai (commented):
The migration behavior remains sound; one non-blocking single-source simplification is inline.

examples/custom-validation/run.mjs
```diff
@@ -8,10 +8,31 @@ const root = await mkdtemp(join(tmpdir(), "codex-security-validation-demo-"));
 const target = join(root, "target");
 const output = join(root, "scan");
 await mkdir(target);
-for (const name of ["app.py", "validate.py"]) {
+for (const name of ["app.mts", "validate.mts"]) {
   await copyFile(new URL(name, import.meta.url), join(target, name));
 }
 
+const build = spawnSync(
```
mldangelo-oai, line 15: **[P3] Reuse `build:examples` instead of spelling out tsc**

`run.mjs` and the test both reconstruct the TypeScript executable and `tsconfig.examples.json`, even though `package.json` already defines `build:examples`. Spawn `node --run build:examples -- --outDir <target>` with `cwd` set to `sdk/typescript`, and do the same in the test. This leaves one canonical build command and removes roughly 16 lines.

kmbroai: Updated in dc64739f. The runner and focused test now invoke `node --run build:examples -- --outDir <target>` from `sdk/typescript`, so the package script owns the compiler configuration. The focused test, types, formatting, and runner smoke passed on Node 22.13; default and custom output paths produce identical bytes. `node --run` preserves compiler diagnostics and the stop-before-scan guard, while returning status 1 for a compiler failure.

PR #794, opened 2026-09-03
refactor(plugin): add Python-free Unix OS primitives

kmbroai:
## Summary

Add the Unix OS primitives needed to replace Python filesystem, locking, and account-lookup helpers with TypeScript callers.

## Changes

- Add nine Node-API 8 bindings for descriptor-relative operations, metadata, raw link targets, locking, and named-user home lookup. TypeScript owns caller behavior.
- Compile the typed build and proof tools through the existing TypeScript build. Build with Cargo, without Python or node-gyp.
- Add Linux and macOS x64/arm64 artifact jobs with Node 20.0 and 22.13 proofs, compatibility-floor checks, and build-path remapping. Use a relative macOS install name so linker metadata does not embed the build directory.

## Testing

- On the refreshed head: frozen SDK/MCP installation, CI-tool compilation, Ruff, portable source checks, and all nine checker tests passed. Earlier scoped SDK/MCP type, format, and Rust checks passed; the implementation patches are unchanged.
- Fresh Linux builds and native behavioral proofs passed on Node 20.0 and 22.13, including Python lock interoperability during independent review.
- The original implementation is unchanged by the rebase. Earlier platform and full SDK validation is recorded in the PR history; fresh hosted jobs will verify this head.
- Local workstation builds do not establish the glibc 2.28 compatibility floor. The hosted manylinux and macOS jobs remain required.

## Risk and rollout

Stacked on #792. This foundation is opt-in preparation. Windows bindings and universal package assembly follow before product commands use the native payloads. Hosted macOS and arm64 execution remain required validation.

## Public disclosure review

- [x] No customer, partner, prospect, or user identities, data, or identifying details are included.
- [x] No credentials, personal data, private source, scan findings, or nonpublic links or tickets are included.
- [x] I reviewed the branch name, title, description, commits, changes, comments, logs, screenshots, attachments, and links for public disclosure.

PR #795, opened 2026-09-03
refactor(plugin): add Python-free Windows OS primitives

kmbroai:
## Summary

Add the Windows OS primitives needed for TypeScript to replace the Python filesystem and locking helpers.

## Changes

- Use Rust `File` to own Windows handles and provide ordinary I/O, cursor-preserving truncation, flush, and exclusive whole-file locks. Keep Win32 metadata, identity, paths, and exact-handle rename/deletion. Preserve raw UTF-16 paths and numeric Windows errors, including closed-handle error 6 and contention error 33.
- Split the native crate into Unix and Windows modules, retaining the existing Unix implementation.
- Add x64 and arm64 Windows artifact jobs with Node 20.0 and 22.13 proofs for directory sharing, junctions, handle lifetime, and cross-process locks.

## Testing

- Both x64 and arm64 MSVC-target Clippy checks, Rust formatting, CI-tool compilation, proof syntax, Prettier, actionlint, Ruff, portable source checks, and all nine checker tests passed locally.
- Windows proofs cover I/O errors, cursor/EOF behavior, handle lifetime, whole-file contention, and unlock/close/process-death release. A separate Node 22 invocation compares both directions against the existing Python completion-lock functions.
- Actual x64/arm64 Windows execution on Node 20.0 and 22.13, including the optional Python comparison, remains a hosted CI gate for this head.

## Risk and rollout

Stacked on #794. This is opt-in preparation; product commands do not use the new bindings yet. Universal package assembly and successful native Windows proofs are required before the TypeScript callers replace Python.

## Public disclosure review


mldangelo-oai:
Reviewed [90e2987](https://github.com/openai/codex-security/commit/90e298788d60baf6cf64865d720dc538b194ac9f). The native Windows x64 and arm64 proofs pass, and I found no actionable binding issues.

Holding approval for the [Windows package verification timeout](https://github.com/openai/codex-security/actions/runs/33700983259/job/100480236694). The log reports successful installed-package validation before the outer smoke process hits its 210-second timeout. Please rerun this job on the reviewed head. If it repeats, investigate smoke-process exit and cleanup; the log alone does not establish a native-binding regression.

mldangelo-oai (approved):
The [Windows package verification rerun](https://github.com/openai/codex-security/actions/runs/33700983259/job/100485082131) passed on the unchanged reviewed head, clearing my earlier hold. Both native Windows x64 and arm64 proofs pass. No actionable findings in the binding changes.

plugins/codex-security/native/src/windows.rs
```diff
@@ -0,0 +1,416 @@
...
+fn overlapped(offset: u64) -> OVERLAPPED {
+    let mut value = OVERLAPPED::default();
+    value.Anonymous.Anonymous.Offset = offset as u32;
+    value.Anonymous.Anonymous.OffsetHigh = (offset >> 32) as u32;
+    value
+}
+
+#[napi]
+pub struct WindowsHandle {
+    handle: Option<OwnedHandle>,
```
mldangelo-oai: **[P3] Use `File` for the safe half of this handle API**

`WindowsHandle` owns a raw handle, then reimplements ordinary I/O and generalized byte-range/shared locking through Win32 FFI. This crate pins Rust 1.97, so `std::fs::File` can own the handle and provide safe read/write/seek/flush/truncate plus whole-file exclusive lock/try-lock/unlock; no production caller needs arbitrary ranges or shared locks. That removes about eight of this file’s 24 unsafe sites without changing the Win32-specific metadata, identity, path, rename, and delete operations. Preserve the numeric error contract—including mapping `WouldBlock` to error 33—and keep cursor/EOF plus two-way Rust↔Python contention and release tests; retain a raw primitive only where a focused test shows the standard API changes a required contract.

mldangelo-oai (commented):
The exact-head Windows proofs are green; this is a non-blocking simplification follow-up.

plugins/codex-security/native/src/windows.rs
kmbroai: Updated in 7eaffc5c. `File` now owns the handle and provides read/write/seek, cursor-preserving EOF, `sync_all`, and exclusive whole-file lock/try-lock/unlock; the Win32-specific metadata/path/identity/rename/delete calls remain. Closed handles still return 6, `WouldBlock` maps to 33, and close is idempotent. The proofs now cover cursor/EOF and I/O errors plus two-way contention and unlock/close/process-death release against the existing Python completion-lock functions. Both MSVC targets passed Clippy and the portable source/type/format checks passed; Windows execution is covered by the hosted jobs.

PR #796, opened 2026-09-03
build(plugin): add Python-free musl native artifacts

kmbroai:
## Summary

Prepare native OS primitives for Linux installations that use musl, so the TypeScript migration can preserve Alpine support.

## Changes

- Distinguish GNU and musl Linux payloads using the Node diagnostic report header.
- Add native x64 and arm64 musl builds with pinned Rust and Node container images, offline compilation, and no Python dependency.
- Fetch locked dependencies with the minimal pinned Rust toolchain; leave formatting, linting, and compilation to their existing jobs.
- Check ELF architecture and libc requirements, then test each unchanged binary on Node 20.0 and 22.13.

## Testing

- Locked Cargo dependency fetch with `cargo +1.97.1 fetch --locked`, workflow actionlint, and formatting passed.
- Three independent reviews and verification passed. The existing native behavioral and source checks passed locally.
- Hosted x64/arm64 musl builds and Node 20/22 runtime proofs validate the pinned container artifacts.

## Risk and rollout

Stacked on #795. This prepares build artifacts; product commands continue using their existing helpers. Universal package assembly and platform proofs remain required before TypeScript callers replace Python.

## Public disclosure review

### DIFF
--- a/.github/workflows/native-musl.yml
+++ b/.github/workflows/native-musl.yml
@@ -1,0 +1,112 @@
+name: native-musl
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  primitives:
+    name: primitives / linux-${{ matrix.arch }}-musl
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runner: ubuntu-24.04
+            arch: x64
+            rust-image: rust@sha256:83b16fd204613a557527e81ef4f87fd513229187c49635e3c7f96b08dfa43c33
+            node22-image: node@sha256:133cdce957f50f47236d6d926592fb1db7a120ac3c33191e611b60dfab63e324
+            node20-image: node@sha256:2ffec31a58e85fbcd575c544a3584f6f4d128779e6b856153a04366b8dd01bb0
+          - runner: ubuntu-24.04-arm
+            arch: arm64
+            rust-image: rust@sha256:ec97f172f73f5ec381a238dad6ddd99e0e9c14fd3709a05beb41ccdf50eb666a
+            node22-image: node@sha256:9090fa64b73b14ab658d907e7916afcb1633e4bdd5134f9b1be4a21a6c56f0b8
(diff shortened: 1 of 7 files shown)


mldangelo-oai:
I see the correction in [12303d2](https://github.com/openai/codex-security/commit/12303d2b01dac954d0e16cfe833447f24ea309ca). The validator now attributes version requirements to their providing libraries, including the `libgcc_s.so.1` compatibility case.

The [current hosted arm64 musl job](https://github.com/openai/codex-security/actions/runs/33702401535/job/100484284760) is still running. Holding approval until this head passes the ELF check, runs the same artifact through the Node 20 and 22 proofs, and uploads the verified artifact. The previous head stopped before the runtime proofs and upload; the new native-runner result will verify the fix.

mldangelo-oai (approved):
Reviewed the validator correction in 12303d2. The [hosted arm64 musl job](https://github.com/openai/codex-security/actions/runs/33702401535/job/100484284760) now passes ELF validation, runs the same artifact successfully on Node 20.0.0 and 22.13.0, and uploads the verified artifact. Both musl architectures and the current SDK CI run pass on this head. This clears my earlier approval hold.

.github/workflows/native-musl.yml
```diff
@@ -0,0 +1,112 @@
...
+        with:
+          node-version: "22.13.0"
+      - name: Compile TypeScript tools
+        working-directory: .
+        run: |
+          pnpm --dir sdk/typescript install --frozen-lockfile
+          pnpm --dir sdk/typescript run build:ci
+      - name: Fetch locked Rust dependencies
+        run: |
+          rustup toolchain install 1.97.1 --profile minimal --component rustfmt --component clippy
```
mldangelo-oai: **[P3] Install only the toolchain this job uses**

This step only runs Cargo to populate the registry; formatting and Clippy run elsewhere, and the actual compilation happens in the pinned container. Could we install `1.97.1 --profile minimal` without `rustfmt`/`clippy`, then run `cargo +1.97.1 fetch --locked`? The explicit `+` keeps `rust-toolchain.toml` from adding those components back. The current job downloads five components on both architectures when it needs three.

kmbroai: Updated the dependency-fetch step to install only `1.97.1 --profile minimal` and invoke `cargo +1.97.1 fetch --locked`. The locked fetch, workflow lint, and formatting pass locally. Updated in `50ac3573`.

mldangelo-oai (approved):
Re-reviewed exact head 50ac3573. The musl dependency fetch now installs only pinned Rust 1.97.1 with the minimal profile and uses cargo +1.97.1 fetch --locked. The prior thread is resolved, the stack base is exact, and all required Linux, macOS, and Windows checks pass.

PR #797, opened 2026-09-03
build(plugin): bundle verified native runtime artifacts

kmbroai:
## Summary

Include verified native OS primitives in the standalone plugin and npm package so TypeScript helpers can run without compiling or downloading native code at runtime.

## Changes

- Assemble all eight Linux, macOS, and Windows payloads once in PR Node CI and share the artifact with reusable container and test-quality jobs. Standalone and tag-release runs assemble their own artifact.
- Preserve required Node check names and release permissions. Serialize container releases with one concurrency group and retain the repository authorization guard. Remove the unused native download from the MCP test job.
- Copy the complete native tree and dependency notices from the single inventory in `plugin-files.json`. Missing declared inputs fail the build; undeclared files are excluded.
- Check installed native loading with an empty PATH and link artifact preparation before both README source-build recipes.
- Use asynchronous subprocesses in two runner tests while preserving their assertions and deadlines.

## Testing

- Workflow actionlint and formatting passed.
- Independent reviews exercised 49 focused package/workflow tests and nine source-contract tests. An isolated MCP build matched the manifest inventory and loaded its Linux native payload with an empty PATH.
- Prior source-build recipe and installed-package smoke results remain recorded in the PR history; hosted CI validates the refreshed universal artifact and downstream platform jobs.

## Risk and rollout

Stacked on #796. Adds about 3.8 MB of native binaries and notices before compression. Source builds require the documented prebuilt inputs. Hosted CI covers universal artifact assembly and downstream platform jobs. Product helpers switch to TypeScript in subsequent changes.

## Public disclosure review


plugins/codex-security/mcp-app/scripts/build_mcp_app.mjs
```diff
@@ -20,6 +20,33 @@ export async function buildMcpApp({ output }) {
   await mkdir(mcpDir, { recursive: true });
 
   await writeRuntime("server", "main.ts");
+  for (const target of [
```
mldangelo-oai: Could we make `plugin-files.json` the single native-file inventory instead of maintaining these target and notice arrays here too? This block repeats all eight target/name pairs and all six notice paths, including the Windows-versus-Unix filename rule. The package path already exact-compares the generated tree with `plugin-files.json`, and the direct-builder and tarball checks repeat that contract.

A manifest-driven copy would preserve immediate failure for a missing declared input and continue excluding undeclared files while removing roughly 15–18 production lines. A recursive `fs.promises.cp(native/prebuilt, mcp/native, { recursive: true })` is shorter still—about 22 lines deleted—but then the standalone builder needs a small exactness check because `cp` alone will accept a partial tree and preserve extras until the downstream package check. Either shape leaves one inventory rather than two that can drift.

plugins/codex-security/mcp-app/scripts/build_mcp_app.mjs
```diff
@@ -20,6 +20,33 @@ export async function buildMcpApp({ output }) {
...
+    "linux-x64-gnu",
+    "linux-x64-musl",
+    "win32-arm64",
+    "win32-x64"
+  ]) {
+    const name = target.startsWith("win32-") ? "windows.node" : "unix.node";
+    const destination = join(mcpDir, "native", target);
+    await mkdir(destination, { recursive: true });
+    await copyFile(
+      join(root, "../native/prebuilt", target, name),
```
mldangelo-oai: This copy is now a mandatory input to every `build:plugin`/`prepack`, but the public source-build instructions do not prepare it. In a clean checkout, `native/prebuilt` is ignored and absent, so the “From a source checkout” sequence in `sdk/typescript/README.md` reaches this line and fails with `ENOENT`; the README’s Docker source-build recipe reaches the same failure through `pnpm pack`. The narrower testing and Docker guides were updated, but these two public recipes were not.

Please add the same universal-artifact preparation step or an explicit link before both README recipes, and validate each sequence from a fresh source snapshot with no `native/prebuilt` directory. This is introduced by the unconditional copy, not an existing setup requirement.

plugins/codex-security/mcp-app/scripts/build_mcp_app.mjs
kmbroai: Fixed in bd11e36b. The builder now selects native inputs from plugin-files.json. A direct build matches the manifest exactly, fails when a declared payload is missing, and excludes undeclared inputs.

plugins/codex-security/mcp-app/scripts/build_mcp_app.mjs
kmbroai: Added the artifact-preparation link before both README source-build recipes in bd11e36b. Verified them using separate source snapshots that initially had no native/prebuilt directory: after downloading the universal artifact, both the SDK and Docker builds completed and their findings APIs returned HTTP 200.

.github/workflows/container-ci.yml
```diff
@@ -38,7 +42,11 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
+  native:
+    uses: ./.github/workflows/native-artifacts.yml
```
mldangelo-oai, line 38: **[P2] Build the native matrix once per PR**

`node-ci`, `container-ci`, and `container-release` are separate top-level PR workflow runs, and each invokes `native-artifacts`; `test-quality` adds a fourth copy when its path filter matches. Because artifacts are scoped to one run, those callers cannot reuse the native artifact produced by `node-ci`. At this head the four runs each executed eight platform builds plus a bundle job: 36 jobs and 67m18s of runner time. A normal plugin/SDK/container-path PR still pays for the first three copies: 27 jobs and 49m32s. Please make one PR-level workflow own `native-artifacts` and invoke the consumer validations as reusable jobs within that run. Tag-triggered releases can remain self-contained so release provenance stays tied to the tag SHA.

.github/workflows/node-ci.yml
```diff
@@ -319,6 +328,8 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+      - name: Prepare native runtime
```
mldangelo-oai: **[P3] Drop the native download from the MCP test job**

This job restores the already-packaged `_bundled_plugin` and runs the MCP package's direct `node --test` script. It never invokes `build_mcp_app.mjs`—the only code here that reads `native/prebuilt`—and the MCP tests reference the restored `sdk/typescript/_bundled_plugin`. Removing this step saves the redundant artifact download and two YAML lines without reducing coverage.

sdk/typescript/scripts/smoke-package.mjs
```diff
@@ -394,6 +394,39 @@ try {
...
+    "native",
+    `${process.platform}-${process.arch}${libc}`,
+    process.platform === "win32" ? "windows.node" : "unix.node",
+  );
+  run(
+    process.execPath,
+    [
+      "--input-type=commonjs",
+      "--eval",
+      `const assert = require("node:assert/strict");
```
mldangelo-oai: **[P3] Keep this smoke at the add-on load boundary**

The distinct package-level check here is platform selection, tarball assembly, ABI compatibility, and Node-API loading under an empty `PATH`. The platform proofs already exercise `openWindowsFile` and `duplicate` before these artifacts are uploaded. Replacing this eval body with `require(process.argv[1])` and removing the extra fixture argument preserves the installed-package signal while deleting about eight lines of duplicate syscall assertions.

mldangelo-oai (commented):
Re-reviewed the exact current head after the stacked fixes. The packaged-artifact behavior looks sound and the earlier manifest/docs comments are resolved. I found three concrete CI/test simplifications below.

.github/workflows/container-ci.yml
kmbroai: Confirmed. Updated in 0e3c1d97: Node CI owns the PR native matrix and invokes container and test-quality validation as reusable jobs within the same run. Existing path selection and required Node check names are preserved. Standalone runs and tag releases remain self-contained. Workflow validation covers routing, native dependency propagation, and unchanged release permissions.

.github/workflows/node-ci.yml
kmbroai: Confirmed. Removed the unused native download in 0e3c1d97; this MCP job continues to test the restored bundled plugin.

sdk/typescript/scripts/smoke-package.mjs
kmbroai: Confirmed. Simplified the installed-package smoke in 0e3c1d97 to require the selected add-on with an empty PATH and removed the duplicate syscall assertions. Normal pack/prepack and fresh installed-package smoke passed on Node 22.13.0.

.github/workflows/container-release.yml
```diff
@@ -27,215 +14,14 @@ permissions:
...
-        env:
-          IMAGE: codex-security:release-candidate
-        run: bun sdk/typescript/scripts/smoke-findings-service.ts "$IMAGE"
+  validate:
+    needs: native
+    if: github.repository == 'openai/codex-security'
+    uses: ./.github/workflows/container-validate.yml
 
   authorize:
     if: github.repository == 'openai/codex-security' && github.event_name != 'pull_request'
```
mldangelo-oai: **[P3] Remove the dead PR branches**

This workflow no longer has a `pull_request` trigger (and it does not expose `workflow_call`), so this ternary always selects `container-release-publish` and the `authorize` check at line 27 is always true on the event side. Could we collapse this to `concurrency: container-release-publish` and make `authorize`'s `if` just the repository guard? That preserves release serialization and fork protection while removing two dead YAML branches.

kmbroai: Simplified the workflow to `concurrency: container-release-publish` and kept only the repository guard on `authorize`. Workflow lint and formatting pass; release serialization and the repository restriction remain in place. Updated in `2aeff7a6`.

mldangelo-oai (approved):
Re-reviewed exact head 2aeff7a6. The release workflow now uses one publication concurrency group and the repository-only authorization guard, removing the dead pull-request branches. All prior threads are resolved, the stack base is exact, and all required checks pass.

PR #798, opened 2026-09-03
refactor(plugin): preserve Windows filenames in native helpers

kmbroai:
## Summary

Preserve Windows filenames when replacing Python path handling with TypeScript. Helpers receive original UTF-16 arguments, environment values, directory names, and paths, including lone surrogates.

## Changes

- Use `std::fs::read_dir` and cached entry types for one typed directory enumerator, preserving raw names and numeric Windows errors.
- Delegate absolute paths to Rust `std::path::absolute`, preserving native errors and the existing empty-input result. Normalize verbatim paths below their drive or share root, including doubled separators and parent traversal.
- Use the existing handles for metadata, reads, and writes, and Rust `create_dir_all` for recursive directory creation. Retain raw process-value transport.
- Clear the Windows error state before resolving empty input; retain the compatibility branch because the public Rust path API rejects empty paths before reaching Win32.
- Use Rust-created surrogate fixtures to test the adapter. A held file-sharing guard verifies enumeration succeeds while a data read fails.

## Testing

- Both MSVC-target Clippy checks, Rust formatting, CI-tool compilation, TypeScript formatting, Ruff, portable source checks, and all nine checker tests passed.
- Twelve adapter normalization cases passed on Node 22.13.0, including doubled separators under drive and UNC roots.
- Windows CI runs the actual filesystem and process proofs on x64 and arm64 with Node 20 and 22. The proofs cover recursive creation, existing-file errors, and empty-input error-state independence.

## Risk and rollout

Stacked on #797. Adds native operations without changing command routing or public CLI syntax. Junction directory entries now carry Rust's directory and symbolic-link flags; the following resolver still excludes their traversal. Sharing restrictions do not block attribute-only requests, so the proof checks a data read without changing product stat access.

## Public disclosure review


plugins/codex-security/native/windows-files.mts
```diff
@@ -0,0 +1,236 @@
...
+      const result = handle.finalPath(0);
+      check(result.error, path);
+      return result.path;
+    } finally {
+      check(handle.close(), path);
+    }
+  }
+
+  function realpath(path: Buffer): Buffer {
+    // Match CPython's Windows normalization of UNC and device prefixes.
```
mldangelo-oai: This is exactly the kind of CPython filesystem behavior we should avoid reimplementing. On this head the custom normalizer already has two deterministic root-sensitive bugs:

- `\\?\UNC\server\share\..\other\file` is passed to `openWindowsFile` as `\\?\UNC\server\other\file`, dropping the share root.
- `C:.\..\sentinel` becomes `C:sentinel`; both `path.win32.normalize` and `ntpath.normpath` preserve it as `C:..\sentinel`.

For ordinary paths, `absolute()` already delegates drive-relative/UNC resolution to `GetFullPathNameW`, and `finalPath()` delegates canonicalization to `GetFinalPathNameByHandleW`. Please collapse the ordinary path to those native operations plus, if the proof requires it, a small root-aware trailing-separator trim. Explicit `\\?\` input is the one caveat: Windows disables normal dot-segment parsing under that prefix, so keep only a small, table-tested verbatim-path normalizer if Python-compatible verbatim spellings are intentionally in scope.

Please also prove that this filesystem adapter is needed on the actual minimum runtime. Node 22.13’s libuv uses WTF-8 for Windows filesystem input, `scandir`, and `realpath`, including an [upstream unpaired-surrogate round trip](https://github.com/libuv/libuv/pull/2970); Node 20.0 is the older gap. Run this PR’s Rust-created surrogate fixtures directly against `node:fs` on the supported floor. If 20.0 remains a hard contract, isolate that gap in the native boundary; otherwise use Node core rather than maintaining a second Windows path implementation. At minimum this removes roughly 10–25 production lines and eliminates both current bugs.

plugins/codex-security/native/src/windows.rs
```diff
@@ -38,6 +46,197 @@ fn wide_path(bytes: Buffer) -> napi::Result<Vec<u16>> {
...
+pub fn windows_directory_names(path: Buffer) -> napi::Result<DirectoryResult> {
+    let result = windows_directory_entries(path)?;
+    Ok(DirectoryResult {
+        error: result.error,
+        value: result.value.into_iter().map(|entry| entry.name).collect(),
+    })
+}
+
+#[napi]
+pub fn windows_directory_entries(path: Buffer) -> napi::Result<DirectoryEntriesResult> {
```
mldangelo-oai, line 146: Can we delete this custom Win32 directory search and use `std::fs::read_dir(os_string(path)?)`? The pinned Rust implementation already joins `*`, handles empty/missing/non-directory cases, uses the shared long/verbatim conversion for relative, drive-relative, UNC, and already-verbatim paths, and preserves lone UTF-16 surrogates through `OsString`. [`DirEntry::file_type()`](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.file_type) is documented as cached/no-extra-syscall on Windows and does not follow links, so it also preserves the important “do not open every child” property.

Mapping construction and iteration failures through `io::Error::raw_os_error()` retains the numeric-error contract. This removes `directory_search_path`, the unsafe `FindFirstFileW`/`FindNextFileW` loop, and roughly 65–75 net Rust lines.

There is one real semantic decision to make explicit: stable Rust classifies junction/mount-point name-surrogate reparse points as `is_symlink_dir`, whereas this adapter currently documents junctions as directories but not symlinks. The policy resolver suppresses traversal either way. Unless that exact label is independently required, I would use the standard-library semantics and update the proof/README; if it is required, document it as the narrow reason custom enumeration must remain.

plugins/codex-security/native/src/windows.rs
```diff
@@ -38,6 +46,197 @@ fn wide_path(bytes: Buffer) -> napi::Result<Vec<u16>> {
...
+                error: 0,
+                value: wide_bytes(absolute[..length as usize].iter().copied()),
+            });
+        }
+        absolute.resize(length as usize + 1, 0);
+    }
+}
+
+#[napi]
+pub fn windows_directory_names(path: Buffer) -> napi::Result<DirectoryResult> {
```
mldangelo-oai: `windowsDirectoryNames` is duplicate surface: it immediately calls `windowsDirectoryEntries` and discards the two cached type booleans. That duplication then spreads into a second N-API result type, the TypeScript binding, `windowsFileSystem.entries`, README wording, and parallel proof assertions.

Please keep one typed directory enumerator and map `.name` only at a caller that truly needs names. On this head the names-only route is proof-only, while the stacked resolver uses `entriesWithTypes`. Removing it preserves error, order, and raw UTF-16 behavior, avoids an extra projection/allocation, and cuts about 30 net lines with no shipped CLI or SDK compatibility impact.

plugins/codex-security/native/examples/windows-wide-launcher.rs
```diff
@@ -0,0 +1,178 @@
...
+        process::Command,
+        ptr::null_mut,
+    };
+    use windows_sys::Win32::Security::*;
+
+    fn raw(prefix: &str, unit: u16) -> OsString {
+        OsString::from_wide(&prefix.encode_utf16().chain([unit]).collect::<Vec<_>>())
+    }
+
+    fn deny_file_access(path: &Path, operation: impl FnOnce() -> io::Result<()>) -> io::Result<()> {
```
mldangelo-oai: This is too much unsafe ACL machinery for a proof fixture. The property we need is only “directory enumeration sees an entry without opening that child.” Rust exposes a much smaller setup through [`OpenOptionsExt::share_mode(0)`](https://doc.rust-lang.org/std/os/windows/fs/trait.OpenOptionsExt.html#tymethod.share_mode).

Open the fixture file with sharing disabled, keep that guard alive while the Node child runs, and assert that enumeration still reports it while an explicit `stat` fails with a sharing violation. Let the guard drop before cleanup. That proves the same cached-enumeration behavior and deletes roughly 65 lines of security-descriptor allocation, DACL replacement, protection-state restoration, and unsafe Win32 calls. The exact fixture error changing from access denied to sharing violation is not product behavior.

plugins/codex-security/native/windows-files.mts
kmbroai: Fixed both root cases in 08a2ac57. Ordinary paths now use GetFullPathNameW and GetFinalPathNameByHandleW with root-aware separator trimming. Only explicit verbatim inputs retain dot-segment normalization, covered by the drive/UNC tables.

The direct Rust-fixture comparison confirms your runtime point: Node 22.13.0 WTF-8 Buffer reads, directory names, and realpath preserve the surrogate; Node 20.0.0 fails those comparisons. JavaScript string paths lose it on both. Node 22.13.0 is the SDK floor; 20.0.0 is additional foundation coverage. [Windows x64 results](https://github.com/openai/codex-security/actions/runs/33926760752/job/101196934966).

The remaining native requirement is the resolver's generic reparse-point metadata check, which Node's Stats API does not expose, plus raw absolute resolution and Python-compatible verbatim spelling. I retained the existing UTF-16 boundary for the other I/O rather than introducing a second WTF-8 representation alongside it. That is a reuse choice, not a claim that Node 22 Buffer I/O is incapable of these operations.

plugins/codex-security/native/src/windows.rs
kmbroai: Changed to std::fs::read_dir and cached DirEntry::file_type() in 08a2ac57; removed the custom search-path builder and unsafe FindFirstFileW/FindNextFileW loop. Construction and iteration failures retain raw Windows error numbers. Junctions now carry both directory and symlink flags, documented and covered by the proof; the stacked resolver still suppresses their traversal. The Windows x64 proof passes on Node 22.13.0 and 20.0.0.

plugins/codex-security/native/src/windows.rs
kmbroai: Removed the names-only export, result type, binding, adapter method, and parallel proof route in 08a2ac57. The single typed enumerator preserves names, ordering, numeric errors, and raw UTF-16, and the stacked resolver uses entriesWithTypes.

plugins/codex-security/native/examples/windows-wide-launcher.rs
kmbroai: Replaced the ACL fixture with an OpenOptions share_mode(0) guard in 08a2ac57. One correction to the proposed assertion: Windows sharing restrictions do not block attribute-only stat access, so the proof asserts an explicit data read fails with sharing error 32 while enumeration succeeds. The guard remains alive during the Node child and drops before cleanup. This passes in the Windows x64 proof on both Node versions.

plugins/codex-security/native/windows-files.mts
```diff
@@ -0,0 +1,219 @@
...
+    if (pathText(path).startsWith("\\\\?\\")) {
+      // Verbatim paths bypass Win32 dot parsing; normalize only below their root.
+      const text = pathText(path).replaceAll("/", "\\");
+      const root =
+        /^\\\\\?\\(?:UNC\\[^\\]+\\[^\\]+(?:\\|$)|[^\\]+\\)/iu.exec(text)?.[0] ??
+        win32.parse(text).root;
+      normalizedText =
+        root +
+        win32
+          .normalize(`\\${text.slice(root.length)}`)
```
mldangelo-oai: **[P2] Avoid UNC parsing after a repeated root separator**

The root match already consumes one separator, so an input such as `\\?\C:\\..\file` leaves another here and makes `win32.normalize()` parse `\\..\file` as UNC. It consequently preserves the dot component and passes `\\?\C:\\..\file` to `CreateFileW`; verbatim paths do not expand `..`, so `realpath` fails or targets a literal spelling instead of the canonical file. `win32.join("\\", text.slice(root.length))` provides the intended rooted normalization without inventing UNC. Please add doubled-separator drive and UNC regression cases.

plugins/codex-security/native/src/windows.rs
```diff
@@ -38,6 +44,102 @@ fn wide_path(bytes: Buffer) -> napi::Result<Vec<u16>> {
...
+
+#[napi]
+pub fn windows_environment(name: Buffer) -> napi::Result<Option<Buffer>> {
+    Ok(std::env::var_os(os_string(name)?).map(|value| wide_bytes(value.encode_wide())))
+}
+
+#[napi]
+pub fn windows_absolute_path(path: Buffer) -> napi::Result<BufferResult> {
+    let path = wide_path(path)?;
+    let mut absolute = vec![0_u16; 256];
```
mldangelo-oai: **[P3] Let Rust own `GetFullPathNameW`**

Pinned Rust 1.97 already provides `std::path::absolute`. On Windows it calls `GetFullPathNameW` for non-verbatim paths, preserves verbatim paths, and `OsString` round-trips ill-formed UTF-16. Replacing this 23-line unsafe buffer-growth loop with decode → `std::path::absolute` → encode removes about nine Rust lines and lets the standard library own `LastError` and buffer sizing; the TypeScript verbatim short-circuit then becomes redundant too. Preserve `raw_os_error()` when present and handle the empty-input `InvalidInput` case explicitly.

plugins/codex-security/native/proof-windows-wide.mts
```diff
@@ -0,0 +1,332 @@
...
+      name.equals(rawName),
+    ),
+  );
+  const coreBufferRealpath = coreSupports(() =>
+    realpathSync.native(rawCwd, { encoding: "buffer" }).equals(rawCwd),
+  );
+  return {
+    nodeStringPathsReadReplacementSibling: true,
+    nodeStringNamesAndRealpathLoseSurrogates: true,
+    nodeWtf8BufferReads: coreBufferRead,
```
mldangelo-oai: **[P3] Remove the observational Node-core probes**

These three results are only copied into the printed JSON; no workflow assertion or consumer reads them. The exact-head Node 20 jobs demonstrate that by passing while all three are `false`. Since product behavior does not depend on Node's Buffer-path support, this is one-time characterization rather than a durable regression contract. Please delete the roughly 42-line probe block and its README description instead of retaining non-enforcing test output. If the matrix is intended as a permanent contract, assert the pinned Node 22.13 `true/true/true` and Node 20 `false/false/false` results instead.

mldangelo-oai (commented):
Re-reviewed the exact current head after the latest fixes. The four earlier threads are resolved. I found one remaining Windows path correctness bug and two nonblocking simplifications.

plugins/codex-security/native/windows-files.mts
kmbroai: Confirmed and fixed in c58c6d0f with rooted win32.join normalization. Added doubled-separator drive and UNC regression cases; all 12 portable path cases pass on Node 20 and 22.

plugins/codex-security/native/src/windows.rs
kmbroai: Confirmed. Switched non-empty paths to std::path::absolute in c58c6d0f, preserved raw_os_error(), and removed the TypeScript verbatim bypass. Empty input retains a single GetFullPathNameW call to preserve its native error behavior without the buffer-growth loop. Both Windows targets pass the CI Clippy scope; the runtime proof also covers the empty result.

plugins/codex-security/native/proof-windows-wide.mts
kmbroai: Confirmed. Removed the observational Node-core probes, their unused fixture and output fields, and the README description in c58c6d0f. The raw UTF-16 filesystem, process, and sharing assertions remain.

plugins/codex-security/native/src/windows.rs
```diff
@@ -63,6 +68,106 @@ fn wide_path(bytes: Buffer) -> napi::Result<Vec<u16>> {
...
+
+#[napi]
+pub fn windows_environment(name: Buffer) -> napi::Result<Option<Buffer>> {
+    Ok(std::env::var_os(os_string(name)?).map(|value| wide_bytes(value.encode_wide())))
+}
+
+#[napi]
+pub fn windows_absolute_path(path: Buffer) -> napi::Result<BufferResult> {
+    let path = os_string(path)?;
+    if path.is_empty() {
```
mldangelo-oai, line 118: **[P3] Let the same Rust path primitive handle empty input**

Small correction to my earlier review premise: pinned Rust 1.97's `std::path::absolute` does not reject an empty path. It calls `GetFullPathNameW` through `fill_utf16_buf`, clears `LastError` before the call, and preserves any real OS error. This special case instead reads ambient `LastError` after a zero result, so it can return a stale error (or `0`) and adds 12 lines. Can we delete lines 118–129 and let the existing `match` handle every path?

plugins/codex-security/native/windows-files.mts
```diff
@@ -0,0 +1,214 @@
...
+    const result = native.windowsDirectoryEntries(operationPath(path));
+    check(result.error, path);
+    return result.value.map(({ name, isDirectory, isSymbolicLink }) => ({
+      name,
+      isDirectory: () => isDirectory,
+      isSymbolicLink: () => isSymbolicLink,
+    }));
+  }
+
+  function mkdir(path: Buffer): void {
```
mldangelo-oai, line 145: **[P3] Use Rust's recursive directory primitive**

Could this adapter call one native `std::fs::create_dir_all(os_string(path)?)`, mapped through the existing `io_status`, instead of recursing and then doing a separate `stat` fallback in JavaScript? Keep passing `operationPath(path)` so empty inputs retain their current failure path. Rust already owns parent creation, concurrent existing-directory races, wide/verbatim paths, and raw OS errors here. That removes roughly 13 production lines and the duplicated filesystem state machine.

plugins/codex-security/native/src/windows.rs
kmbroai: The public `std::path::absolute` entry point in the pinned compiler still rejects empty input before calling its Windows implementation ([source](https://github.com/rust-lang/rust/blob/8bab26f4f68e0e26f0bb7960be334d5b520ea452/library/std/src/path.rs#L4137)). Removing this branch would replace the numeric native result with an invalid-argument exception. I kept that compatibility branch, added `SetLastError(0)` before `GetFullPathNameW`, and added a proof that a preceding failed operation cannot change the empty-input result. Updated in `d8951fce`.

plugins/codex-security/native/windows-files.mts
kmbroai: The adapter now calls one native `std::fs::create_dir_all` operation through `io_status`, still passing `operationPath(path)`. Removed the JavaScript recursion and stat fallback. The Windows proof covers nested and existing directories, raw/verbatim long paths, empty input, and existing-file failures. Updated in `d8951fce`.

mldangelo-oai (approved):
Re-reviewed exact head d8951fce. Recursive directory creation is now one std::fs::create_dir_all call through the native error adapter. The retained empty-path branch is justified by pinned Rust 1.97 rejecting empty input before Win32, with LastError cleared before the compatibility call. All prior threads are resolved and all required checks pass.

PR #799, opened 2026-09-03
refactor(plugin): port security policy resolution to TypeScript

kmbroai:
## Summary

Replace the Python SECURITY.md resolver and its tests with a typed helper bundled alongside the MCP runtime. Policy discovery and root-to-leaf guidance use the existing Node launcher.

## Changes

- Enforce scan-root containment and preserve symlink and junction behavior, byte limits, UTF-8 handling, home expansion, inventory ordering, accepted arguments, and output formats.
- Preserve POSIX path bytes and Windows UTF-16 process values and paths; establish Windows ancestry by comparing native volume and file identities while walking parent directories.
- Invoke `launch_codex_security_mcp[.cmd] --helper resolve-security-md --repo PATH [--list | --scope PATH] [--out PATH]`, defaulting to stdout. Limit the SDK's Python interpreter instruction to `.py` helper scripts.
- Keep operator guidance focused on accepted options, quoted tilde paths, and output behavior. Replace complete Markdown snapshots with semantic source/content/order assertions and dedicated framing checks.
- Retain the compatible path resolver after native realpath comparisons showed different existing-file parent behavior. Remove the completed characterization probe and describe the active TypeScript policy helper in the native documentation.

## Testing

- SDK/MCP type checks, formatting, CI-tool compilation, Rust formatting, both MSVC-target Clippy checks, Ruff, portable source checks, and all nine checker tests passed.
- Focused policy/launcher tests passed: 34 passed, with five Windows-only tests skipped locally. All 12 adapter normalization tests passed.
- Windows x64/arm64 CI exercises the actual helper and native binding, including directory-identity checks for scope and policy paths. The raw-path proof also runs on Node 20 and 22.
- Existing behavioral tests retain argument parsing, home expansion, file-parent resolution, missing-path, symlink-cycle, ordering, and raw-byte coverage.

## Risk and rollout

Stacked on #798. The helper entrypoint changes from the Python script to the shared launcher, with callers updated together. Previously accepted arguments and home expansion remain supported intentionally. GNU Linux `file/..` compatibility requires retaining the existing resolver; native realpath is not a drop-in replacement. Other Python helpers remain for subsequent migration slices. There are no new top-level CLI flags.

## Public disclosure review


plugins/codex-security/mcp-app/src/helpers/posix-path.ts
```diff
@@ -0,0 +1,86 @@
...
+        /^[\udc80-\udcff]$/u.test(part)
+          ? Buffer.from([part.charCodeAt(0) - 0xdc00])
+          : Buffer.from(part),
+      ),
+  );
+}
+
+export class SymlinkLoopError extends Error {}
+
+export function resolvePosixPath(value: Buffer): Buffer {
```
mldangelo-oai, line 41: Before we own another component-by-component symlink resolver in containment-sensitive code, please prove that Node’s native implementation is insufficient. [`fs.realpathSync.native`](https://nodejs.org/api/fs.html#fsrealpathsyncnativepath-options) accepts a `Buffer` path and can return a `Buffer`; that is the representation already used here for undecodable POSIX filenames. If the direct matrix passes, this class can become one native call plus the existing `ELOOP` error mapping, removing roughly 35–40 production lines.

The validation should run on the supported Node floor across glibc, musl, and macOS and include a `0xff` name, invalid-byte symlink targets, relative links, `symlink/..`, `file/..`, missing components, and cycles. Node documents a conservative UTF-8/musl caveat, so an observed failure may justify a narrow fallback; it does not justify keeping a full Python-style resolver without first exercising the built-in. Incidental `pathlib` spellings such as `existing-file/..` should be an explicit compatibility decision, not the reason to own `realpath` indefinitely.

plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts
```diff
@@ -0,0 +1,502 @@
...
+): number {
+  try {
+    const options = {
+      repo: { type: "string" },
+      list: { type: "boolean" },
+      scope: { type: "string" },
+      out: { type: "string", default: "-" },
+      help: { type: "boolean", short: "h" },
+    } as const;
+    const names = Object.keys(options) as (keyof typeof options)[];
```
mldangelo-oai, line 412: Non-blocking compatibility/size question: this preprocessing loop is about 50 production lines whose only job is reproducing inherited `argparse` behavior—unique long-option prefixes, `-hfoo`/`-hh`, help short-circuiting other parse errors, and detached dash-leading values—and it drives about 71 lines of tests. I searched the shipped docs and non-test callers on this head; they use only full `--repo`, `--list`/`--scope`, and `--out` spellings.

Calling `node:util.parseArgs` directly would remove roughly 110–115 net lines; dash-leading paths remain unambiguous with `--repo=-1` and the corresponding full-option `=` forms. I do not think that is safe as a silent deslop, though: the Python helper shipped publicly, `argparse` accepted these inputs, this repository treats accepted CLI values as API, and this PR promises argument parity. Please either document these spellings as intentional compatibility or make the narrower contract explicit in the help/docs and PR risk (or split it into an approved follow-up). What we should avoid is freezing accidental parser behavior without deciding that it is actually part of the product contract.

plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts
```diff
@@ -0,0 +1,502 @@
...
+  try {
+    return windowsFiles().realpath(path);
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === "ELOOP")
+      throw new SymlinkLoopError(`Symlink loop from ${decodePath(path)}`);
+    throw error;
+  }
+}
+
+function expandHome(path: string, posixHome: string | undefined): string {
```
mldangelo-oai, line 103: This ports almost all of `Path.expanduser`, not just the useful current-user case: `~other`, Unix passwd lookup, Windows sibling-profile inference, `USERPROFILE` versus `HOMEDRIVE`/`HOMEPATH` precedence, drive-relative homes, and empty/unset-variable behavior. The implementation is roughly 48 lines, and the associated fixtures consume about 200 lines. Shipped callers pass an explicit repository root, and I found no documented caller using named-user expansion.

Could we make an explicit compatibility decision here? If current-user tilde expansion is sufficient, support only `~`, `~/...`, and `~\...` using the already transported `HOME`/raw `USERPROFILE` value with `os.homedir()` as the ordinary fallback. That removes the Unix `userHome` dependency, sibling-profile synthesis, and most of those tests. Because the old helper accepted these values, this should not be removed silently; either document named-user/pathological environment behavior as supported API or deliberately narrow it and call out the change.

sdk/typescript/tests-ts/security-policy-helper.test.ts
```diff
@@ -0,0 +1,876 @@
...
+    for (const name of ["\u{10000}", "\ue000", "\u0080", "\u007f"]) {
+      write(root, `${name}/SECURITY.md`, "policy\n");
+    }
+    const result = inventory(root);
+    expect(result.status, result.stderr).toBe(0);
+    expect(result.stdout).toBe(
+      '["\\u007f/SECURITY.md", "\\u0080/SECURITY.md", "\\ue000/SECURITY.md", "\\ud800\\udc00/SECURITY.md"]\n',
+    );
+    expect(resolve(root, "\u{10000}").stdout).toBe(
+      '## SECURITY.md source: "\\ud800\\udc00/SECURITY.md"\n\npolicy\n',
```
mldangelo-oai: The scoped `sdk/typescript/AGENTS.md` explicitly says to avoid tests of exact Markdown wording, but this file repeatedly snapshots complete documents—including headings, JSON quoting, blank lines, content, and trailing newlines—in tests whose subject is path selection, home expansion, encoding, or containment. The deleted Python tests asserted sources, ordering, and policy bodies independently, so the behavior does not require whole-document equality.

Please factor a small semantic assertion helper and have each case check only the source path(s), policy content, ordering, and framing invariant it owns. That removes roughly 30–40 repeated lines and prevents an unrelated wording/blank-line edit from breaking every path regression while retaining the security and platform coverage.

plugins/codex-security/references/security-guidance.md
```diff
@@ -7,9 +7,11 @@
 Compile the full `SECURITY.md` policy for a file or directory with:
 
 ```
-<python_command> <plugin_dir>/scripts/resolve_security_md.py --repo <repo_root> --scope <file_or_directory> --out <output_path_or_dash>
+<plugin_dir>/scripts/launch_codex_security_mcp --helper resolve-security-md --repo <repo_root> --scope <file_or_directory> --out <output_path_or_dash>
```
mldangelo-oai, line 10: This command is now intentionally invoked through the Node launcher, but the SDK-generated scan prompt still says to use `<python_command>` “for every plugin helper,” and its API test preserves that wording. An agent following both instructions can try `python launch_codex_security_mcp --helper ...`, which fails because the launcher is a shell or batch script, not Python.

Please narrow the SDK prompt to Python helper scripts (for example, `.py` files) and update its assertion. A focused prompt test should confirm that `resolve-security-md` remains a direct launcher invocation; the existing empty-`PATH` launcher coverage can continue proving managed Node execution.

plugins/codex-security/mcp-app/src/helpers/posix-path.ts
kmbroai: Added the direct Buffer realpath matrix in 02169139. Node 22.13.0 and 20.0.0 preserve raw filename/link-target bytes on glibc and musl, resolve relative links and symlink/.., and report missing paths and cycles correctly. macOS preserves raw link targets; its filesystem rejects the invalid filename fixture. Both Linux runtimes return ENOTDIR for file/.. and link-to-file/..; macOS resolves them to the parent. [glibc](https://github.com/openai/codex-security/actions/runs/33926927909/job/101197398900), [musl](https://github.com/openai/codex-security/actions/runs/33926927909/job/101197398844), [macOS](https://github.com/openai/codex-security/actions/runs/33926927909/job/101197398942).

I also ran the deleted Python helper against the file-parent cases: it accepts them. This migration intentionally preserves that accepted input, now documented in security-guidance.md and the PR risk section, so I retained the resolver. The evidence does not support an undecodable-byte fallback; the reason for retaining it is the explicit file-parent compatibility decision.

plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts
kmbroai: Made the compatibility choice explicit in security-guidance.md and the PR risk section in 02169139. This migration retains unique long-option prefixes, inherited help forms/error ordering, and the previously accepted detached dash-leading values. The guide recommends full option names and unambiguous = forms. Removing accepted arguments would narrow the public helper contract, so this change preserves them deliberately.

plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts
kmbroai: Documented named-user expansion and the home-variable rules as intentional compatibility in 02169139, including empty/unset values and Windows profile inference. This migration retains the public helper's accepted paths, with expansion limited to --repo and --scope as before; --out keeps tildes literal. I kept the implementation and behavior tests rather than silently narrowing that contract.

sdk/typescript/tests-ts/security-policy-helper.test.ts
kmbroai: Replaced the repeated whole-document comparisons with expectGuidance in 02169139. Path, home, encoding, and containment cases now assert ordered source paths and policy bodies independently of heading wording and surrounding blank lines. Dedicated framing/BOM/newline assertions remain where those are the behavior under test. The focused policy and launcher suite passes.

plugins/codex-security/references/security-guidance.md
kmbroai: Fixed in 02169139: the SDK prompt now limits <python_command> to plugin Python helper scripts (.py files). The focused API test checks that wording and verifies the bundled resolve-security-md command remains a direct launcher invocation. That test and the existing launcher coverage pass.

plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts
kmbroai: Follow-up in ad511edc: malformed short-help arguments such as -h-, -hh-, and -h--help now retain the previous error exit status. All 19 help-suffix comparisons match the deleted Python helper, including accepted forms such as -hfoo-.

plugins/codex-security/scripts/launch_codex_security_mcp
```diff
@@ -6,6 +6,12 @@ export PATH
 
 launcher_dir=$(CDPATH= cd "$(dirname "$0")" && pwd)
 server_path=$launcher_dir/../mcp/server.mjs
+if [ "${1:-}" = "--helper" ]; then
+  server_path=$launcher_dir/../mcp/helpers.mjs
```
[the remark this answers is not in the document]
kmbroai: @codex security dismiss out-of-scope Reproduced the runtime selection with a harmless marker. It requires the caller to put the repository in PATH and have no managed Node candidate. This change does not add the repository to PATH; the launcher only appends fixed system directories. With an ordinary external PATH, the repository executable is not selected, and a configured managed runtime takes precedence.

The helper adds a call from the scan directory, whereas the old SDK helper used its protected Python selection. That difference does not remove the caller-controlled PATH prerequisite. The existing POSIX launcher selects the same repository executable under that configuration. SECURITY.md's local-tool trust model and out-of-scope criteria cover operator-selected PATH executables; this report does not establish an input crossing a separate protected boundary.

plugins/codex-security/native/proof.mts
```diff
@@ -70,6 +71,95 @@ const fixtureName = (prefix: string, byte: number) =>
     ? Buffer.from(`${prefix}-é`)
     : Buffer.from([prefix.charCodeAt(0), byte]);
 
+function realpathProof(root: string) {
```
mldangelo-oai: **[P3] Remove the one-time realpath characterization**

This matrix answered the earlier realpath question. Now that its results are preserved in the resolved thread and PR testing notes, can we remove `realpathProof` from the permanent native proof? It adds 92 net lines, tests `node:fs` rather than this resolver, accepts either result for the differentiating `file/..` and macOS-fixture cases, and has no workflow consumer for its JSON. The end-to-end helper tests already lock the compatibility behavior we retain. Please also trim the source and native README claims that CI still runs this matrix.

plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts
```diff
@@ -0,0 +1,503 @@
...
+  return left.length - right.length;
+}
+
+function listSecurityMd(repo: string, posixHome: string | undefined): string[] {
+  const root = resolveRoot(repo, posixHome);
+  const policies: string[] = [];
+  function walk(directory: Buffer, prefix: string): void {
+    const entries = (
+      windows
+        ? windowsFiles().entriesWithTypes(directory)
```
mldangelo-oai, line 265: **[P3] Refresh the native usage note**

This shipped Windows `--list` path now calls `entriesWithTypes`, so `native/README.md` still saying “product commands do not use it yet” is no longer true on this head. Please update or remove that sentence in this PR, and refresh the opening “SDK and CLI continue to use their existing helpers” wording while there.

mldangelo-oai (commented):
Re-reviewed the exact current head and reconciled all prior threads. The compatibility behavior is well covered; the remaining comments are nonblocking simplification and documentation cleanup.

plugins/codex-security/native/proof.mts
kmbroai: Confirmed. Removed realpathProof and its output and documentation claims in 9986e31d. The helper tests still cover the compatibility resolver, including file/.., symlink cycles, missing paths, and raw bytes.

plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts
kmbroai: Confirmed. Updated native/README.md in 9986e31d to describe the active TypeScript policy helpers and the Windows --list use of entriesWithTypes.

plugins/codex-security/references/security-guidance.md
```diff
@@ -7,9 +7,17 @@
...
 ```
 
+On Windows, use `launch_codex_security_mcp.cmd` with the same arguments. The launcher reuses the plugin's configured or bundled Node runtime and preserves the working directory for relative helper paths.
+
+The launcher preserves the public Python helper's argument and path behavior. Prefer the full option names shown above; unique long-option prefixes such as `--r`, `--s`, and `--o` remain accepted. The inherited help forms (`-h`, `--help`, `-hh`, and `-hfoo`) and help short-circuiting of unrelated parse errors are retained. Missing option values and invalid attached values still fail before later help. Detached negative-number paths and otherwise unrecognized dash-leading values containing spaces remain accepted as option values; `--repo=-1` and corresponding full-option `=` forms are unambiguous.
```
mldangelo-oai: **[P3] Keep migration archaeology out of operator guidance**

Thanks for making the compatibility choice explicit. These three paragraphs now spend 274 words on parser error ordering, pathological home-variable states, and a GNU `realpath` discrepancy—details an agent invoking the helper cannot act on. Could we replace them with the actionable contract, for example:

> For compatibility, the resolver retains the former Python helper's option abbreviations, help parsing, home expansion, and path resolution; prefer the full option names shown above.
>
> Quote tilde paths so the helper expands them for `--repo` and `--scope`; `--out` keeps tildes literal, and `--out -` writes to stdout.

That preserves the supported surface while cutting about 227 words and avoids turning implementation quirks into durable product guidance.

kmbroai: Replaced the implementation-history paragraphs with the actionable compatibility and quoted-tilde/output contract. The existing behavioral tests continue to cover argument parsing and path expansion. Updated in `90e02bd9`.

mldangelo-oai (approved):
Re-reviewed exact head 90e02bd9. Windows containment now compares native directory identities while walking canonical parents, covering the Unicode casing collision without recreating path matching in JavaScript. The operator guidance is reduced to the actionable contract; all prior threads are resolved and all required checks pass.

PR #805, opened 2026-09-03
feat(plugin): suggest scan models and reasoning settings

daneschneider-oai:
## Summary

Add one pre-scan nudge for non-Astra cyber models, available model upgrades, and reasoning below `xhigh`.

## Changes

Use effective host settings and catalog upgrade metadata. Extract a shared app-server client; keep permission checks in Deep Scan. Ask before a new chat-started desktop scan begins.

## Interaction and copy

Show one combined, applicable question through the app's blocking input form:

> {current_model} is designed for dynamic exploitation. You may get better vulnerability scanning results with {recommended_non_cyber_model}. A newer recommended model, {new_model}, is available in your Codex model catalog. For the best scanning results, would you like to use {new_model} with xhigh reasoning? You can change the model or reasoning selector, or keep going with your current settings.

Offer **Change settings (Recommended)** and **Continue current settings**. Continue only after an explicit Continue; Change stops before scan creation so the user can adjust the selector and resubmit. If input is unavailable, ask in chat and stop. Unknown `xhigh` support adds `if your model supports it`. Cyber-only advice has no question and continues after a visible warning. Native Security tab scans skip repeated advice when their launch instructions explicitly say the New scan form handled model guidance before launch; older native continuations retain a declarative advisory. Headless scans receive one warning.

## Testing

All 25 MCP scripts, 17 custom-validation tests, typecheck, formatting, and source checks passed; verified source and bundled runtimes.

## Risk and rollout

Named upgrades require catalog evidence; missing metadata leaves generic advice. Catalog data may not reflect app-specific filtering or runtime-only profile overrides.

Companion: [CLI #806](https://github.com/openai/codex-security/pull/806).

## Public disclosure review


sylvesterkaczmarek (commented):
Hi, `handleMessage()` does not actually ignore every server-initiated request as the comment says. Codex `RequestId` allows integer IDs, so a server request with a numeric `id` is treated as the pending client response and can trigger the mismatched-ID/protocol failure. Could this distinguish requests by `method` (or responses by `result`/`error`) before matching the ID?

sylvesterkaczmarek (commented):
One protocol edge case in `readModelCatalog()`: the Codex `ModelListResponse` schema makes `nextCursor` optional, but this loop only treats explicit `null` as end-of-pagination. A valid final page that omits the field is rejected as an invalid cursor. Could the termination check use `page.nextCursor == null`?

plugins/codex-security/mcp-app/src/model-guidance.ts
```diff
@@ -0,0 +1,102 @@
...
+  const { model, reasoningEffort } = settings;
+  const current =
+    model && catalog ? findCatalogModel(model, catalog) : undefined;
+  const upgrade =
+    model && catalog ? latestAvailableUpgrade(model, catalog) : undefined;
+  const nonCyberModel =
+    upgrade && !isCyberModel(upgrade.model)
+      ? upgrade
+      : catalog
+        ? recommendedNonCyberModel(catalog)
```
mldangelo-oai, line 33: **[P2] Keep the selected chain's non-cyber recommendation**

`latestAvailableUpgrade()` returns the final eligible target, including Astra. For `legacy-cyber -> scan-specialist -> astra-cyber`, `upgrade` is Astra, so this branch discards `scan-specialist` and falls back to an unrelated catalog default. The resulting guidance can recommend `unrelated-default` for vulnerability scanning while separately offering `astra-cyber`, even though the selected model explicitly points to `scan-specialist`. Please retain the latest visible non-cyber candidate while walking the selected model's declared chain, then fall back to a visible non-cyber default and its chain; cover both paths.

plugins/codex-security/mcp-app/src/app-server-client.ts
```diff
@@ -0,0 +1,264 @@
...
+    const args: string[] = [];
+    for (const override of options.configOverrides) {
+      args.push("--config", override);
+    }
+    args.push("app-server", "--stdio");
+
+    this.child = spawn(executablePathForSpawn(options.codexPath), args, {
+      cwd: options.cwd,
+      ...(options.env === undefined ? {} : { env: options.env }),
+      stdio: ["pipe", "pipe", "pipe"],
```
mldangelo-oai, line 69: **[P2] Hide the pre-scan subprocess on Windows**

Please set `windowsHide: true` on this spawn. This shared client now backs routine model guidance, so an ordinary Windows desktop scan can start the catalog app-server with Node's default `windowsHide: false` and flash a console window before the scan. The repository's other production Codex spawns hide their windows, and #806's equivalent model-catalog spawn does too. This one option preserves the stdio behavior and avoids the Windows UI regression.

plugins/codex-security/mcp-app/tests/test_model_guidance.mjs
```diff
@@ -0,0 +1,283 @@
...
+  "Classify the catalog model behind a selected ID.",
+);
+assert.match(advice, /scan-upgrade/);
+assert.doesNotMatch(
+  advice,
+  /scan-default/,
+  "Prefer the declared non-cyber upgrade to an unrelated default.",
+);
+assert.match(advice, /use scan-upgrade with xhigh reasoning\?/);
+for (const cyber of [false, true]) {
```
mldangelo-oai, line 90: **[P3] Collapse the duplicated guidance matrix**

The MCP runtime matrix below already covers cyber-only warning, low-effort actionable guidance, their combination, and no guidance through both source and bundled servers; `assertDeliveryInstruction` owns the delivery contract, and the `advice` assertions above cover the maximal combined suggestion. Keep the one distinct catalog-upgrade-only case plus a single-question assertion on `advice`, but collapse this 2x2x2 table. That preserves the meaningful branches while removing roughly 18-20 lines of repeated expectations.

mldangelo-oai (commented):
Reviewed exact draft head `3615bb06584944ba691550ff1b907bb99e3b9894`. Three focused comments below; the existing JSON-RPC request-classification and pagination feedback still applies.

PR #806, opened 2026-09-03
feat(cli): offer scan model and reasoning upgrades

daneschneider-oai:
## Summary

Offer one model/effort nudge before scanning. Keep startup and prompts in the normal terminal until scan activity begins.

## Changes

Use catalog-backed upgrades and share choices across matching scans. Check native ChatGPT authentication first; require TTY input and output for interactive UI.

## Interaction and copy

Combine applicable sentences into one stderr warning:

> codex-security: warning: {current_model} is designed for dynamic exploitation. You may get better vulnerability scanning results with {recommended_non_cyber_model}. {new_model} is available as an upgrade to {current_model} and may give better scanning results. The configured reasoning effort is {current_effort}. Use xhigh for the best scanning results.

Ask one confirmation for eligible changes:

- `Use {new_model} with xhigh reasoning for this scan?`
- `Use {new_model} for this scan?`
- `Use xhigh reasoning for this scan?`

Default **No**. **Yes** applies the proposed changes for this command. Headless/JSON/CI runs warn without prompts or changes. Dry runs skip the lookup. Missing catalog metadata still permits the `xhigh` opt-in; explicitly unsupported efforts are excluded.

Detected sign-in failure:

> Codex Security's stored ChatGPT sign-in is no longer available. Run 'codex-security login' and retry.

## Testing

293 focused tests, real PTY checks, and bundled Codex 0.149.1 with synthetic local authentication responses. Typecheck and formatting passed.

## Risk and rollout

Catalog failures remain advisory. Saved settings and cost limits are preserved. Authentication uses native conditional refresh.

Companion: [plugin #805](https://github.com/openai/codex-security/pull/805).

## Public disclosure review


sylvesterkaczmarek (commented):
Hi, the no-catalogue path treats an unknown model as supporting xhigh (`model === undefined`), so the CLI can still print “Use xhigh for the best scanning results” even when it could not load the catalogue or verify support. Could the warning keep the same “if your model supports it” qualifier instead of asserting xhigh for an unknown model?

sdk/typescript/src/cli-model-selection.ts
```diff
@@ -0,0 +1,106 @@
...
+  prompt: Pick<BulkScanPrompt, "isInteractive" | "confirm">;
+  write(message: string): void;
+}): ScanModelSelector {
+  const choices = new Map<string, Promise<ScanModelConfiguration>>();
+  return (configuration, loadModels, signal) => {
+    const key = JSON.stringify(configuration);
+    let choice = choices.get(key);
+    if (choice === undefined) {
+      choice = choose(configuration, loadModels, signal);
+      choices.set(key, choice);
```
mldangelo-oai, line 22: **[P2] Evict rejected selections from the shared cache**

`choices` retains a rejected promise indefinitely. If the first selection for a configuration rejects during authentication discovery or confirmation, later component workers and bulk-scan retries reuse that rejection without invoking their own loader or prompt. Please preserve the useful in-flight and successful-choice caching, but remove the entry on rejection with an identity guard so an older promise cannot delete a replacement. Add a regression showing that a second call invokes a fresh loader and succeeds after the first rejects.

sdk/typescript/src/cli.ts
```diff
@@ -3625,37 +3666,50 @@ export async function main(
                 ? {}
                 : { maxCostUsd: options.maxCost }),
             },
-            createSecurity: dependencies.createSecurity,
+            createSecurity: (config) =>
```
mldangelo-oai, line 3669: **[P2] Apply the accepted model choice to the whole component command**

Passing `chooseModel` only into each `CodexSecurity` worker makes the accepted choice worker-local. With `--auto`, planning has already run with `config` before selection occurs, and cross-component matching later receives that same unchanged config. Accepting the offered upgrade therefore changes component scans and the dashboard but leaves planning and root-cause matching on the old model and effort, despite the README saying the choice applies to the current command. Please resolve selection once before component orchestration and pass one command-local effective config to planning, workers, matching, and the dashboard; add an integration test covering all three model-call stages.

sdk/typescript/src/model-advice.ts
```diff
@@ -0,0 +1,108 @@
...
+}): ScanModelAdvice {
+  const selectedModel = findModel(availableModels, model);
+  const modelUpgrade = latestUpgrade(
+    selectedModel,
+    availableModels,
+    (candidate) => !isNonAstraCyberModel(candidate.model),
+  );
+  let cyberWarning: string | undefined;
+  if (isNonAstraCyberModel(selectedModel?.model ?? model)) {
+    const defaultModel = availableModels?.find(
```
mldangelo-oai, line 25: **[P3] Drive both guidance surfaces from one policy contract**

`getScanModelAdvice()` duplicates the pure upgrade/default/cyber/reasoning policy also added in #805, and the implementations already diverge. For example, #805 can follow a cyber default's declared chain to a visible non-cyber successor, while this implementation filters out that default before traversal. Keep the package-specific app-server, authentication, and prompt adapters separate, but put the pure policy in a shared module if packaging permits, or at least run both implementations against shared conformance vectors. That prevents the two surfaces from producing different advice for the same catalog.

mldangelo-oai (commented):
Reviewed exact draft head `5281ef76d3021a42b216aeac2e96831fa6fc7847`. Two correctness issues and one cross-surface simplification are inline below. I did not repeat the existing unknown-model/xhigh feedback, which remains applicable.

PR #808, opened 2026-09-04
perf: speed up confirmed finding matching

mldangelo-oai:
## Summary

Matching many distinct finding identities repeatedly scanned both occurrence lists for every confirmed group. Indexing occurrences by identity removes that quadratic validation work while preserving the existing grouping rules.

## Changes

- Index occurrences once and validate each confirmed identity group using its members.
- Add generated regression coverage for repeated occurrences, transitive aliases through absent historical identities, one-sided groups, and reordered inputs.
- Give publication fixtures their own state directories and credential fixtures explicit environments so ambient API keys and state paths cannot change test results.

## Testing

- Focused matching suite: 107 passed, 3 platform-specific skips; comparison properties ran with 1,000 cases each.
- Publication and credential tests under a synthetic ambient API key and unusable ambient database path: 86 passed across four files.
- Independent graph oracle: 1,000 generated cases; results equal main and remain correct with reversed input order.
- Built CLI: 8 checks passed, covering saved matches, cache reuse, comparison output, malformed responses, cancellation, missing artifacts, and unchanged sealed artifact hashes.
- SDK build, types, and formatting passed.
- Packed and installed-package checks passed, including public imports, types, CLI, lifecycle, bundled plugin files, MCP initialization, and dashboard assets.

Local Node.js 24 benchmark, median of three runs per case, using synthetic distinct identities and no model calls:

| Finding pairs | Main | This change |
| --- | ---: | ---: |
| 1,000 | 43 ms | 4 ms |
| 2,000 | 168 ms | 8 ms |
| 4,000 | 919 ms | 17 ms |
| 8,000 | 3,598 ms | 37 ms |

These measurements describe the deterministic identity-matching path; model latency is outside the benchmark.

## Risk and rollout

The added index uses memory proportional to the number of occurrences. Existing match validation, CLI behavior, stored comparisons, and sealed scan artifacts remain compatible. The environment changes apply only to tests. This can ship through the normal release process.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/scan-comparison.ts
+++ b/sdk/typescript/src/scan-comparison.ts
@@ -1258,18 +1258,16 @@ function validateComparison(
     ...[...new Set(findingIds.values())].map((findingId) => [findingId]),
   ]);
   if (enforceConfirmedIdentities) {
+    const occurrencesByFinding = Map.groupBy(
+      findingIds.keys(),
+      (occurrenceId) => findingIds.get(occurrenceId)!,
+    );
     for (const knownGroup of confirmedGroups) {
-      const knownFindingIds = new Set(knownGroup);
-      const knownBefore = input.before.filter(({ occurrenceId }) => {
-        const findingId = findingIds.get(occurrenceId);
-        return findingId !== undefined && knownFindingIds.has(findingId);
-      });
-      const knownAfter = input.after.filter(({ occurrenceId }) => {
-        const findingId = findingIds.get(occurrenceId);
-        return findingId !== undefined && knownFindingIds.has(findingId);
-      });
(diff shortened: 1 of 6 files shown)


kmbroai (approved):
not sure people are using this, or should be using this, but approved

sylvesterkaczmarek (commented):
The indexing preserves the identity semantics while removing the repeated before/after scans: each finding ID is resolved to its occurrence set once, and the later checks still distinguish matched groups, unmatched occurrences, and presence on both sides. The transitive historical-ID property test plus reversed input order is particularly useful because it guards against an accidental dependence on iteration order. Narrowing the spawned-test environment also makes these regressions less sensitive to ambient credentials/state.

PR #810, opened 2026-09-04
fix(multiscan): persist scan warnings in receipts

### LINKED ISSUE #248: bulk-scan discards run warnings, so a drifted repository is recorded as completed with no signal

sylvesterkaczmarek:
## Summary

Preserve bulk-scan run warnings in the durable `results.jsonl` ledger so successful repositories do not lose warning evidence after the live progress event has passed.

Part of #248. Follow-up #973 tracks surfacing persisted run warnings in the campaign summary.

Current `main` already forwards `security.run()` warnings to live progress. This PR closes the persistence/replay gap: a completed scan can now retain its run warnings in the receipt and surface them again when a completed/incomplete receipt is reused on campaign resume.

## Changes

- collect every warning emitted by `security.run()` for an attempt;
- write non-empty warnings to the attempt receipt as `warnings` while keeping the existing singular coverage `warning` field compatible;
- replay persisted warnings through `onProgress` when a completed/incomplete receipt is reused on campaign resume;
- extend the warning regression to verify multiple warnings survive the ledger and a resume does not rerun the repository.

## Testing

After rebasing onto current `main`:
- `tsc --noEmit` — passed
- Prettier check on both changed files — passed
- `git diff --check` — passed
- Bun is not installed in the local rebase environment, so the focused Bun suite could not be rerun locally. The pre-rebase focused suite was independently verified by review as 52 passed, 1 skip, 0 failed, and the rebased test conflict applied cleanly.

## Risk and rollout

Low. The ledger change is additive and older receipts without `warnings` remain readable. Scan status, retry behavior, coverage handling, policy-failure accounting, recovery behavior, and live warning delivery are unchanged. Resumed campaigns gain warning replay for receipts that include the new field.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/multiscan.ts
+++ b/sdk/typescript/src/multiscan.ts
@@ -56,6 +56,7 @@ interface MultiscanReceipt extends MultiscanTask {
   cost?: ScanCost;
   error?: string;
   warning?: string;
+  warnings?: string[];
 }
 
 export interface MultiscanOptions {
@@ -166,6 +167,16 @@ async function runCampaign(
         receipt.outputDir === selectedArtifactOutput) &&
       (await hasArtifacts(artifactOutput))
     ) {
+      if (receipt.status !== "failed") {
+        for (const warning of receipt.warnings ?? []) {
+          notifyProgress(options, {
+            repository: task.id,
+            status: receipt.status,
+            attempt: receipt.attempt,
+            warning,
+          });
+        }
+      }
       if (receipt.status === "completed") {
         completed += 1;
         continue;
@@ -228,6 +239,7 @@ async function runCampaign(
         notifyProgress(options, { ...progress, status: "started" });
         let failure: string | undefined;
         let warning: string | undefined;
+        const runWarnings: string[] = [];
         let coverage: CoverageDocument["completeness"] | undefined;
         let cost: Readonly<ScanCost> | null = null;
         let exhaustedBudget = false;
@@ -272,12 +284,14 @@ async function runCampaign(
             ...(options.maxCostUsd === undefined
               ? {}
               : { maxCostUsd: options.maxCostUsd }),
-            onWarning: (warning) =>
+            onWarning: (warning) => {
+              runWarnings.push(warning);
               notifyProgress(options, {
                 ...progress,
                 status: "started",
                 warning,
-              }),
+              });
+            },
             ...(options.signal === undefined ? {} : { signal: options.signal }),
           });
           cost = result.cost;
@@ -317,6 +331,7 @@ async function runCampaign(
             ...(cost === null ? {} : { cost }),
             ...(failure === undefined ? {} : { error: failure }),
             ...(warning === undefined ? {} : { warning }),
+            ...(runWarnings.length === 0 ? {} : { warnings: runWarnings }),
           })}\n`,
         );
         notifyProgress(options, {
--- a/sdk/typescript/tests-ts/multiscan.test.ts
+++ b/sdk/typescript/tests-ts/multiscan.test.ts
@@ -369,6 +369,7 @@ describe("multiscan", () => {
         paths,
         client(async (_repository, scanOptions = {}) => {
           scanOptions.onWarning?.("Could not run post-scan instructions.");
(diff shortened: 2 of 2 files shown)


feiiiiii5 (commented):
Ran this locally to check two things a reader cannot get from the diff: whether the extended test actually bites, and where the branch stands against current `main`.

Environment: macOS arm64, bun 1.3.14, node 24.19.0, `pnpm install --frozen-lockfile`, with `_bundled_plugin` staged locally. The command in every case is `bun test --timeout 30000 tests-ts/multiscan.test.ts` from `sdk/typescript`.

**The extended test bites.**

- at your head `aafc4cd1`: **52 pass, 1 skip, 0 fail** — the same count your description reports.
- your test file run against your base `f9b686a6`: **51 pass, 1 skip, 1 fail**, and the failure is `surfaces optional post-scan warnings without failing completed scans` at `multiscan.test.ts:386`, i.e. the new ledger and resume assertions. So the persistence gap is real, and the test pins the fix instead of merely describing it.

**The branch is 102 commits behind `main`, and that shows up in ways CI will not explain for you.** Running your test file against current `main` (`70d5b2ed`) gives **49 pass, 1 skip, 3 fail**. One of the three is the test above, which is expected because `main` has no `warnings` field. The other two are `keeps sealed partial-coverage CLI runs fail-closed without retrying` and `keeps sealed unknown-coverage CLI runs fail-closed without retrying`, which this PR does not touch: they drive `main` from `../src/cli.js`, and `src/cli.ts` changed by +2450/-882 lines between your base and current `main` (including `9c06032f`, recover interrupted bulk campaigns, and `408ce446`, share scan settings across the CLI and SDK). Those two are staleness rather than a regression from your change, but a rebase is needed before any CI result on this branch means something.

**`Fixes #248` would close more than this diff does.** The issue's suggested direction has two parts. Part 1 — an `onWarning` observer plus warnings on the attempt receipt — is what this implements, and the field shape matches what the issue asked for: omitted when empty, like the receipt's own `error` and `cost`. Part 2, "surface them in the campaign summary the way failures already are", is not in the diff. The summary counters still key off the coverage warning only, so a repository whose scan emitted run warnings is still counted as a plain `completed` (`src/multiscan.ts:503-506`, with `warning` assigned only in the coverage branch at `:448`). Either "Part of #248" with a named follow-up, or the summary accounting here — otherwise the issue closes with half of its direction undone.

**One bullet in the description states behavior `main` already has.** "preserve successful complete scans as `completed` rather than conflating operational warnings with incomplete coverage": `status` derives from `failure` and the coverage `warning` (`:467-472`), and run warnings never fed into either, so a warned-but-complete scan is already recorded as `completed`. The diff does not touch that computation. Worth rewording so a reviewer does not go looking for a status change that is not there.

**Two questions on the replay path, neither blocking.**

1. Live, a run warning is emitted with `status: "started"` (`:422-428`); the replay emits it with the receipt's final status, so the same text travels with `status: "completed"` on resume. Your test asserts that shape, so it looks intentional — but a consumer keying on `status` now sees two shapes for one warning. Is that the intended contract?
2. The replay loop skips `failed` receipts, while `warnings` is written for every attempt including failed ones. A retry re-emits them live if they recur, so nothing is lost in the common case, but the ledger can hold warnings that no consumer ever surfaces. Deliberate?

I also checked the compatibility claim rather than assuming it: `results.jsonl` has no JSON schema in this repository (`sdk/typescript/schemas/` holds only `project-config.schema.json`), `readReceipts` is a tolerant `JSON.parse` cast (`:859-889`), and no Markdown documents the receipt fields. So an additive `warnings` key needs no schema or documentation sync, and older ledgers stay readable, as the description says.

sylvesterkaczmarek:
Rebased onto current main in signed+DCO commit c1040247. The conflict resolution keeps current recovery and policy-failure handling and layers warning persistence/replay on top. I also changed the PR from "Fixes #248" to a partial fix and opened #973 for the campaign-summary accounting you called out. The description no longer claims this PR changes completed-status semantics. TypeScript and Prettier checks pass after the rebase; Bun is not installed on this machine, so I could not rerun the exact Bun suite locally.

PR #811, opened 2026-09-04
fix(sdk): advance Linear pagination connections

### LINKED ISSUE #516: Linear project import can loop after the first page of issues
## Summary

Importing issues from a Linear project can loop indefinitely when the project has more than one page of matching issues.

## Root cause

`importLinearIssues()` fetches the first 50 project issues and then does:

```ts
const page = await projects.nodes[0]!.issues({ first: 50, filter });
while (page.pageInfo.hasNextPage) await page.fetchNext();
issues.push(...page.nodes);
```

The Linear SDK pagination API returns the next connection from `fetchNext()`; the caller is expected to use that returned connection. The current code discards it and continues checking `page.pageInfo.hasNextPage` on the original first page.

If that first page reports `hasNextPage: true`, the loop can repeatedly request the same next page without ever advancing `page`. Even if an SDK implementation happened to mutate internal state, only the original page's `nodes` are appended after the loop, so relying on mutation would make the collection semantics ambiguous.

Linear's SDK documentation shows the intended pattern explicitly:

```ts
const issues = await linearClient.issues({ first: 10 });
const nextIssues = await issues.fetchNext();
```

## Expected behavior

Accumulate the current page's nodes, assign the result of `fetchNext()` as the next current page, and stop when that page reports no next page.

## Impact

`--linear-project` intake can hang for projects whose filter matches more than 50 issues, instead of importing the complete project issue set.

## Suggested fix

Advance through returned connection objects explicitly and add a mock pagination regression that proves the second page is consumed exactly once.

Hughhhhcoder:
Fixes #516

## Reproduction

When a Linear project has more than one page of matching issues, the first connection reports `pageInfo.hasNextPage: true`, but `fetchNext()` returns the next connection. The current code discards that return value and keeps checking the first page, so the import can loop indefinitely.

The same discarded-connection pattern also affected paginated issue comments.

## Expected behavior

Advance through each returned Linear connection and import every page exactly once.

## Actual behavior

The project import repeatedly fetches from the original first page and never reaches the terminal page when the SDK returns a new connection object.

## Root cause

Both pagination loops awaited `fetchNext()` without assigning its returned connection, then read only the original connection's nodes.

## Fix

Consume the current page before assigning the connection returned by `fetchNext()`. Apply the same traversal to issue comments so imported evidence is complete.

## Validation

- `bun test --timeout 30000 ./tests-ts/cli-skills.test.ts`: 30 passed
- Added immutable multi-page project and comment fixtures that fail under the old implementation
- Prettier check: passed
- TypeScript `tsc --noEmit`: passed
- `git diff --check origin/main...HEAD`: passed

The branch was based on and tested against the latest upstream `main` at `f9b686a6`.

### DIFF
--- a/sdk/typescript/src/linear.ts
+++ b/sdk/typescript/src/linear.ts
@@ -83,9 +83,12 @@ export async function importLinearIssues(options: {
         );
       }
 
-      const page = await projects.nodes[0]!.issues({ first: 50, filter });
-      while (page.pageInfo.hasNextPage) await page.fetchNext();
-      issues.push(...page.nodes);
+      let page = await projects.nodes[0]!.issues({ first: 50, filter });
+      while (true) {
+        issues.push(...page.nodes);
+        if (!page.pageInfo.hasNextPage) break;
+        page = await page.fetchNext();
+      }
       if (issues.length === 0) {
         throw new CodexSecurityError(
           `No open Linear issues matched project "${options.project}" and its filter.`,
@@ -114,8 +117,13 @@ export async function importLinearIssues(options: {
 
     const imports: ImportedIssue[] = [];
     for (const issue of issues) {
-      const comments = await issue.comments({ first: 50 });
-      while (comments.pageInfo.hasNextPage) await comments.fetchNext();
+      let comments = await issue.comments({ first: 50 });
+      const commentNodes = [];
+      while (true) {
+        commentNodes.push(...comments.nodes);
+        if (!comments.pageInfo.hasNextPage) break;
+        comments = await comments.fetchNext();
+      }
       imports.push({
         source: "linear",
         id: issue.identifier,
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Correct pagination ownership. `fetchNext()` returns the successor connection, so advancing the connection variable and consuming each page before fetching the next one avoids both the first-page loop and dropped terminal-page nodes. Applying the same traversal to comments is important because incomplete comment evidence can change the security context even when issue enumeration itself is correct. The immutable page fixtures are a good regression because they no longer accidentally mask the bug through in-place mutation.

sylvesterkaczmarek (commented):
This fixes the pagination contract at the right level. The important part is consuming the current connection before replacing it with the object returned by `fetchNext()`, and doing the same for comments rather than assuming the SDK mutates the original connection in place.

The regression fixtures are particularly useful because they return distinct, immutable next-page objects; that would immediately catch a future reintroduction of the old mutation assumption. The issue and comment traversal now look consistent to me.

PR #814, opened 2026-09-04
feat(cli): create GitLab merge requests for verified patches

kmbroai:
## Summary

`patch --create-pr` currently invokes GitHub's CLI for every repository. It now creates a draft GitLab merge request with `glab` when the origin push URL identifies GitLab.

## Changes

- Support GitLab.com and self-hosted GitLab through the existing `GITLAB_HOST` setting (including glab's `GITLAB_URI` and `GL_HOST` aliases).
- Bind the merge request's source and target projects to the origin push URL, including SSH remotes and subgroup paths.
- Reuse verified patch commits, patch-risk summaries, and `patch --resume-pr` recovery. Preserve the `pullRequest: { branch, url }` JSON result and existing GitHub arguments.
- Update command help, the interactive patch browser, documentation, and publication/retry regression tests.

## Testing

- `bun test --timeout 30000 tests-ts/cli-patch.test.ts --seed 12345`: 42 passed, 0 failed.
- `pnpm run types` and `pnpm run format` passed.
- Hosted CI is running for the updated commit.

## Risk and rollout

The existing `--create-pr` and `--resume-pr` syntax and defaults are unchanged. GitLab requires an installed, authenticated `glab`; self-hosted repositories also require the matching GitLab host setting when creating or resuming publication. Other hosts retain the GitHub workflow. GitLab API calls are mocked in regression tests; no live merge request was created during validation.

## Public disclosure review

- [x] No customer, partner, prospect, or user identities, data, or identifying details are included.
- [x] No credentials, personal data, private source, scan findings, or nonpublic links or tickets are included.
- [x] I reviewed the branch name, title, description, commits, changes, comments, logs, screenshots, attachments, and links for public disclosure.

PR #817, opened 2026-09-07
fix(cli): correct GitLab patch publication

kmbroai:
## Summary

Match GitLab patch merge requests by branch and project identity, normalize SSH repository selectors, and preserve generated descriptions as advisory text. This follows up on #814.

## Changes

- Reuse an existing merge request only when its source and target project IDs match. Request only those IDs and the URL from glab to avoid buffering full descriptions.
- Normalize SCP-style origins to explicit SSH URLs for glab source and target project selection.
- Escape quick-action text in GitLab descriptions while preserving the CLI report and GitHub body.
- Check the default body in the routing matrix and remove duplicate GitLab retry rows; retain missing-client recovery and repeated reuse coverage.

## Testing

- Focused CLI patch and patch-browser tests: 55 passed, 0 failed, including fork-only and fork-before-same-project results, SSH selectors, and description handling.
- `pnpm run types` and `pnpm run format` passed.
- A subprocess probe reproduced the buffer failure with 1.23 MB of MR metadata; projecting the required fields returned 4.5 KB and preserved the selected MR.
- GitLab API calls are mocked in regression tests; no live GitLab merge request was created during validation.

## Risk and rollout

The existing command syntax, defaults, JSON result, and GitHub publication arguments are unchanged. These changes apply to GitLab selection and publication through the existing `--create-pr` and `--resume-pr` paths.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -5387,6 +5387,9 @@ async function publishPatchBranch(
             gitlabHost.includes("://") ? gitlabHost : `https://${gitlabHost}`,
           ));
     const command = gitlab ? "glab" : "gh";
+    const gitlabRepository = remote.includes("://")
+      ? remote
+      : `ssh://${remote.replace(":", "/")}`;
     const existing = await run(
       command,
       gitlab
@@ -5399,7 +5402,7 @@ async function publishPatchBranch(
             "--output",
             "json",
             "--repo",
-            remote,
+            gitlabRepository,
           ]
         : [
             "pr",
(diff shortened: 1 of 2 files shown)


sdk/typescript/src/cli.ts
```diff
@@ -5416,6 +5419,16 @@ async function publishPatchBranch(
             ".[0].url // empty",
           ],
     );
+    let url = gitlab
```
mldangelo-oai, line 5422: Could we keep this in the existing glab --jq pipeline? Using map(select(.source_project_id == .target_project_id))[0].web_url // empty returns the same URL-or-empty contract as the GitHub branch, so url can remain the direct result of run and this JSON.parse/type block disappears. That removes roughly 10 production lines and keeps selection in the tool already parsing the response; the regression test can return the filtered URL again.

PR #820, opened 2026-09-08
fix(scan): include workflow files in diff inventories

### LINKED ISSUE #819: Diff inventory silently excludes .github workflows and can return zero review items
## Summary

A committed diff containing only an added GitHub Actions workflow produces an empty Codex Security review inventory. The inventory helper exits successfully and the native review-item tools return zero items, even though Git identifies the changed workflow.

The diff inventory applies `path_is_excluded()` from the general ranking helper, whose `EXCLUDED_DIRS` contains `.github`. This is a deterministic selection problem before model-based review.

## Environment and verification scope

- Codex Security desktop plugin: **0.1.23**.
- Windows, OS build 10.0.26200.
- Python 3.14.7; Git 2.55.0.windows.5.
- Prompt-driven native diff scan, frozen committed base/head revisions, repository-root scope.
- Reproduced through the installed inventory helper, separately from the original native scan.
- The same filtering logic remains in public default-branch source at [`746fb84a15c906850d266dbea7c9b9e9751394bc`](https://github.com/openai/codex-security/commit/746fb84a15c906850d266dbea7c9b9e9751394bc). This is source verification; I did not install or run the current public SDK/package.

## Minimal reproduction

Use a disposable Git repository with two commits: the base has no workflow, and the head adds only a regular UTF-8 text file at `.github/workflows/example.yml`. For example:

```yaml
name: Example
on:
  workflow_dispatch:
jobs:
  example:
    runs-on: ubuntu-latest
    steps:
      - run: echo example
```

The workflow does not need to be uploaded or executed.

In PowerShell, set `$python` to the Python executable, `$pluginRoot` to the installed plugin directory, `$repo` to the disposable repository, `$base` and `$head` to its full commit SHAs, and `$inventory` to a new output path outside that repository. Then run:

```powershell
git -C $repo diff --name-status $base $head
& $python -B "$pluginRoot/scripts/generate_in_scope_files.py" --repo $repo --scope . --diff-base $base --diff-head $head --diff-mode revisions --out $inventory
$LASTEXITCODE
(Get-Item -LiteralPath $inventory).Length
```

This recipe describes the minimal input shape. The observed run used an existing real two-revision diff with one added, 51-line workflow; the example filename and contents above are synthetic.

## Observed behavior

Sanitized inventory reproduction output (workflow filename replaced):

```text
A       .github/workflows/example.yml
Recorded 0 in-scope files.
0
0
```

The final two values are the helper exit code and output file size in bytes.

During the original native diff scan:

```text
prepare_codex_security_review_items -> {"reviewItemsTotal":0}
list_codex_security_review_items    -> {"items":[]}
```

The generated `artifacts/02_discovery/in_scope_files.txt` was empty. Reproducing the helper behavior required no additional model scan or network service.

## Expected behavior

The explicitly selected changed workflow should be available to diff review. If a policy intentionally excludes it, the tool should report the excluded path and reason so the caller can distinguish an unsupported or incomplete review scope from an empty Git diff.

A nonzero inventory is not sufficient protection: applying the same filter to a mixed workflow/source diff would still omit the workflow. That mixed case follows from the source predicate; it was not independently exercised in this reproduction.

## Source trace

At the pinned public revision:

- [`generate_diff_in_scope_files()`](https://github.com/openai/codex-security/blob/746fb84a15c906850d266dbea7c9b9e9751394bc/plugins/codex-security/scripts/generate_in_scope_files.py#L206-L211) filters Git's changed paths through `path_is_excluded()` before writing inventory rows.
- [`EXCLUDED_DIRS`](https://github.com/openai/codex-security/blob/746fb84a15c906850d266dbea7c9b9e9751394bc/plugins/codex-security/scripts/generate_rank_input.py#L48-L54) includes `.github`.
- [`path_is_excluded()`](https://github.com/openai/codex-security/blob/746fb84a15c906850d266dbea7c9b9e9751394bc/plugins/codex-security/scripts/generate_rank_input.py#L288-L294) excludes a path when any component matches that set.

## Impact and suggested correction

Workflow-only changes cannot be reviewed through the generated native worklist, and operators must detect the missing coverage and arrange a separate review. This report establishes file-selection failure; it does not claim a demonstrated CI-gate bypass or a missed vulnerability.

Please distinguish targeted diff selection from broad repository-ranking exclusions, preserving explicit scope restrictions and file-safety checks. Useful regressions would cover workflow-only additions and edits, mixed source/workflow diffs, and workflow renames/deletions. Comparing expected changed paths with accounted-for reviewed or explicitly excluded paths would also catch partial omissions.

## Related work and disclosure

- #200 already identifies the same `.github` exclusion, but its description explicitly leaves the shared worklist constants unchanged. This report supplies the native diff symptom and deterministic reproduction.
- #706 concerns missing PowerShell extensions, a different predicate in the same selection path.

I searched existing issues and pull requests and found no dedicated report for this workflow-only diff failure. No private repository source, credentials, security findings, full scan bundle, or local user paths are included.

Hughhhhcoder:
## Summary

Fixes #819. Diff scans currently reuse repository-ranking exclusions, so a change under `.github/workflows` is silently omitted and a workflow-only diff can produce an empty review inventory.

## Changes

- Add a diff-specific exclusion predicate that retains `.github/workflows/**` while preserving the existing broad exclusions for repository ranking and all other excluded paths.
- Use the predicate in both canonical diff inventory generation and the legacy diff rank-input path.
- Cover committed revision diffs and local working-tree patches for both entry points, while retaining the existing assertions that unrelated excluded paths stay omitted.

## Testing

- `uv run --project plugins/codex-security --extra test pytest -q plugins/codex-security/tests/test_generate_rank_input.py plugins/codex-security/tests/test_generate_in_scope_files.py` — 87 passed.
- `uv run --project plugins/codex-security --extra test pytest -q plugins/codex-security/tests` — 1108 passed, 5 skipped, 110 subtests passed.
- Required Ruff check and format-check commands passed.
- `.github/scripts/check_plugin_source_compatibility.py` passed.

## Risk and rollout

The change only broadens explicit diff inventories to include text-code files under `.github/workflows`. Repository-wide ranking behavior and exclusions for tests, fixtures, generated files, and other `.github` content remain unchanged. No migration or rollout step is required.

## Public disclosure review

### DIFF
--- a/plugins/codex-security/scripts/generate_in_scope_files.py
+++ b/plugins/codex-security/scripts/generate_in_scope_files.py
@@ -187,7 +187,7 @@ def generate_diff_in_scope_files(
 ) -> int:
     """Reuse the existing diff selection without generating previews or duplicate worklists."""
     sys.path.insert(0, str(Path(__file__).resolve().parent))
-    from generate_rank_input import git_changed_paths, path_is_excluded
+    from generate_rank_input import git_changed_paths, path_is_diff_excluded
     from rank_preview import (
         DEFAULT_PREVIEW_BYTES,
         TEXT_CODE_EXTENSIONS,
@@ -206,7 +206,7 @@ def generate_diff_in_scope_files(
         eligible = [
             (path, status)
             for path, status in changed
-            if not path_is_excluded(path.relative_to(repository))
+            if not path_is_diff_excluded(path.relative_to(repository))
             and path.suffix.lower() in TEXT_CODE_EXTENSIONS
         ]
         revision_paths = [
--- a/plugins/codex-security/scripts/generate_rank_input.py
+++ b/plugins/codex-security/scripts/generate_rank_input.py
@@ -293,6 +293,13 @@ def path_is_excluded(path: Path) -> bool:
     return path.name.endswith((".min.js", ".map"))
 
 
+def path_is_diff_excluded(path: Path) -> bool:
+    """Apply repository exclusions while retaining changed workflow files."""
+    if path.parts[:2] == (".github", "workflows"):
+        return False
+    return path_is_excluded(path)
+
+
 def windows_stream_component(path: Path) -> str | None:
     """Return the first NTFS alternate-data-stream component."""
 
@@ -683,7 +690,7 @@ def make_diff_rank_input(args: argparse.Namespace) -> None:
     changed = [
         (path, status)
         for path, status in git_changed_paths(repo, args.base, args.head, args.mode)
-        if not path_is_excluded(path.relative_to(repo))
+        if not path_is_diff_excluded(path.relative_to(repo))
         and path.suffix.lower() in TEXT_CODE_EXTENSIONS
     ]
     revision_paths = [
--- a/plugins/codex-security/tests/test_generate_in_scope_files.py
+++ b/plugins/codex-security/tests/test_generate_in_scope_files.py
@@ -359,6 +359,7 @@ def test_diff_inventory_keeps_changed_and_deleted_source_files(tmp_path: Path) -
     write_file(repository, "app/new handler.py", b"handler = True\n")
     write_file(repository, "app/binary.py", b"\x00\xff\x01")
     write_file(repository, "tests/demo.py", b"excluded = True\n")
+    write_file(repository, ".github/workflows/ci.yml", b"name: CI\n")
     (repository / "app/évidence.py").unlink()
     git(repository, "add", ".")
     git(repository, "commit", "-qm", "change")
@@ -374,6 +375,7 @@ def test_diff_inventory_keeps_changed_and_deleted_source_files(tmp_path: Path) -
 
     assert result.returncode == 0, result.stderr
     assert output.read_text(encoding="utf-8").splitlines() == [
+        ".github/workflows/ci.yml",
         "app/new handler.py",
         "app/routes.py",
         "app/évidence.py",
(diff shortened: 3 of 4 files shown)


mldangelo-oai (commented):
Reviewed `742583f6dd95cf994c21293e6a39a30f2dfaf622`. No actionable findings from the critical review, simplification pass, and QA.

This fixes a reproduced omission: a workflow-only committed diff returns zero MCP review items on the parent commit and the exact workflow path with this patch. Both inventory entry points use the shared predicate, and repository ranking retains its existing behavior. The seven-line helper is a reasonable amount of code for that distinction; I do not see a worthwhile simplification. There are no new CLI options, dependencies, or runtime compatibility layers.

Validation:

- **135 focused Python tests passed on the PR head.** Applying the patch to current main (`fa5fe21`) passed **136**, including the module-extension regression from #859. All four modified regression tests fail on the parent because the workflow row is missing.
- Exercised both helper CLIs with real Git repositories: workflow-only and mixed diffs; workflow edits, deletions, and renames; staged/unstaged/untracked changes; `.yml`/`.yaml` and uppercase extensions; CRLF and UTF-8/UTF-16 BOMs; spaces, Unicode, and `&` in paths. Committed previews still use the selected revision after the working tree changes. Binary files and unrelated excluded paths stay excluded.
- Actual MCP prepare/list calls preserve the exact workflow path. The existing MCP inventory tests, Ruff checks, and source compatibility checks passed. The patch also passes current main's SDK build and all nine native source-checker regression tests.
- [CI completed successfully: all 30 jobs passed](https://github.com/openai/codex-security/actions/runs/34186736187), including Windows, macOS, and Ubuntu jobs. The Python job reports **1,114 passed, 109 subtests passed, and 9 skipped**; all four modified regression cases passed.

Platform coverage: the new Python regression ran on macOS locally and Ubuntu in CI. Windows CI covers the surrounding SDK/package runtime; I also checked Windows path-component semantics locally. I did not separately invoke the workflow fixture through PowerShell or cmd. The changed path predicate is shell-independent, and the MCP caller uses an argument vector with `shell: false`.

Thanks for keeping this focused and extending the existing behavior tests.

mldangelo-oai (approved):
Thanks for the focused fix and regression coverage. This closes the workflow inventory gap with a small, shared change, and the existing tests are extended cleanly.

The critical review and QA found no actionable issues. I rechecked the reviewed head (`742583f`): all 32 PR checks are passing, including the Windows, macOS, and Ubuntu CI gates. Approved — thanks for keeping this straightforward.

PR #821, opened 2026-09-08
fix(cli): parse GitLab merge request URLs

Hughhhhcoder:
## Summary

`glab mr create` writes a merge request summary followed by the merge request URL. The patch workflow currently treats the complete stdout as the URL, so GitLab patch publication can persist and print a multiline value instead of a usable link.

## Changes

- extract the standalone HTTP(S) URL from `glab mr create` output
- report a clear publication error when the output has no URL
- exercise the GitLab path with the multiline format emitted by `glab`

## Testing

- `bun test --timeout 30000 --seed=12345 tests-ts` with Bun 1.3.14: 2415 passed, 43 skipped, 0 failed
- `bun test --timeout 30000 tests-ts/cli-patch.test.ts`: 42 passed, 0 failed
- `node sdk/typescript/scripts/generate-models.cjs --check`
- TypeScript checks for the SDK and MCP app
- Prettier check
- `git diff --check`

## Risk and rollout

Low risk. GitHub publication is unchanged. GitLab publication now stores the final standalone URL from the existing CLI output; output without a URL fails through the existing pull-request creation error path.

## Public disclosure review

### DIFF
--- a/sdk/typescript/tests-ts/cli-patch.test.ts
+++ b/sdk/typescript/tests-ts/cli-patch.test.ts
@@ -1467,7 +1467,10 @@ describe("scan and patch workflow", () => {
             }
             expect(command).toBe(client);
             publicationCommands.push(args);
-            return args[1] === "create" ? url : "";
+            if (args[1] !== "create") return "";
+            return client === "glab"
+              ? `!14 fix: patch verified security findings (codex-security/patch-scan-1)\n${url}`
+              : url;
           },
         },
         {
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -5418,7 +5418,7 @@ async function publishPatchBranch(
     );
     if (!url) {
       await run("git", ["push", "--set-upstream", "origin", branch]);
-      url = await run(
+      const createOutput = await run(
         command,
         gitlab
           ? [
@@ -5449,6 +5449,7 @@ async function publishPatchBranch(
               body,
             ],
       );
+      url = gitlab ? glabMergeRequestUrl(createOutput) : createOutput;
     }
     stderr.write(
       `${gitlab ? "Merge" : "Pull"} request: ${safePatchText(url)}\n`,
@@ -5460,6 +5461,14 @@ async function publishPatchBranch(
     );
     throw error;
   }
+}
+
+function glabMergeRequestUrl(output: string): string {
+  for (const line of output.split(/\r?\n/u).reverse()) {
+    const candidate = line.trim();
+    if (/^https?:\/\/\S+$/u.test(candidate)) return candidate;
+  }
+  throw new CodexSecurityError("glab did not return a merge request URL.");
 }
 
 function patchRemoteHost(remote: string): string | undefined {


sdk/typescript/tests-ts/cli-patch.test.ts
```diff
@@ -1467,7 +1467,10 @@ describe("scan and patch workflow", () => {
             }
             expect(command).toBe(client);
             publicationCommands.push(args);
-            return args[1] === "create" ? url : "";
+            if (args[1] !== "create") return "";
+            return client === "glab"
+              ? `!14 fix: patch verified security findings (codex-security/patch-scan-1)\n${url}`
+              : url;
```
mldangelo-oai, line 1473: **[P2] Reproduce piped glab output before adding a parser**

This mock supplies terminal-formatted output, but `runRepositoryCommand` captures stdout with `execFile`. glab therefore sees a pipe: both v1.64.0 and v1.117.0 pass `io.IsaTTY` to `DisplayMR`, whose non-TTY branch returns only `mr.WebURL` ([create call](https://gitlab.com/gitlab-org/cli/-/blob/v1.117.0/internal/commands/mr/create/mr_create.go#L870), [formatter](https://gitlab.com/gitlab-org/cli/-/blob/v1.117.0/internal/commands/mr/mrutils/mrutils.go#L109-115)). I verified this with both official release binaries, this PR's complete `mr create` argument list, and a local synthetic GitLab API: each produced exactly one URL on stdout. Feeding those captured outputs through the built parent CLI already succeeds. The seven new parent failures come from injecting the terminal format here, rather than reproducing the production invocation. Please identify the glab version/configuration that actually produces multiline stdout through this subprocess path and use that capture as the regression. Otherwise, I would remove `glabMergeRequestUrl` and keep the existing direct assignment; the extra parsing and failure branch currently lack a demonstrated use case.

mldangelo-oai (commented):
Reviewed `5c8d3b3d63f1ae21383d7ccf3bedf2ea9a9b3c24`.

**One finding below about whether this parser is needed.** The implementation handles the supplied multiline fixture correctly, but released glab binaries already produce a plain URL when invoked through the existing subprocess path. I would establish a real failing invocation before keeping this extra code.

### Independent QA

- **205 focused tests passed** across the patch workflow, CLI, and patch TUI suites; no failures or skips.
- Applying the patch to main at `b45e332f44da267630a060853e8b195e8a044e0e` passed all **42 patch-workflow tests**.
- On the parent, **35 passed and seven failed** with the changed fixture. All seven failures are the mocked GitLab multiline cases; actual output captured from released glab versions works on the parent.
- Built Node CLI checks with real Git passed **11 publication cases and 11 repeat-publication checks** on each revision. On the PR, LF/CRLF output, surrounding whitespace, HTTP URLs with ports, and subgroup paths resolve correctly. Missing/non-standalone URLs produce the clear error and preserve the saved patch commit. Retrying an already-created MR avoids another push or creation.
- Independently ran official glab **1.64.0** and **1.117.0** release binaries, with verified release checksums, against a loopback GitLab API fixture using the production create arguments and `execFile` pipes. Both returned just the URL.
- Local type checking, formatting, plugin assembly, and SDK builds passed. [CI passed](https://github.com/openai/codex-security/actions/runs/34206885253): **32 successful PR checks**. The seven GitLab regression cases passed in native Windows, macOS, and Linux JUnit reports.

No additional runtime defect or Python-to-Node compatibility layer surfaced. The simplification opportunity is to avoid adding this parser unless the reported subprocess failure can be reproduced. Local released-glab execution was on macOS; Windows/Linux coverage is CI with mocked provider responses, not a live GitLab service.

Hughhhhcoder:
Thanks — the finding is correct. I rechecked historical release source to look for a supported failing case. `glab` v1.20 emitted multiline output even through a pipe, but that release cannot run the existing publication flow because `mr list` did not yet support the current `--output json` / `--jq` invocation. By v1.40, `DisplayMR` already returned only `WebURL` for non-TTY output; v1.64 behaves the same, and v1.117 supports the current `--jq` path and is also URL-only under `execFile`. I could not establish a released `glab` version or configuration that reaches this path and produces multiline stdout.

So the mocked terminal-formatted output is not a valid production regression, and I agree the parser should not be kept. I will not push further code changes here. Thank you for the careful reproduction and QA.

PR #822, opened 2026-09-08
fix: show repository names in the findings dashboard

akrupa-appto:
## Summary

the findings dashboard shows opaque repository IDs because imports drop the scan's display name. this makes the repository column and filter hard to use.

the publisher now sends the name, and the dashboard displays it while retaining the original IDs.

## Changes

adds optional `repositoryName` metadata to the existing bulk import endpoint and stores it separately from finding identities. names appear in the list, inspector, and repository filter, and text search matches them. imports without a name keep the existing label or fall back to the ID.

## Testing

- 20 focused findings-server and custom-publication tests passed, including the new regression test.
- type checks, formatting, ruff, and portable plugin source checks passed.
- 275 Python workbench and migration tests passed.
- the full TypeScript suite reported 2,331 passes, 41 skips, and three failures. all three also reproduced on the unchanged base: terminal link rendering, issue-link sanitization, and a launcher assertion affected by the local temporary-directory environment. no unrelated fixes are included.
- built the Docker image and verified the additive migration preserves finding contents and passes SQLite integrity checking.
- verified the rendered dashboard in a headless browser. the screenshot below uses only synthetic findings. it is committed as a documentation asset because uploading an attachment to the upstream repository requires write access.

repository names on synthetic imported findings

## Risk and rollout

the import field is optional and requires `repositoryId`. the migration adds a display-name table without changing existing findings or repository associations. an empty finding batch can update a label without calling the embedding provider. no authentication or network exposure changes are included.

## Public disclosure review


akrupa-appto:
@coderabbitai review

PR #836, opened 2026-09-08
refactor(plugin): add wide Windows candidate file operations

kmbroai:
## Summary

Add the Windows file operations needed to move candidate normalization into the Node helper while preserving UTF-16 paths and existing filesystem behavior.

## Changes

- Add native `read_link` and typed file reads, exclusive iterable writes, replacement, and deletion using the existing handle binding.
- Resolve missing paths through their existing ancestors, preserving literal link targets and avoiding recursion for deeply nested paths.
- Share the existing Windows parser between the adapter and policy helper.
- Preserve the parent of missing alternate data streams when reattaching filename components.
- Resolve relative paths using native Windows normalization, preserve bare UNC share roots, and report access denied when reading a directory as a file.
- Extend portable path tests and the Windows process proof with raw filenames, dangling and looping links, replacement cleanup, and long missing paths.

## Testing

- Rust format and Clippy passed for Linux x64, Windows x64, and Windows arm64, including the Windows proof launcher.
- SDK/MCP types, CI-tool compilation, plugin build, formatting, Ruff, and portable source checks passed.
- Portable Windows-path tests: 24 passed on Node 20.0.0 and 22.13.0, including whitespace-only paths and UNC share roots. The four stream-path regressions failed before the earlier fix and pass afterward.
- Policy/helper launcher tests: 34 passed, 5 Windows-only skips.
- The preceding head passed hosted Windows x64 and arm64 proofs. The updated proof also checks whitespace-only paths and directory-read errors; these new assertions await hosted Windows execution.

## Risk and rollout

This is the prerequisite for the stacked candidate-normalizer migration. Existing strict path resolution remains the default. Non-strict resolution supports missing outputs and can retain unresolved components; callers must check containment independently. The new native export must ship with its matching typed adapter; the existing native build and packaging workflow supplies both.

## Public disclosure review


plugins/codex-security/native/windows-files.mts
```diff
@@ -67,24 +100,89 @@ export function windowsFileSystem(native: WindowsBinding) {
...
+      while (true) {
+        try {
+          value = finalPath(value);
+          break;
+        } catch (error) {
+          if (strict) throw error;
+          const winerror = (error as { winerror?: number }).winerror;
+          // Match pathlib's non-strict Windows resolution errors.
+          if (
+            ![
```
mldangelo-oai, line 152: **[P2] Do not turn an unresolved ancestor into a canonical path**

This copies `pathlib`'s broad non-strict error policy, but the candidate-normalizer contract is much narrower: the fallback is only for a deleted inventory row after strict resolution reports a missing path. Once this loop reaches an ancestor that returns `ERROR_ACCESS_DENIED` (5), `ERROR_SHARING_VIOLATION` (32), a bad-network error, `ERROR_INVALID_NAME`, or another opaque error in this list, we no longer know whether that ancestor is an ordinary directory or a reparse point elsewhere. The loop currently swallows that fact, climbs higher, and appends the unresolved components as though they had been canonicalized. For example, a deterministic binding where `C:\repo\opaque\deleted.py` returns 3, `C:\repo\opaque` returns 5, and `C:\repo` resolves successfully returns `C:\repo\opaque\deleted.py`; a later containment check would accept ancestry it never inspected.

Please keep this operation scoped to genuinely missing components and reject any other unresolved existing ancestor. The smaller shape is: validate the repository-relative spelling, walk upward only while `lstat` says the component is absent, require strict `realpath` to succeed for the first physically existing ancestor, append the missing suffix, and then check containment. A dangling symlink is an existing object, not a deleted Git path. The ENOENT-only loop in `canonicalWindowsCreationPath` (`sdk/typescript/src/multiscan.ts`) is a useful precedent, with the additional `lstat` distinction needed here. Keeping this at the candidate caller should also let us delete the generic CPython error table, manual read-link loop, `seen` state, and most of the new path-join machinery instead of preserving a Python filesystem quirk.

plugins/codex-security/native/windows-files.mts
```diff
@@ -5,6 +5,39 @@ import { windowsFlags as flags } from "./windows-flags.mjs";
 export const widePath = (path: string): Buffer => Buffer.from(path, "utf16le");
 export const pathText = (path: Buffer): string => path.toString("utf16le");
 
+export function windowsParts(value: string): [string, string, string] {
```
mldangelo-oai, line 8: **[P3] Delete the second Windows path parser**

`windowsParts` and `windowsJoin` are byte-for-byte copies of the private helpers at `plugins/codex-security/mcp-app/src/helpers/resolve-security-md.ts:39-70`, and that module already imports `windows-files.mjs`. Please import these exports there and remove the older copy in this PR (or move the pair into one small shared Windows-path module). That is about 32 net lines deleted immediately, with no behavior change, and avoids having two root/UNC parsers whose bug fixes can drift.

mldangelo-oai:
## Critical review with GPT-6 Astra at ultra

Reviewed `06112a516431a9e12de644e8621b0176946da0a5` against merge base `b39e2457b317f619ff9b14f96a7b2415703866be`, after the [combined-stack QA](https://github.com/openai/codex-security/pull/843#issuecomment-5610091473).

**No additional inline finding survived validation.** The native file operations are a useful prerequisite for the helper migration: raw Windows paths, exclusive writes, replacement and cleanup have concrete consumers. The duplicate `windowsParts`/`windowsJoin` implementation has been removed, satisfying the [earlier parser comment](https://github.com/openai/codex-security/pull/836#discussion_r3963362496).

The main simplification target remains the [non-strict resolver's broad Python error policy](https://github.com/openai/codex-security/pull/836#discussion_r3963362346). That discussion needs two corrections:

- The original denied-open example is rejected by the actual downstream file-identity check. I verified this through the normalization command, so that example does not demonstrate accepted containment.
- Missing output files and rank-directory comparisons also use non-strict resolution. Preserve those workflows and deliberate dangling-output-link behavior when narrowing the tolerated errors; deleted inventory is not the entire contract.

Focus the follow-up on replacing the generic 15-error whitelist with a defined missing-path contract and propagating opaque lookup failures. A separate mocked final-name-failure case still accepts an unresolved row, but this has not been demonstrated as a real Windows boundary bypass. Keep that evidence distinction in the existing thread.

**Targeted verification:** 24 adapter tests, 34 policy/launcher tests, 15 independent I/O/consumer checks and 9 portable-checker tests passed; five Windows-only tests skipped on Linux. I independently replayed the seven consumer assertions. Plugin build, MCP/CI-tool types, portable source and Ruff checks passed. Initial fixture failures reproduced on main and passed after correcting the fixture environment.

Exact-head hosted Windows/macOS CI, including Windows native and PowerShell checks, was inspected. Local Windows behavior was modeled. The combined-tree platform gap is documented in the shared QA comment.

plugins/codex-security/native/windows-files.mts
kmbroai: Keeping the non-strict resolver for this migration. As your follow-up confirms, the denied-open example is rejected by the downstream file-identity check, and the callers also rely on missing output paths, rank-directory comparisons, and deliberate dangling-output-link behavior. An ENOENT-only replacement would change those contracts; the mocked final-name failure has not established a real Windows boundary bypass. The strict input-identity and containment checks remain in place. Resolving with that narrower assessment and the existing compatibility behavior.

plugins/codex-security/native/windows-files.mts
kmbroai: Already fixed in the current head: resolve-security-md imports windowsParts and windowsJoin from windows-files.mjs, and its duplicate implementations are removed. I rechecked both modules. Resolving this thread.

kmbroai:
Addressed the general review and both inline threads. The duplicate Windows parser is already removed. I retained the non-strict resolver for the missing-output, directory-comparison, and dangling-output-link contracts described in the follow-up, and recorded that the denied-open example is rejected by the downstream identity check; the mocked final-name case is not evidence of a real Windows boundary bypass. Both threads have replies and are resolved.

Updated head: `06112a516431a9e12de644e8621b0176946da0a5`. [Hosted CI](https://github.com/openai/codex-security/actions/runs/34392609626) passes on this head, including Linux/macOS/Windows tests, native proofs, installed-package checks, plugin source contracts, types, and formatting. Codex Security Review also passes. The combined stack at `5bd1d1909ad99d2749fe8d3ec49a9a144ae2bac0` includes this head and has also passed platform CI.

PR #837, opened 2026-09-09
refactor(plugin): port candidate normalization to TypeScript

kmbroai:
## Summary

Replace the Python candidate normalizer with the bundled Node helper, including its MCP artifact-discovery caller. Remove the Python normalizer and its owned tests.

## Changes

- Preserve candidate filtering, deleted-file scope, canonical artifact bytes and IDs, and atomic replacement.
- Use JSON.parse with bounded integer line checks and deterministic input-path sorting. Integral JSON spellings such as 1.0 and 1e0 are accepted.
- Reuse the Windows file adapter and native identity checks; preserve raw Unix filenames and reject malformed UTF-8 JSON.
- Update the helper launcher, MCP caller, skills, package inventory, and tests; remove the Python normalizer.

## Testing

- Combined stack tip `ba12023`: both full SDK runs passed 2,656 tests with 50 skips and zero failures (seeds 12345 and 4201856736).
- The six affected helper suites passed 213 tests with four Windows-only skips; all 23 MCP test processes passed.
- CI compilation, plugin build, types, formatting, Ruff, portable source checks, and nine source-checker tests passed.
- Rust formatting and cross-target Clippy for the Windows x64 native proof passed. Local execution was Linux; hosted CI supplies platform runtime coverage.

## Risk and rollout

Stacked on the Windows file-operations prerequisite. Invocation becomes `launch_codex_security_mcp[.cmd] --helper normalize-candidates`; arguments and defaults are preserved. Integral line spellings are accepted, and malformed-JSON wording or the first invalid input reported can change. Canonical output, line bounds, containment, and atomic replacement remain enforced.

## Public disclosure review

- [x] No customer, partner, prospect, or user identities, data, or identifying details are included.
- [x] No credentials, personal data, private source, scan findings, or nonpublic links or tickets are included.
- [x] I reviewed the branch name, title, description, commits, changes, comments, logs, screenshots, attachments, and links for public disclosure.

PR #838, opened 2026-09-09
refactor(plugin): port patch-risk validation to TypeScript

kmbroai:
## Summary

Replace the Python patch-risk assessment validator with the bundled TypeScript helper.

## Changes

- Preserve schema checks, assessment invariants, file/stdin input, numeric values, duplicate-key detection, and property order.
- Use JSON.parse to validate string escapes and check schema items before Set-based string uniqueness; remove recursive aggregate equality.
- Update the skill, package inputs, and helper dispatch; remove the Python validator and its owned tests.

## Testing

- Combined stack tip `ba12023`: both full SDK runs passed 2,656 tests with 50 skips and zero failures (seeds 12345 and 4201856736).
- The six affected helper suites passed 213 tests with four Windows-only skips; all 23 MCP test processes passed.
- CI compilation, plugin build, types, formatting, Ruff, portable source checks, and nine source-checker tests passed.
- Rust formatting and cross-target Clippy for the Windows x64 native proof passed. Local execution was Linux; hosted CI supplies platform runtime coverage.

## Risk and rollout

Stacked on candidate normalization. Invocation becomes `launch_codex_security_mcp[.cmd] --helper validate-patch-risk-assessment <assessment.json>`; `-` still reads stdin. Schema acceptance and decision rules are unchanged. Invalid-item errors now precede uniqueness errors, and malformed-string diagnostic wording comes from JSON.parse.

## Public disclosure review


plugins/codex-security/mcp-app/src/helpers/contract-schema.ts
```diff
@@ -0,0 +1,192 @@
...
+  if (Array.isArray(value)) {
+    if (schema.minItems !== undefined && value.length < schema.minItems)
+      fail("array has too few items");
+    if (schema.maxItems !== undefined && value.length > schema.maxItems)
+      fail("array has too many items");
+    if (
+      schema.uniqueItems === true &&
+      value.some((item, index) =>
+        value.slice(0, index).some((other) => equal(item, other)),
+      )
```
mldangelo-oai: [P3] Check string items before using Set uniqueness

Validate `schema.items` first, then use `new Set(value).size !== value.length`. The only consumer is the patch-risk assessment schema: its `uniqueItems` nodes resolve to strings, and all 13 `const`/`enum` rules contain scalar values. Recursive object/array equality and copying/scanning every preceding array prefix are unnecessary for this contract.

This lets us delete the [aggregate branches of `equal`](https://github.com/openai/codex-security/blob/b408067753cb908d97a9de20e1339b766cff5cf6/plugins/codex-security/mcp-app/src/helpers/contract-schema.ts#L51-L65). A disposable version removes 19 lines. All 931 assessment cases retained the Python baseline's accept/reject result, independently replayed on Node 22.13, 24, 26 and Bun. For invalid `autoMergeExclusions: [1, 1.0]`, the error becomes an invalid-item error instead of a duplicate-items error; both reject. Keep numeric constant comparison and duplicate-string rejection, and remove assertions that require the old error precedence for invalid members.

plugins/codex-security/mcp-app/src/helpers/python-json.ts
```diff
@@ -0,0 +1,179 @@
...
+  const take = () => tokens[index++]?.[0];
+  function expect(token: string): void {
+    if (tokens[index]?.[0] !== token) error(`Expecting '${token}' delimiter`);
+    index++;
+  }
+  function string(token: string, start: number): string {
+    const unterminated = token.length === 1;
+    const contents = unterminated ? source.slice(start) : token;
+    const end = contents.length - (unterminated ? 0 : 1);
+    for (let offset = 1; offset < end; offset++) {
```
mldangelo-oai: [P3] Let JSON.parse validate string escapes

This escape scan duplicates the validation performed by `JSON.parse(token)` at line 69. Replace the scan with that call and wrap parse failures through the existing `JsonSyntaxError` path. Raw control characters, invalid escapes and unterminated strings still fail. Reproducing Python's exact error wording and offset does not justify a second string parser.

A disposable replacement removes 15 lines. All 1,182 direct string inputs retained their acceptance and decoded values; I independently replayed the checks on Node 22.13, 24, 26 and Bun. The full-stack serializer also retained bytes in 262 comparisons across 131 fixtures. Keep malformed-input rejection, duplicate-key detection, numeric handling and key order; adjust diagnostic-only assertions. This targets the extra scan added here, separately from the candidate-only parser simplification discussed on #837.

mldangelo-oai:
Reviewed `b408067753cb908d97a9de20e1339b766cff5cf6` against its recorded #837 base using **GPT-6 Astra at ultra effort**, with the PR deslop and critical review passes.

Assessment validation works through the bundled helper without Python. I found no correctness regression in the covered contracts and left two nonblocking simplification comments: remove the redundant JSON string escape scan, and validate string items before Set-based uniqueness. Disposable versions remove 34 lines; malformed-input diagnostics change intentionally.

Verification beyond the [complete-stack comparison with main](https://github.com/openai/codex-security/pull/843#issuecomment-5610091473):

- 85 focused tests passed; 2 Windows-only skips.
- 931 assessment cases matched the Python baseline; 1,182 string inputs checked parsing acceptance and values. The original and both proposed simplifications passed independent replays on Node 22.13, 24, 26 and Bun.
- 84 launcher invocations passed across the Python baseline and Node 22.13/24/26, preserving input files, modes and symlinks.
- 262 downstream serialization comparisons preserved artifact bytes. Builds, types, formatting and portable source checks passed.

Local execution was Linux. [Exact-head CI](https://github.com/openai/codex-security/actions/runs/34297433179) supplies separate Windows/macOS/native coverage; it does not cover the synthetic combined stack. Shared numeric, ordering and raw-path behavior has real consumers, so those parts need a separate proof before deletion.

plugins/codex-security/mcp-app/src/helpers/contract-schema.ts
kmbroai: Fixed in afd609b. Schema items are now checked before Set-based uniqueness. The current uniqueness rules resolve to strings, and const/enum rules are scalar, so the recursive aggregate equality branches are removed. Numeric constant comparison and duplicate-string rejection remain; tests now expect invalid-item errors for invalid members.

plugins/codex-security/mcp-app/src/helpers/python-json.ts
kmbroai: Fixed in afd609b. Removed the duplicate escape scan. JSON.parse validates string tokens, with failures wrapped in the existing JsonSyntaxError path. Malformed strings still fail, and duplicate-key handling, numeric values, property order, and downstream serialization remain covered.

kmbroai:
Addressed both simplification comments: the duplicate string-escape scan and recursive aggregate equality are removed, and validated string items use Set uniqueness. The shared numeric representation, duplicate-key handling, and property order remain for their actual consumers. The updated PR description records the diagnostic changes.

Updated head: `afd609bce44577068eef895f27750fa62dc08bb3`. [Hosted CI](https://github.com/openai/codex-security/actions/runs/34544008024) passes on this head, including Linux/macOS/Windows tests, native proofs, installed-package checks, plugin source contracts, types, and formatting. Codex Security Review also passes. The Windows Node 24 package job passed on an unchanged-commit retry after npm install timed out. The combined stack at `5bd1d1909ad99d2749fe8d3ec49a9a144ae2bac0` includes this head and has also passed platform CI.

PR #839, opened 2026-09-09
refactor(plugin): port deep-review input to TypeScript

kmbroai:
## Summary

Move deep-review worklist creation to the bundled TypeScript helper.

## Changes

- Preserve row selection, ordering, output formatting, argument grammar, and raw-path behavior.
- Read inputs directly through the shared file helper; filesystem errors reach the existing failure handler before output is written.
- Update helper routing and workflow instructions; remove the Python commands and tests.

## Testing

- Combined stack tip `ba12023`: both full SDK runs passed 2,656 tests with 50 skips and zero failures (seeds 12345 and 4201856736).
- The six affected helper suites passed 213 tests with four Windows-only skips; all 23 MCP test processes passed.
- CI compilation, plugin build, types, formatting, Ruff, portable source checks, and nine source-checker tests passed.
- Rust formatting and cross-target Clippy for the Windows x64 native proof passed. Local execution was Linux; hosted CI supplies platform runtime coverage.

## Risk and rollout

Stacked on patch-risk validation. `copy-deep-review-input` and `select-deep-review-input` move to `launch_codex_security_mcp[.cmd] --helper`. Options, accepted argument grammar, and defaults are preserved. Missing or invalid paths now report the underlying read error.

## Public disclosure review

- [x] No customer, partner, prospect, or user identities, data, or identifying details are included.
- [x] No credentials, personal data, private source, scan findings, or nonpublic links or tickets are included.
- [x] I reviewed the branch name, title, description, commits, changes, comments, logs, screenshots, attachments, and links for public disclosure.

PR #840, opened 2026-09-09
refactor(plugin): add raw Unix directory entries

kmbroai:
## Summary

Expose raw Unix directory entries through the existing native binding for ranking and the remaining workbench migration.

## Changes

- Expose raw directory-entry names with optional cached entry types, symlink flags, and per-entry filesystem errors.
- Support names-only enumeration without child metadata queries.
- Keep the native API and its proofs as a prerequisite for ranking and the remaining workbench traversal/copy migration.

## Testing

- Combined stack tip `ba12023`: both full SDK runs passed 2,656 tests with 50 skips and zero failures (seeds 12345 and 4201856736).
- The six affected helper suites passed 213 tests with four Windows-only skips; all 23 MCP test processes passed.
- CI compilation, plugin build, types, formatting, Ruff, portable source checks, and nine source-checker tests passed.
- Rust formatting and cross-target Clippy for the Windows x64 native proof passed. Local execution was Linux; hosted CI supplies platform runtime coverage.

## Risk and rollout

Stacked on deep-review input. The native export and typed declaration must ship together through the existing native build and packaging workflow. This adds no public CLI surface.

## Public disclosure review

### DIFF
--- a/plugins/codex-security/native/src/unix.rs
+++ b/plugins/codex-security/native/src/unix.rs
@@ -1,8 +1,10 @@
 use napi::bindgen_prelude::Buffer;
 use napi_derive::napi;
 use std::{
-    ffi::{CStr, CString},
-    io,
+    ffi::{CStr, CString, OsStr},
+    fs, io,
+    os::unix::ffi::OsStrExt,
+    path::Path,
 };
 
 #[napi(object)]
@@ -33,6 +35,55 @@ fn retry_eintr(mut operation: impl FnMut() -> i32) -> SyscallResult {
 
 fn path(value: Buffer) -> napi::Result<CString> {
     CString::new(value.as_ref()).map_err(|_| napi::Error::from_reason("Path contains a NUL byte"))
+}
+
+#[napi(object)]
+pub struct DirectoryEntry {
+    pub name: Buffer,
+    pub is_directory: bool,
+    pub is_symbolic_link: bool,
+    pub errno: i32,
+}
+
+#[napi(object)]
+pub struct DirectoryResult {
+    pub value: Vec<DirectoryEntry>,
+    pub errno: i32,
+}
+
+#[napi]
+pub fn directory_entries(name: Buffer, with_types: bool) -> napi::Result<DirectoryResult> {
+    let name = path(name)?;
+    let entries = fs::read_dir(Path::new(OsStr::from_bytes(name.to_bytes()))).and_then(|entries| {
+        entries
+            .map(|entry| {
+                let entry = entry?;
+                let mut value = DirectoryEntry {
+                    name: entry.file_name().as_bytes().to_vec().into(),
+                    is_directory: false,
+                    is_symbolic_link: false,
+                    errno: 0,
+                };
+                if with_types {
+                    match entry.file_type() {
+                        Ok(kind) => {
+                            value.is_directory = kind.is_dir();
+                            value.is_symbolic_link = kind.is_symlink();
+                        }
+                        Err(error) => value.errno = error.raw_os_error().unwrap(),
+                    }
+                }
+                Ok(value)
+            })
+            .collect::<io::Result<Vec<_>>>()
+    });
+    Ok(match entries {
+        Ok(value) => DirectoryResult { value, errno: 0 },
+        Err(error) => DirectoryResult {
+            value: Vec::new(),
+            errno: error.raw_os_error().unwrap(),
+        },
+    })
 }
 
 #[napi]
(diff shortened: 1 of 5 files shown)


plugins/codex-security/native/src/unix.rs
```diff
@@ -35,6 +37,55 @@ fn path(value: Buffer) -> napi::Result<CString> {
...
+#[napi(object)]
+pub struct DirectoryResult {
+    pub value: Vec<DirectoryEntry>,
+    pub errno: i32,
+}
+
+#[napi]
+pub fn directory_entries(name: Buffer, with_types: bool) -> napi::Result<DirectoryResult> {
+    let name = path(name)?;
+    let entries = fs::read_dir(Path::new(OsStr::from_bytes(name.to_bytes()))).and_then(|entries| {
```
mldangelo-oai, line 57: [P3] Use Node's names-only API instead of adding a native export

Retire this export and use `readdirSync(encodePosixPath(directory), { encoding: "buffer" }).map(decodePosixBytes).filter(matches)` in #841's `shardNames`, preserving its existing error catch and Windows branch. That is the only production consumer in the stack and always passes `false`; none of the type flags or per-entry metadata errors is consumed.

Names-only enumeration preserves raw parent/name bytes without querying child metadata. The replacement passed permission, symlink and forced-unknown-type probes on Node 22.13, 24 and 26; both implementations passed 100 shard/pool tests, and 52 synthetic helper invocations preserved artifact bytes and receipt hashes. An independent replay also passed all 100 tests with the earlier native binary, which lacks this export.

Enumeration order intentionally changes which malformed shard is reported first when several are invalid. Both implementations reject before writing; valid assignments and merges sort names before producing artifacts. There is no need to reproduce the unused typed-directory contract with another wrapper.

Remove the new Rust structs/function, typed binding, directory proof and export-specific docs/package assertion together: 212 net lines in this PR, plus four in the downstream caller. Keep the existing native primitives and ranking tests that exercise raw filenames.

mldangelo-oai:
Reviewed `63534280b1de796dfcde43ab04cdf88cbab18820` against its #839 base using **GPT-6 Astra at ultra effort**, including the PR deslop and critical review passes.

I found no demonstrated correctness or security regression. I left one nonblocking simplification comment: remove this new native directory API and use Node's names-only Buffer API in the subsequent ranking caller. The only production caller discards all type metadata. Removing the export, types, proof and package coupling cuts 212 net lines; the coordinated caller change cuts four more.

Verification beyond the [complete-stack comparison with main](https://github.com/openai/codex-security/pull/843#issuecomment-5610091473):

- Exact-head native build/proof, plugin build and bundled export assertion passed.
- Raw-name, permission, symlink and forced-unknown-type checks passed on Node 22.13, 24 and 26.
- Each implementation passed 100 shard/pool tests, with one Windows-only skip. Independent artifact/receipt comparisons covered 52 helper invocations.
- An independent 100-test replay loaded the earlier native binary in 303 helper processes; that binary has no directory-enumeration export.
- Portable source checks, formatting, Clippy and the focused policy suite passed.

Only first-invalid-name diagnostic precedence changes intentionally; valid artifact bytes and receipt hashes remain unchanged in the comparisons. Local OS execution was Linux. [Exact-head CI](https://github.com/openai/codex-security/actions/runs/34300007596) provides separate platform/native evidence; the synthetic combined stack still needs platform CI after restacking. The shipped helper uses Node, so direct Bun directory-object differences do not justify this native addition.

plugins/codex-security/native/src/unix.rs
kmbroai: Retaining the shared native directory API. The remaining workbench migration has concrete traversal and directory-copy consumers that use raw filename bytes, entry types, symlink flags, and per-entry metadata errors; those consumers are already implemented locally. Names-only readdir is sufficient for this ranking caller, but deleting the export would remove their prerequisite and require reintroducing it in the next stack. I am keeping the existing native proof and raw-filename coverage with that API.

kmbroai:
Retained the native directory API for the implemented workbench traversal and directory-copy consumers in the remaining migration. They use raw names, type/symlink metadata, and per-entry errors, so the export and its proofs remain a shared prerequisite. The inline thread is resolved with that rationale, and the PR description now explains these consumers.

Updated head: `04394804f360fa5e1a424d96bd76892282bdae13`. [Hosted CI](https://github.com/openai/codex-security/actions/runs/34544617818) passes on this head, including Linux/macOS/Windows tests, native proofs, installed-package checks, plugin source contracts, types, and formatting. Codex Security Review also passes. The Windows shards passed on an unchanged-commit retry after PowerShell and Python startup-probe failures. The combined stack at `5bd1d1909ad99d2749fe8d3ec49a9a144ae2bac0` includes this head and has also passed platform CI.

PR #848, opened 2026-09-09
fix(plugin): assemble file-authored scan identity envelopes

### LINKED ISSUE #790: Supported recovery for terminal unsealed scan missing findings/coverage scanId envelopes
## Summary

Codex Security plugin `v0.1.22` produced a terminal local scan directory whose semantic scan artifacts exist, but the canonical identity envelopes were not fully assembled:

- `scan-manifest.json` contains a valid canonical `scan.id`;
- `findings.json` exists but has no `scanId`;
- `coverage.json` exists but has no `scanId`;
- the scan is unsealed (`sealedAt` absent) and has no `report.md`;
- the terminal finalizer fails closed with:

```text
findings.scanId: must match manifest scan id
```

The artifacts have not been manually modified.

## Investigation so far

The installed `v0.1.22` finalizer correctly requires the manifest, findings, and coverage scan identities to agree. It does not synthesize missing `findings.scanId` / `coverage.scanId` without the trusted internal completion binding.

The installed CLI exposes no documented terminal resume/reassembly command for this state. Manually adding the IDs or calling undocumented internal completion-binding APIs would bypass the intended evidence chain, so we have not done that.

We also reviewed the newer explicit `recover-scan-results` path. As documented in later releases, it applies to registered failed, non-canceled workbench scans with validated retained checkpoints; this local terminal scan is not an authoritative registered workbench row and does not appear eligible for that recovery path.

## Policy / recovery question

For this exact condition, what is the supported operator recovery path?

1. **Deterministic resume/reassembly of the existing unsealed scan** — if supported, please provide the public/operator-supported command or procedure for reconstructing the canonical findings/coverage identity envelopes from trusted state; or
2. **Preserve the incomplete scan and start a fresh scan** — if that is the intended policy, please confirm that the unsealed directory should remain preserved as incomplete evidence and that a new scan with a new scan ID/directory is the supported next step.

If a newer plugin version supports recovery of this specific `v0.1.22` terminal state, please identify the minimum compatible version and whether it can safely adopt/recover the existing terminal directory.

We are intentionally not retrying the finalizer, rewriting the JSON envelopes, deleting the incomplete scan, or starting a replacement scan until the supported recovery contract is clear.

IvanCaceres:
## Summary

Fixes #790. File-authored scans can reach finalization with semantic findings and coverage but without their top-level scan IDs. The terminal finalizer has no host completion binding and correctly rejects those files.

Standard scan instructions in both v0.1.22 and the current base delegate canonical file construction to the model, without a shared envelope assembly step. They verify file existence but do not explicitly bind both artifact IDs before completion. This change assembles all three documents from the producer's single resolved scan identity before the first write.

The reported artifact state and exact finalizer failure were reproduced with synthetic data against both `npm-v0.1.22` and the current finalizer. Fresh helper-authored envelopes then sealed successfully under both versions. This reproduces the failure condition, not the original model conversation, and does not adopt already-incomplete directories.

## Changes

- Add a bundled internal Python assembly helper that wraps new scan metadata, finding records, and coverage fields into all three canonical envelopes, carrying the same scan ID into each. It performs no filesystem reads, recovery, host registration, or sealing.
- Route file-authored Standard and terminal diff skills through the helper, with a shared executable recipe and an identity check before completion. Preserve SDK-owned completion.
- Distinguish compact workbench coverage from full file-authored coverage and clarify that a terminal scan still needs a local identity even without host registration.
- Reproduce the missing-envelope error before exercising the shipped assembly recipe through the real terminal finalizer for repository, scoped-path, and diff modes, with and without findings. Verify that the incomplete directory remains byte-for-byte unchanged while the fresh scan seals.
- Preserve regression coverage for missing artifact IDs and unchanged incomplete files. Retain recovery guidance for legacy directories: preserve the original evidence and start a fresh scan.

## Testing

Passed:
- `python -m ruff check --config plugins/codex-security/pyproject.toml plugins/codex-security`
- `python -m ruff format --check --config plugins/codex-security/pyproject.toml plugins/codex-security`
- `pnpm --dir sdk/typescript run build:ci`
- `node .github/scripts/check_plugin_source_compatibility.mjs`
- `node --test .github/scripts/test_check_plugin_source_compatibility.mjs` (8 passed, 1 skipped)
- `python -m pytest plugins/codex-security/tests/test_scan_draft.py plugins/codex-security/tests/test_plugin_contract.py plugins/codex-security/tests/test_scan_contract_examples.py -q` (31 passed; 44 subtests passed)
- `python -m pytest plugins/codex-security/tests/test_finalize_scan_contract.py -q -k 'terminal_missing_scan_ids or completion_binding'` (4 passed; 3 subtests passed)
- `git diff --check`

A separate historical reproduction extracted scripts and schemas from `npm-v0.1.22` into a temporary directory. Both the tagged and current finalizers rejected synthetic unsealed artifacts missing both artifact IDs with `findings.scanId: must match manifest scan id`, without changing files. Fresh outputs from the new helper sealed and generated `report.md` under both finalizers. No completion binding was supplied and the incomplete artifacts remained unchanged.

The schema-example suite initially could not collect because the local environment lacked jsonschema. After installing the repository-pinned jsonschema test dependency in an isolated local dependency directory, the combined suite passed.

The full finalizer test module was also run on Windows during this PR: 138 passed, 15 failed, 3 skipped, and 1 teardown error. An unchanged copy of the same module from the base commit reproduced the same 15 failures and teardown error (137 passed). Existing failures involve Windows symlink privileges, file handling/error expectations, and text decoding. Temporary baseline test files were removed.

## Risk and rollout

The helper changes initial artifact production through the scan skills. It does not relax finalizer validation, provide a trusted completion binding, or adopt existing incomplete files. No public commands, flags, environment variables, schemas, or defaults change. The helper is included in the plugin shipping manifest.

Tests cover deterministic assembly and finalization; they do not exercise a live model's adherence to the skill. Existing incomplete scans still require the documented preservation/fresh-scan path. No migration is performed.

## Public disclosure review


IvanCaceres:
@kmbroai @mldangelo-oai can you review this PR?

PR #852, opened 2026-09-09
feat(sdk): pipeline dedupe with configurable concurrency

kmbroai:
## Summary

Deduplication waits for each candidate lookup and review in sequence, then waits for every Luna screening before starting Sol verification. Use a configurable worker pool so independent reviews run concurrently and ready Sol pairs can overlap with unrelated Luna screenings. Completed workflow checkpoints survive failures and restarts.

## Changes

- Add `dedupe --concurrency N` and SDK `concurrency` options for saved-scan and scan-directory deduplication. Both default to `8`, accept positive safe integers, and support `1` for serial execution.
- Fetch candidate neighborhoods concurrently, then run two queues sharing one pool: pending Luna screenings and ready Sol pair reviews. Refill free slots immediately, preferring ready Sol work. The configured limit applies across both models combined.
- Wait for every Luna screening covering a pair before verifying it, preserving `DISTINCT` vetoes. Preserve input-order finding records, pair prompts, and grouping results regardless of completion order.
- Retry transient candidate lookups and idempotent group writes up to three attempts, and transient or invalid Codex reviews up to three fresh sessions. Use exponential backoff with jitter and honor HTTP `Retry-After`; retain the corrective turn for invalid submissions.
- Stop both queues after a terminal failure and drain active jobs so successful reviews can save checkpoints. Preserve cancellation. Existing workflow IDs reuse completed Luna and Sol reviews even when concurrency changes; incomplete runs do not publish groups.
- Update CLI help, SDK documentation, package inventory, and regression tests for scheduling, retries, failure draining, and durable resume.

## Testing

- Seeded full SDK suite: **2,669 passed, 48 skipped, 0 failed** (seed `12345`).
- Focused deduplication tests: **27 passed**. Focused SQLite checkpoint/resume tests: **3 passed**.
- A disposable differential check matched the previous serial implementation across **500 generated scenarios**, including exact Luna and Sol review inputs and grouping results.
- `pnpm run types`, `pnpm run format`, and `pnpm run build` passed. Built CLI help describes the shared limit and default; `git diff --check` passed.
- Installed-package smoke was not rerun for this scheduler update. The earlier local attempt reached smoke installation and failed with the registry error `ETARGET: No matching version found for fast-uri@3.1.6` for the existing dependency.

## Risk and rollout

The default changes from serial execution to eight concurrent jobs, increasing simultaneous service requests and Codex processes. Use `--concurrency 1` or SDK `concurrency: 1` for serial operation, or lower the count to fit available capacity. Retry backoff occupies its worker slot; retries can add model usage.

Candidate retrieval completes before model scheduling so pair dependencies and review inputs are stable. Sol work overlaps with unrelated Luna work, but throughput still depends on available capacity, review durations, and pair dependencies. Live-service latency improvements have not been benchmarked.

Result shapes, model settings, repository scope, validation rules, and checkpoint identities remain unchanged. Durable resume still requires `--workflow-id` or SDK `workflowId`. No database migration is needed.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/deduplication/deduplication.ts
+++ b/sdk/typescript/src/deduplication/deduplication.ts
@@ -1,10 +1,53 @@
 import type { Finding } from "../models.js";
 import type { FindingNeighborhood } from "../finding-retrieval.js";
+import { CodexSecurityError } from "../errors.js";
 import {
   pairKey,
   screeningPairSlot,
   type DeduplicationReviewer,
 } from "./deduplication-reviewer.js";
+
+export const DEFAULT_DEDUPE_CONCURRENCY = 8;
+
+export function deduplicationConcurrency(
+  concurrency = DEFAULT_DEDUPE_CONCURRENCY,
+): number {
+  if (!Number.isSafeInteger(concurrency) || concurrency < 1)
+    throw new CodexSecurityError(
+      "concurrency must be a positive safe integer.",
+    );
+  return concurrency;
+}
+
+async function mapConcurrent<T, R>(
+  items: readonly T[],
+  concurrency: number,
+  operation: (item: T) => Promise<R>,
+  signal?: AbortSignal,
+): Promise<R[]> {
+  const results = new Array<R>(items.length);
+  let next = 0;
+  let failure: { error: unknown } | undefined;
+  async function worker(): Promise<void> {
+    while (failure === undefined) {
+      try {
+        signal?.throwIfAborted();
+        const index = next++;
+        if (index >= items.length) return;
+        results[index] = await operation(items[index]!);
+      } catch (error) {
+        failure ??= { error };
+      }
+    }
+  }
+  // Drain started jobs so successful reviews can finish saving their checkpoints.
+  await Promise.all(
+    Array.from({ length: Math.min(concurrency, items.length) }, worker),
+  );
+  signal?.throwIfAborted();
+  if (failure !== undefined) throw failure.error;
+  return results;
+}
 
 export interface DeduplicationResult {
   uniqueFindingIds: string[];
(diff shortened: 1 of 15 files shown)


sdk/typescript/src/deduplication/deduplication.ts
```diff
@@ -179,47 +234,96 @@ export class FindingDeduplicator {
...
     const findings = new Map<string, Finding>();
-    const nominated = new Map<string, [string, string]>();
-    const rejected = new Map<string, [string, string]>();
-    for (const id of ids) {
+    const neighborhoods = new Array<Finding[]>(ids.length);
+    await runQueued(
+      ids.map((id, index) => async () => {
+        const result = await this.candidates.potentialDuplicates(id);
+        this.signal?.throwIfAborted();
+        neighborhoods[index] = [result.finding, ...result.potentialDuplicates];
```
daneschneider-oai, line 250: **Non-blocking performance follow-up: avoid retaining duplicate neighborhood payloads on large scans.**

This array retains every fully hydrated neighborhood before model scheduling. Each HTTP lookup parses its own finding objects, so overlapping neighborhoods retain repeated copies even though active jobs are capped. Lowering `--concurrency` does not reduce this particular allocation.

An isolated comparison of the scheduler/grouping source with synthetic inputs (80 anchors, 40 shared neighbors, 48 KiB of inline evidence per finding) used about 170 MB peak heap here versus 16 MB in the pre-PR serial implementation, with identical output; concurrency 1 and 8 had similar retention. This is not a full SDK benchmark or evidence that those document sizes are typical.

I would not block expected small local scans on this. For larger or evidence-heavy scans, consider sharing identical finding payloads and keeping compact per-neighborhood metadata while preserving exact review inputs and deterministic ordering.

daneschneider-oai (commented):
One non-blocking memory-retention follow-up for larger scans; this does not block approval of the overlapping scheduler.

daneschneider-oai (approved):
LGTM on the Luna/Sol overlap update. The shared concurrency limit, pair-specific screening dependencies and DISTINCT vetoes, deterministic inputs/results, and failure draining look sound. Eight isolated tests of the updated queue passed; I did not rerun the full SDK suite. The inline memory-retention note is a non-blocking follow-up for larger scans, not a request to hold this PR.

sdk/typescript/src/deduplication/deduplication.ts
kmbroai: [CODSEC-564: Reduce duplicate finding payload retention during dedupe neighborhood prefetch](https://linear.app/openai/issue/CODSEC-564).

PR #853, opened 2026-09-09
fix(cli): improve scan usage and cost reports

ianw-oai:
## Summary

Scan summaries omitted cache-write and total token counts. Missing cache-write usage appeared as zero, and some model prices were missing or outdated. This change shows the full token breakdown and records the prices used for each estimate.

## Changes

- Show uncached input, cache reads, cache writes, output, and total tokens. Mark missing cache-write counts and the derived uncached count as unavailable.
- Add GPT-5.5 and GPT-6 Astra prices. Update GPT-5.6, Sol, and Daybreak Blue prices. Show unavailable cost when prices or usage are missing.
- Save the price source, date, tier, context category, and rates with each estimate. Preserve known cache-write charges when other workers omit that data.
- Keep cache-write data from matching session logs when the completion receipt contains a default zero.
- Wrap longer dashboard summaries and document the estimate limits.

## Testing

- TypeScript full suite: 2,423 passed, 43 skipped, 0 failed, with isolated test state.
- The seeded full run hit one timeout in the unchanged 4,096-file copy test. That test passed on its own on retry.
- Focused accounting, SDK event, CLI, and dashboard tests: 344 passed, including four cases through the real SDK parser.
- Scan-history Python tests: 26 passed.
- Plugin Ruff checks, Ruff format checks, and portable source checks passed.
- Git whitespace check passed.

## Risk and rollout

The added result fields are optional for compatibility with older saved results. Estimates use standard short-context prices. Missing cache-write data can understate cost. Updated prices also change when existing cost limits stop a scan. If the Codex runtime converts omitted cache-write counts to zero before logging them, the CLI cannot recover that distinction.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/cost-model.ts
+++ b/sdk/typescript/src/cost-model.ts
@@ -38,11 +38,9 @@ export interface ScanTokenUsage {
 }
 
 const MODEL_PRICING_NANODOLLARS: Readonly<Record<string, ModelPricing>> = {
-  // https://developers.openai.com/api/docs/models/gpt-5.5
   // GPT-5.5 has no additional cache-write charge.
   "gpt-5.5": [5_000, 500, 5_000, 30_000],
   "gpt-5.5-2026-04-23": [5_000, 500, 5_000, 30_000],
-  // https://developers.openai.com/api/docs/pricing
   "gpt-6-astra": [10_000, 1_000, 12_500, 50_000],
   "gpt-5.6": [4_000, 400, 5_000, 20_000],
   "gpt-5.6-sol": [4_000, 400, 5_000, 20_000],
@@ -148,10 +146,7 @@ export function estimateScanCost(
   };
 }
 
-export function formatTokenUsage(
-  value: unknown,
-  formatCount = (count: number) => count.toLocaleString("en-US"),
-): string | null {
+export function formatTokenUsage(value: unknown): string | null {
   const usage = tokenUsage(value);
   if (usage === null) return null;
   const writes =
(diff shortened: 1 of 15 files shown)


sdk/typescript/src/cost-model.ts
```diff
@@ -65,6 +83,10 @@ export function tokenUsage(value: unknown): ScanTokenUsage | null {
     input_tokens: input,
     cached_input_tokens: cached,
     cache_write_input_tokens: cacheWrite,
+    ...(value["cache_write_input_tokens_reported"] === false ||
+    (canonicalCacheWrite == null && legacyCacheWrite == null)
+      ? { cache_write_input_tokens_reported: false }
+      : {}),
```
mldangelo-oai, line 89: **[P2] Preserve missing cache-write usage through SDK normalization**

The pinned Codex SDK (`@openai/codex-sdk@0.149.1`) inserts `cache_write_input_tokens: 0` when a completion event omits it, so this check treats missing usage as a reported zero. I reproduced this using the real SDK with synthetic usage: `{ input_tokens: 120, cached_input_tokens: 30, output_tokens: 15 }`.

The raw session initially formats as `unavailable cache writes`. At completion, `ScanCostTracker.stop(receipt)` prefers the SDK receipt when its total matches the session total, changing the summary to `0 cache writes` and `90 uncached input`; the saved result also loses `cacheWriteInputTokensReported: false`.

Could we preserve the raw session's availability information when merging the completion receipt and add a regression through SDK normalization? This otherwise reintroduces the misleading zero in the new breakdown.

Reviewed `ddf4dc0d6d048f3efbb4d7bf71b2b7d6272bb611`. The 310 focused accounting, dashboard, CLI, and API-event tests passed; the additional SDK reproduction exposes this gap.

sdk/typescript/src/cost-model.ts
```diff
@@ -65,6 +83,10 @@ export function tokenUsage(value: unknown): ScanTokenUsage | null {
     input_tokens: input,
     cached_input_tokens: cached,
     cache_write_input_tokens: cacheWrite,
+    ...(value["cache_write_input_tokens_reported"] === false ||
+    (canonicalCacheWrite == null && legacyCacheWrite == null)
+      ? { cache_write_input_tokens_reported: false }
```
daneschneider-oai, line 88: [P2] Preserve missing cache-write availability through the runtime

This check treats a present canonical zero as reported usage, but pinned Codex SDK 0.149.1 applies `parsed.usage.cache_write_input_tokens ??= 0` before yielding `turn.completed` ([source](https://github.com/openai/codex/blob/ff29a44391deccde0aba0f8390337d7f3c319ea4/sdk/typescript/src/thread.ts#L107-L110)). The pinned Rust provider parser also defaults omitted cache-write counts to zero ([source](https://github.com/openai/codex/blob/ff29a44391deccde0aba0f8390337d7f3c319ea4/codex-rs/codex-api/src/sse/responses.rs#L133-L155)). As a result, provider omissions remain indistinguishable from measured zero on these paths.

Reproduced with 100 input, 20 cached input, and 10 output tokens: a raw rollout without cache writes displays `unavailable cache writes`, then the SDK-normalized final receipt changes the same tracker to `80 uncached input` and `0 cache writes`. The unavailable flag needs to survive normalization before this check. Recording this as a non-blocking follow-up while approving.

daneschneider-oai (approved):
Approving to unblock, with the P2 finding below recorded as a non-blocking follow-up: missing cache-write usage can still appear as a measured zero after Codex runtime normalization.

Validation: 120 focused accounting, dashboard, and CLI summary tests passed. Separately reproduced the availability flag being lost when the tracker receives the SDK-normalized final usage.

PR #854, opened 2026-09-09
feat(cli): add a feedback command

ianw-oai:
## Summary

The CLI can read saved diagnostics but cannot send a support report. Add a `feedback` command that sends a report through Codex and returns a feedback ID.

## Changes

- Add `codex-security feedback [SCAN_ID] --reason "..." [--include-logs] [--json]`.
- Default to the most recently started scan in the current repository, including active or failed scans. Send a general report when no scan exists.
- Include version metadata. Attach saved scan and worker activity only when `--include-logs` is set. Reuse the existing scan log reader.
- Add command help, documentation, and tests for uploads, errors, and cancellation. Include the new module in the package file allowlist.

## Testing

- Focused CLI and feedback tests: 180 passed, including a real saved-history regression where an older scan is still running, a newer scan has failed, and another repository has a later scan.
- Feedback process tests also passed in Windows CI.
- All three SDK CI test groups with seed 12345: 2,432 passed and 43 skipped. One process-group test was blocked by the sandbox's `ps` restriction and passed on a permitted rerun.
- Type checks, formatting, build, and `git diff --check` passed.
- Packed archive validation passed (419 entries).
- Checked compiled command help and schema. Checked the real Codex protocol with feedback disabled; upload tests used a local mock server.

## Risk and rollout

This adds a command without changing existing command output. Logs are off by default. With `--include-logs`, reports can contain prompts, code, findings, and tool output. The command respects `feedback.enabled = false` and removes temporary attachments when it finishes or is canceled.

## Public disclosure review

### DIFF
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -4619,7 +4619,7 @@ export async function main(
           .min(1)
           .optional()
           .describe(
-            "Scan ID or unique prefix (default: latest scan in the current repository).",
+            "Scan ID or unique prefix (default: most recently started scan in the current repository).",
           ),
       }),
       options: z.object({
@@ -4643,10 +4643,14 @@ export async function main(
             "list-scans",
             "--repository",
             dependencies.currentDirectory(),
-            "--limit",
-            "1",
           ]);
-          scanId = (value["scans"] as ScanLogSource[])[0]?.scanId;
+          const scans = value["scans"] as {
+            scanId: string;
+            startedAt: string;
+          }[];
+          scanId = scans.toSorted((a, b) =>
+            b.startedAt.localeCompare(a.startedAt),
+          )[0]?.scanId;
         }
         const scan =
           scanId === undefined
--- a/sdk/typescript/src/cli.ts
+++ b/sdk/typescript/src/cli.ts
@@ -168,7 +168,8 @@ import {
   CODEX_SECURITY_THREAD_SOURCES,
   type CodexSecurityThreadSource,
 } from "./thread-source.js";
-import { readScanLogs } from "./scan-logs.js";
+import { readSavedScanLogs, type ScanLogSource } from "./scan-logs.js";
+import { sendFeedback } from "./feedback.js";
 import {
   renderScanHistory,
   type HistoryCommand,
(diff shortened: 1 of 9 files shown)


mldangelo-oai:
Reviewed `0ade8864b109f35e1644824910837f1de775e2ac`: no remaining correctness findings from my review.

The latest commit fixes the missing `feedback` entry in the package allowlist, and the updated package CI check passes. I ran the CLI and feedback tests: **173 passed**, including scan selection, log opt-in, upload failures, cancellation, and temporary-file cleanup. Upload verification used a mock server; I also checked the request shape against the bundled Codex protocol schema.

Looks good to me once the remaining CI finishes.

sdk/typescript/src/cli.ts
```diff
@@ -4631,6 +4609,76 @@ export async function main(
...
+        let scanId = args.scanId;
+        if (scanId === undefined) {
+          const value = await history([
+            "list-scans",
+            "--repository",
+            dependencies.currentDirectory(),
+            "--limit",
+            "1",
+          ]);
+          scanId = (value["scans"] as ScanLogSource[])[0]?.scanId;
```
daneschneider-oai: [P2] Select the newest scan independently of history's running-first ordering

`list-scans --limit 1` does not necessarily return the latest scan: the underlying query sorts every running scan ahead of failed or completed scans ([query](https://github.com/openai/codex-security/blob/39c9bbba055c06dd033b89bac11e4ebd504236f9/plugins/codex-security/scripts/workbench_scan_history.py#L178-L184)). With yesterday's scan still running and today's scan failed, `feedback --reason "Scan failed" --include-logs` therefore submits yesterday's IDs and logs.

Reproduced this through the actual workbench CLI with two synthetic saved scans. The default should select the most recent scan independently of status; the current latest-selection test mocks the returned array and misses this ordering. Recording this as a non-blocking follow-up while approving.

daneschneider-oai (approved):
Approving to unblock, with the P2 finding below recorded as a non-blocking follow-up: the latest-scan default can select an older running scan instead of the newest failed scan.

Validation: 26 focused feedback, CLI feedback, and scan-log tests passed on the previously reviewed head; also inspected the subsequent Windows-path test adjustment. Separately reproduced the selection issue through the actual workbench CLI using synthetic saved scans.

mldangelo-oai (approved):

PR #862, opened 2026-09-10
feat(plugin): persist scan artifacts through MCP

soyeon-oai:
## Summary

Plugin scans currently use temporary directories, and several skill workflows write retained documents and evidence directly. Route those artifacts through the existing MCP server with explicit temporary or persistent storage. New plugin scans default to `~/.codex/state/plugins/codex-security/scans/...`.

## Changes

- Add `save_codex_security_artifact` and `read_codex_security_artifact` to the existing MCP. They support exact text, binary imports from the same temporary workspace, restart readback, and standalone target-bound document collections.
- Preserve typed canonical-artifact tools. Publish supplemental scan files under the existing completion lock and reject stopped/sealed scans, stale claims, unsafe paths, and canonical-file replacement.
- Route non-SDK skill output and retained helper/PoC evidence through MCP. Build and execution output stays temporary until imported. Keep SDK-owned authoring and bound Deep-worker tools unchanged.
- Honor `CODEX_SECURITY_SCAN_ROOT`, `CODEX_SECURITY_STATE_DIR`, and `CODEX_HOME`. Existing scans keep their saved paths.
- Add the internal workbench helper `save-scan-artifact --scan-id ID --artifact-path PATH --source-path PATH [--claim-token TOKEN]` so MCP publication shares Python finalization locking. No SDK CLI command or flag changes.

## Testing

- Fail-first MCP regression, then passing end-to-end stdio tests: persistent roots, temporary routing, exact UTF-8/binary bytes, restart, temporary cleanup, standalone documents, state-directory aliases, path/claim boundaries, and sealed/completed scan immutability.
- Full MCP suite: 24 test files passed.
- MCP TypeScript checks, SDK `build:ci` and types, plugin bundle generation, Ruff lint/format, and portable-source checks (9 tests) passed.
- SDK custom-validation/runtime/resume: 198 passed, 22 platform skips.
- Python: 1,092 passed, 4 skipped, 110 subtests passed with `env -u PYTHONSAFEPATH python3 -m pytest -q -n 4 plugins/codex-security/tests`. The inherited environment variable also broke the unchanged baseline; unrelated test-runner cleanup printed process-query permission warnings after the passing summary.

## Risk and rollout

- New plugin scans persist by default. This does not migrate older temporary scans or change SDK storage.
- Non-SDK scans require MCP for retained output; an unavailable MCP now surfaces a blocker instead of using direct file-authoring fallback.
- Completed scans remain immutable. Later documents use a separate target-bound collection.
- No retention cleanup policy is introduced. Live Desktop/model and Windows execution have not been tested locally; hosted CI is pending.

## Public disclosure review

## Change impact

MCP publishes artifacts through the existing safe writer; standalone temporary storage works without creating or requiring a persistent collection.

```mermaid
flowchart LR
  subgraph column_0["Plugin workflow"]
    direction TB
    node_0["Documents and evidence"]
  end
  subgraph column_1["MCP publication"]
    direction TB
    node_1["MCP storage selection"]
    node_2["Scan completion lock"]
  end
  subgraph column_2["Existing file writer"]
    direction TB
    node_3["Python and Windows writer"]
  end
  subgraph column_3["Saved output"]
    direction TB
    node_4["Temporary workspace"]
    node_5["Retained scan files"]
    node_6["Standalone documents"]
  end
  node_0 -->|"save/read instructions"| node_1
  node_1 -->|"running scan bytes"| node_2
  node_1 -->|"temporary/standalone bytes"| node_3
  node_2 -->|"validated scan bytes"| node_3
  node_3 -->|"atomic publication"| node_4
  node_3 -->|"atomic publication"| node_5
  node_3 -->|"atomic publication"| node_6
  class node_0 context
  class node_1 changed
  class node_2 changed
  class node_3 context
  class node_4 affected
  class node_5 affected
  class node_6 affected
  classDef changed fill:#d7f5e5,stroke:#237a4b,color:#111
  classDef affected fill:#e6f0ff,stroke:#3569a8,color:#111
  classDef context fill:#f2f3f5,stroke:#6e7781,color:#111
```

> **Limits:** SDK behavior is unchanged. Default state can fall back to OS temporary storage. · The internal save-scan-artifact helper now consumes stdin; its former --source-path staging argument is removed. · Current-head artifact, smoke, fallback and portable checks pass. Hosted CI and live Desktop/Windows validation remain separate.

<details>
<summary>Source evidence (10)</summary>

- [`plugins/codex-security/references/artifact-storage.md:L3-L25`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/references/artifact-storage.md#L3-L25) — Skills use the shared MCP storage policy, which defines the SDK exception once.
- [`plugins/codex-security/mcp-app/src/server/compact-artifact-tools.ts:L252-L288`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/server/compact-artifact-tools.ts#L252-L288) — The existing MCP save/read tools resolve scan or standalone artifact contexts.
- [`plugins/codex-security/mcp-app/src/artifact-storage.ts:L37-L49`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/artifact-storage.ts#L37-L49) — Storage overrides resolve from the plugin directory, matching the workbench.
- [`plugins/codex-security/mcp-app/src/artifact-storage.ts:L53-L76`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/artifact-storage.ts#L53-L76) — Standalone temporary requests bypass persistent creation and validation; retained documents keep their existing collection.
- [`plugins/codex-security/mcp-app/src/artifact-storage.ts:L66-L99`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/artifact-storage.ts#L66-L99) — Resolve existing ancestors without creating the persistent collection, then select the context-specific OS temporary directory.
- [`plugins/codex-security/mcp-app/src/artifact-storage.ts:L154-L165`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/artifact-storage.ts#L154-L165) — MCP sends exact bytes on stdin to the locked scan helper or the temporary/standalone helper.
- [`plugins/codex-security/scripts/workbench_saved_results.py:L1344-L1379`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/scripts/workbench_saved_results.py#L1344-L1379) — The scan helper verifies the claim and seal under the completion lock before calling the file writer.
- [`plugins/codex-security/mcp-app/server.ts:L83-L91`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/server.ts#L83-L91) — Default artifact roots follow the workbench-selected state, including its temporary fallback; explicit scan-root overrides retain precedence.
- [`plugins/codex-security/scripts/workbench_saved_results.py:L1337-L1341`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/scripts/workbench_saved_results.py#L1337-L1341) — Temporary and standalone publication reuse write\_scan\_local\_bytes with the MCP-selected root.
- [`plugins/codex-security/scripts/finalize_scan_contract.py:L521-L624`](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/scripts/finalize_scan_contract.py#L521-L624) — The existing writer holds directory descriptors on POSIX and delegates to the existing Windows handle implementation.

</details>

## Screenshots

succeed to store under `~/.codex` in desktop plugin and CLI

image
image


plugins/codex-security/mcp-app/src/artifact-storage.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  let bytes: Buffer;
+  if (input.sourcePath !== undefined) {
+    if (input.storage !== "persistent") throw new Error("Import retained files into persistent storage.");
+    const scratch = await storageContext(context, "temporary", false);
+    const source = relative(scratch.root, resolve(input.sourcePath));
+    if (isAbsolute(source) || source === ".." || source.startsWith(".." + sep)) {
+      throw new Error("Import source must be inside this artifact context's temporary directory.");
+    }
+    bytes = await readArtifactBytes(scratch, components(source.split(sep).join("/")), "Artifact import");
```
mldangelo-oai, line 150: **P1: bind the imported bytes to the validated source file**

Reproduced on `cc0fa293c39b00811bfa13d49b206b1056b1b8a2`. The new `sourcePath` branch passes a generated file from the temporary workspace to `readArtifactBytes`. That helper checks each component with `lstat`, calls `realpath`, and then separately opens the pathname with `fs.readFile` ([reader](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/artifact-io.ts#L72-L92)). A replacement after validation can redirect that final open outside the staging directory.

The deterministic fixture used this sequence:

1. Prepare temporary storage and create a regular generated-output file inside it.
2. Immediately before the actual `readFile`, replace that file with a symlink to a synthetic sibling file outside staging.
3. Import it with `saveCodexSecurityArtifact`, then read the retained artifact with `readCodexSecurityArtifact`.

The returned content was the sibling file's `SYNTHETIC_PRIVATE_BYTES`, rather than the generated output. All files and contents were synthetic. The test controls the replacement point, as the PR's destination-race tests do; it demonstrates the unsafe I/O sequence, not the success rate of an unsynchronized race.

Sending bytes to Python on stdin fixes the previous Python staging issue, but the bytes can already come from the wrong source at this preceding read. Read through a verified opened file bound to the permitted directory, with protection against replacement of both the final component and its parents. A regression test should assert that this replacement either fails the import or retains the originally opened bytes.

plugins/codex-security/mcp-app/src/artifact-storage.ts
```diff
@@ -0,0 +1,181 @@
...
+  create: boolean
+): Promise<ArtifactContext> {
+  const target = await runWorkbench(["inspect-target", "--target-path", targetPath]);
+  if (typeof target.targetPath !== "string") throw new Error("Missing artifact target.");
+  const repoRoot = await fs.realpath(target.targetPath);
+  const name = basename(repoRoot).replace(/[^a-zA-Z0-9._-]+/g, "-") || "repository";
+  const identity = createHash("sha256").update(repoRoot).digest("hex");
+  const root = join(persistentScanRoot(), name, `artifacts-${identity}`);
+  const existingRoot = await fs.realpath(persistentScanRoot()).catch(() => persistentScanRoot());
+  if (existingRoot === repoRoot || existingRoot.startsWith(repoRoot + sep)) {
```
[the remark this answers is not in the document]
mldangelo-oai: **Rechecked on `cc0fa293c39b00811bfa13d49b206b1056b1b8a2`: this still reproduces.**

I exercised the actual stdio MCP server and Python workbench with a clean synthetic Git repository at `/fixture/repository` and `CODEX_SECURITY_SCAN_ROOT=/fixture`. This call succeeds:

```js
save_codex_security_artifact({
  targetPath: "/fixture/repository",
  storage: "persistent",
  path: "threat_model.md",
  content: "Synthetic document\n"
})
```

The returned path is `/fixture/repository/artifacts-<hash>/threat_model.md`, and `git status --porcelain` reports the new `artifacts-<hash>/` directory.

The current [check](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/artifact-storage.ts#L65-L75) compares `scanRoot` with `repoRoot`, but creates the derived `root`. Validate the resolved final collection path before creating it. This is ordinary storage configuration, and it contaminates the selected source tree despite the intended outside-target restriction.

plugins/codex-security/mcp-app/src/artifact-storage.ts
```diff
@@ -0,0 +1,175 @@
...
+function supplementalPath(input: ArtifactLocation, context: ArtifactContext): string[] {
+  const parts = components(input.path!);
+  if (input.storage === "temporary") return parts;
+  const path = parts.join("/").toLowerCase();
+  const allowed = (parts.length > 1 && ["artifacts", "findings", "hardening"].includes(parts[0]!))
+    || path === "report_validation.md" || (!context.scanId && path === "threat_model.md");
+  if (!allowed || path === "artifacts/deep_discovery" || path.startsWith("artifacts/deep_discovery/")
+    || path === "artifacts/02_discovery/candidate_ledger.jsonl"
+    || path === "artifacts/02_discovery/in_scope_files.txt"
+    || path === "artifacts/01_context/false_positive_feedback.json") {
```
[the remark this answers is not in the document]
mldangelo-oai: **Rechecked on `cc0fa293c39b00811bfa13d49b206b1056b1b8a2`: this still reproduces, and there is an opportunity to simplify the duplicated policy.**

Using the real stdio server and workbench:

1. Start a fresh diff scan over two commits in a synthetic repository.
2. Call `prepare_codex_security_review_items`.
3. Save a supplemental file at `artifacts/02_discovery/candidate_ledger.jsonl/note.txt`.
4. Call `record_codex_security_discovery_candidates` with an empty candidates array.

Step 3 succeeds and creates `candidate_ledger.jsonl` as a directory. Step 4 fails with `discovery candidates: destination is not a regular file.` There is no artifact-tool operation to remove that directory and recover the scan.

Reject descendants of each reserved canonical filename as well as exact matches.

**Simplification:** the allowed prefixes and reserved filenames are separately hard-coded in [TypeScript](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/src/artifact-storage.ts#L112-L124) and [Python](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/scripts/workbench_saved_results.py#L1362-L1377). Both copies miss reserved-file descendants, and their `deep_discovery` exact-path handling differs. Declare these paths once in a small shared data definition and consume it from both sides. Preserve the intentional temporary/persistent and standalone/scan differences, along with Python's authoritative claim, seal, completion-lock, and safe-write checks. This fix should not require a general validation framework.

plugins/codex-security/mcp-app/src/artifact-io.ts
```diff
@@ -219,12 +227,18 @@ export async function artifactDestination(
 export async function replaceArtifactText(
   path: string,
   content: string
+): Promise<void> {
+  await replaceArtifactBytes(path, Buffer.from(content, "utf8"));
+}
+
+export async function replaceArtifactBytes(
```
mldangelo-oai, line 234: **Simplification: remove the remaining JavaScript publication scaffolding**

On `cc0fa293c39b00811bfa13d49b206b1056b1b8a2`, `replaceArtifactBytes` has exactly one caller: `replaceArtifactText`, which converts its string to a buffer to enter this wrapper. The supplemental storage code stopped using it when publication moved to the Python helpers in `1a11172`.

Restore the original text writer with its UTF-8 write option and remove this extra exported byte-writing abstraction. Keep `readArtifactBytes`, which has actual binary-import and readback consumers.

The same cleanup applies to the [replacement test's `fs.open` monkeypatch](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/tests/test_artifact_storage_regressions.mjs#L164-L180). It intercepts the old JavaScript `.lock` path, but both current storage modes publish through Python. Remove that interception and its `finally` restoration, retain the Python directory-replacement hook, and assert that the marker contains `python`. The test still fails if the protected Python path is no longer exercised.

This is a nonblocking cleanup of code left behind by the implementation change.

plugins/codex-security/mcp-app/src/server/compact-artifact-tools.ts
```diff
@@ -242,6 +249,43 @@ export function registerCompactArtifactTools(
...
+  options: CompactArtifactToolOptions,
+  write: boolean,
+  requestContext: unknown
+): Promise<ArtifactContext> {
+  if ((input.scanId === undefined) === (input.targetPath === undefined)) {
+    throw new Error("Provide exactly one scanId or standalone targetPath.");
+  }
+  if (input.scanId !== undefined) return scanContext({ ...input, scanId: input.scanId }, options, write, requestContext);
+  if (input.handoffClaimToken !== undefined) throw new Error("A handoff claim requires a scanId.");
+  const root = await options.resolveScanRoot?.() ?? persistentScanRoot(options.pluginRoot);
```
mldangelo-oai, line 287: **Simplification: require the authoritative root resolver**

At `cc0fa293c39b00811bfa13d49b206b1056b1b8a2`, the only caller of `registerCompactArtifactTools` is [the server registration](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/mcp-app/server.ts#L1320-L1329), and it always supplies `resolveScanRoot: scanRoot`. Making this callback optional retains an unused second route for choosing the artifact root.

Make `resolveScanRoot` required, replace this expression with `await options.resolveScanRoot()`, and remove the `persistentScanRoot` import from this module. This gives storage selection one owner and avoids callers accidentally skipping the workbench-selected fallback state.

Keep the real unwritable-state fallback in `server.ts`; only the unused alternative here needs removal. This is a nonblocking simplification.

plugins/codex-security/references/scan-artifacts.md
```diff
@@ -2,6 +2,8 @@
 
 Use these shared path conventions for Codex Security scan workflows unless the user explicitly provides different input or output paths.
 
+First read `artifact-storage.md` and follow its storage and threat-model cache rules, which take precedence over the legacy temporary base paths below.
```
mldangelo-oai, line 5: **Documentation simplification: update the base paths directly**

The precedence sentence leaves the main guide defining `security_scans_dir` from the OS temporary directory, constructing `scan_id` from a commit/timestamp, and locating the shared threat model under that old directory. The new [storage policy](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/references/artifact-storage.md#L22-L27) instead requires the authoritative MCP directory, and its [cache instructions](https://github.com/openai/codex-security/blob/cc0fa293c39b00811bfa13d49b206b1056b1b8a2/plugins/codex-security/references/artifact-storage.md#L47-L51) use a standalone target collection.

Define the plugin's `scan_dir` from the returned scan context, keep the relative phase layout in this guide, and move any still-needed SDK or legacy formulas into an explicitly scoped subsection. Point shared-cache operations to the single current description.

The explicit precedence handles the conflict, but readers currently have to reconcile two sets of path definitions. Updating the guide removes that extra work without changing SDK behavior. Optional cleanup.

mldangelo-oai:
## Review, QA, and simplification pass

Reviewed and tested head `cc0fa293c39b00811bfa13d49b206b1056b1b8a2` on `codex/plugin-artifact-storage`.

### Three reproduced issues

1. **P1: imported source bytes can come from outside staging.** Replacing the validated source with a symlink before the pathname is opened makes the artifact contain the synthetic sibling file's bytes. The reproduction controls the replacement point; it establishes the unsafe I/O sequence, not the success rate of an unsynchronized race. [Detailed comment and suggested regression test](https://github.com/openai/codex-security/pull/862#discussion_r3985004174).
2. **P2: an ancestor scan root can put standalone artifacts inside the target repository.** The guard checks the configured scan root, while the derived collection path lands in the source tree and makes it dirty. Reproduced through the real stdio server and workbench. [Fresh evidence on the existing thread](https://github.com/openai/codex-security/pull/862#discussion_r3985004370).
3. **P2: a supplemental path can turn a reserved ledger filename into a directory.** Saving beneath `candidate_ledger.jsonl/` succeeds, after which the typed candidate writer fails. Reproduced through the real stdio server and workbench. [Fresh evidence on the existing thread](https://github.com/openai/codex-security/pull/862#discussion_r3985004802).

### Simplification opportunities

- [Share the small supplemental-path policy](https://github.com/openai/codex-security/pull/862#discussion_r3985004802) between TypeScript and Python, while retaining the authoritative claim, seal, completion-lock, and safe-write checks.
- [Remove the obsolete byte-writer wrapper and JavaScript test interception](https://github.com/openai/codex-security/pull/862#discussion_r3985008093) left behind when publication moved to Python.
- [Require the root resolver already supplied by the only caller](https://github.com/openai/codex-security/pull/862#discussion_r3985008570), retaining the actual state-root fallback in the server.
- [Update the artifact guide's base-path definitions](https://github.com/openai/codex-security/pull/862#discussion_r3985008861) so readers do not have to reconcile the old formulas with a precedence note.

### Validation performed

Local runs used an isolated checkout on macOS with Node 24.15.0 and Python 3.12.13.

| Check | Result |
| --- | --- |
| Full Python suite, four workers | 1,093 passed, 4 skipped, 110 subtests passed |
| Full MCP suite | 33 passed |
| Focused artifact/fallback tests | 11 passed; included in the full MCP coverage |
| Ruff lint and formatting | Passed with Ruff 0.16.1 |
| SDK `build:ci` and SDK/MCP type checks | Passed |
| Plugin bundle generation | Passed, 139 files |
| Portable source compatibility check and checker tests | Passed; 9 checker tests |
| Diff whitespace check | Passed |

Coverage includes text and binary retention, restart, temporary cleanup isolation, fallback and relative roots, claims/seals, destination replacement, packaged stdio smoke testing, and Deep Scan discovery/reducer/restart behavior. The three reproductions above use synthetic fixtures and demonstrate cases that the passing suites do not catch.

Package scripts ran through Node's `--run`. Dependencies were reused after comparing lockfiles; generated native assets were reused after confirming their sources were unchanged. This was not a clean dependency-install validation.

### CI and testing limits

At the latest check, Linux and macOS CI pass, but [Windows Node 24 package verification](https://github.com/openai/codex-security/actions/runs/34546543533/job/103101420834) fails with `PluginPythonUnavailableError`, and the Windows aggregate gates fail. The relevant SDK/runtime/package-fixture files are unchanged by this PR. I did not run a same-runner base comparison, so I have not established whether that failure is a PR regression.

Live Desktop/model execution and local Windows execution were not tested.

mldangelo-oai (approved):
Thank you!

PR #870, opened 2026-09-10
fix(plugin): authenticate deep workers with OpenAI API keys

mldangelo-oai:
## Summary

With `OPENAI_API_KEY` set and no stored Codex login, Deep Scan workers forwarded the variable but did not supply the SDK's `apiKey` option. Native `codex exec` reads `CODEX_API_KEY`, so requests could leave the worker without an Authorization header.

Use the same SDK authentication option as CLI scan sessions when Codex reports that the worker has no account and its provider requires OpenAI authentication.

## Changes

- Reuse the existing app-server preflight to read the selected provider's account state before supplying the API key.
- Preserve an existing `CODEX_API_KEY`, stored API/ChatGPT account, custom-provider authentication, and forced ChatGPT login.
- Cover new and resumed workers, credential precedence, blank keys, parent-environment preservation, and the complete MCP worker-launch path using synthetic credentials.
- Make the lifecycle test's cleanup return when its child already exited by signal; repeated cleanup previously waited for an exit event that had already fired.
- Document how worker credentials are selected and inherited from the plugin process.

## Testing

- Failing-first regression: the original worker omitted `CODEX_API_KEY`; the patched worker receives the expected synthetic key.
- Real Codex 0.153.3 against an isolated local HTTP endpoint: the original executor sent no Authorization header; the patched executor sent the expected synthetic Bearer token. Stored API credentials, an explicit Codex key, and a custom provider retained their selected credentials.
- Focused executor, preflight, and MCP lifecycle tests passed.
- Full MCP test suite passed: 23/23 test files.
- MCP TypeScript typecheck passed.
- Ruff 0.16.1 lint and format checks passed.
- SDK `build:ci` and `build:plugin` passed.
- Plugin source compatibility check and its 9 tests passed.
- `git diff --check` passed.

The local pnpm runs used `--config.verify-deps-before-run=false` with existing dependencies.

## Risk and rollout

The fallback applies only to a nonempty `OPENAI_API_KEY` when no explicit Codex key or native account is selected and the provider requires OpenAI authentication. It adds an `account/read` request with token refresh disabled to the existing preflight when considering that fallback. It does not copy stored credentials or modify the parent process environment.

No new CLI flags or settings. Ship through the normal plugin release process. Validation used synthetic credentials and a local endpoint; a successful scan against the live service was not exercised.

## Public disclosure review


mldangelo-oai:
@codex review

Please review head `da3b0605f00411eda6d5b3cb11b9db43e0b22817`, particularly worker authentication and credential precedence.

mldangelo-oai:
@codex review

Please review head `a09c932f22d595863984bc68e961d184f9519ffd`. The follow-up updates the bundled worker shutdown fixture for the environment helper and authentication preflight result. All five shutdown tests now pass.

daneschneider-oai (approved):
The change addresses the OPENAI_API_KEY-only, no-stored-login worker failure by supplying the SDK apiKey option while preserving native credential precedence. The authentication setup covers both new and resumed workers.

Reviewed the executor and account preflight against the native authentication behavior. Independently reran the focused executor, permission-preflight, and stdio-lifecycle test files and the MCP TypeScript check; all passed. No blocking findings.

PR #872, opened 2026-09-10
fix(feedback): collect Desktop and worker session logs

daneschneider-oai:
## Summary

`codex-security feedback --include-logs` can miss sessions started through the Desktop plugin because it only searches the CLI's managed Codex home. It also stops collecting when the parent session log is missing. Extend the existing log reader and upload path to collect the selected scan's available sessions while keeping unrelated children of a reused owner conversation out of the upload.

## Changes

- Expose existing session IDs separately from scan execution roots, using saved CLI registration and recorded Deep Scan worker IDs.
- Search active and archived sessions in the configured Codex home and the CLI's managed home, deduplicate sessions, and follow descendants of recorded execution roots.
- Include shared owner logs explicitly without traversing their children or forwarding those owners to Codex's automatic feedback subtree collection.
- Attach available worker logs when parent logs are absent, using the existing `extraLogFiles` upload path and log consent settings.
- Add regression coverage for unrelated owner children and forks, Desktop standard continuations, missing parents, archives, and legitimate worker descendants.

## Testing

- Focused feedback, reader, CLI, and cost tests: 107 passed.
- CLI workbench tests: 32 passed after updating the fixture to the current workbench response.
- Workbench scan-history and usage tests: 48 passed.
- Portable source compatibility tests: 9 passed.
- SDK/MCP type checks, SDK CI build, Ruff, Prettier, and source compatibility checks passed.
- Full SDK rerun with seed `3582815882`: 2,716 passed, 55 skipped, zero failures. The first run exposed the outdated CLI workbench fixture noted above; its full 32-test module and the subsequent full suite pass.

## Risk and rollout

Run the feedback command on the machine holding the scan logs, with the matching Codex home and Codex Security state directory. Standard scans run inside an existing Codex conversation include only the owner's saved session because their worker-to-scan associations are not recorded. Deep Scans and scans launched by `codex-security` include their recorded execution roots and descendants. Earlier retries that started separate sessions may be missing when their IDs are no longer recorded. The change requires no database migration or Codex host update.

## Public disclosure review


sdk/typescript/src/scan-logs.ts
```diff
@@ -91,30 +103,46 @@ export async function findScanSession(
...
+  const pending = [...included];
+  for (const parentId of pending) {
+    const parent = logs.get(parentId);
+    if (parent !== undefined) sessions.push(parent);
     for (const session of logs.values()) {
       if (
         !included.has(session.threadId) &&
-        (session.parentThreadId === parent.threadId ||
-          (parent === root &&
+        (session.parentThreadId === parentId ||
```
mldangelo-oai, line 143: [P1] Scope Desktop owner descendants to the selected scan

The new `threadIds` set includes the Desktop owner conversation, but this parent-ID match follows every child and fork of that conversation without applying `belongsToScan`. If the conversation is reused for other work, `feedback <scan-id> --include-logs` attaches unrelated sessions, including sessions from a different repository before the scan or after it completed. I reproduced this through `sendFeedback()` using synthetic sessions and the local mock upload fixture: earlier and later unrelated children and an unrelated fork all appeared in the attachment.

Please separate including the owner log from discovering worker descendants, and traverse roots associated with the selected scan. Add a regression that excludes unrelated owner children/forks while preserving descendants of recorded scan workers, including when the owner log is absent.

daneschneider-oai: Fixed in [22e0047](https://github.com/openai/codex-security/commit/22e00471c703bd947d95fa230e8d83189f0a2c2e).

Confirmed the reproduction. The collector now includes the owner log directly and discovers descendants only from recorded scan execution roots. Desktop standard scans collect the owner log only; Deep and CLI scans retain recorded workers and their descendants, including when the owner log is absent. The shared owner ID is also omitted from the native feedback upload request so its subtree cannot be collected through that path.

Regression coverage excludes unrelated owner children and forks before, during, and after the scan, including sessions in another repository, and checks both the collected attachment and the native upload request. The final full SDK suite passed: 2,716 passed, 55 skipped, 0 failed. Focused regressions and 48 Python tests also passed.

mldangelo-oai (approved):

PR #902, opened 2026-09-12
test(cli): stabilize non-Git patch fixtures

mldangelo-oai:
## Summary

The non-Git patch test initialized and committed a repository, then deleted `.git`. On macOS CI, the case intermittently reached Git snapshotting and failed with `git read-tree HEAD` instead of exercising directory snapshots. Its no-op assertion could also accept that setup failure.

## Changes

- Create the non-Git cases as plain directories without initializing a repository or deleting Git metadata.
- Keep real Git detection and the existing successful-patch and no-op assertions.
- Require `NO_PATCH_APPLIED` for the no-op case and include captured stderr in exit-code failures.

## Testing

- Focused patch-result tests using Bun 1.3.14, Git 2.55.0, and the failing CI seed `1183541649`: 9 passed.
- Repeated the two non-Git cases 200 times each with the same tool versions and seed: 400 passed.
- `pnpm run types`: passed.
- `pnpm run format`: passed.
- `git diff --check`: passed.
- Previously injected a synthetic Git inspection failure: the original no-op assertion passed incorrectly, while the strengthened assertion rejected `PATCH_FAILED`. The injection is not part of this change.

## Risk and rollout

Test-only change with no runtime or public CLI changes. The CI log confirms that the intended non-Git case entered Git snapshotting; the underlying create/delete failure did not reproduce locally. Starting with a plain directory removes that unnecessary setup while preserving real command execution and strict outcome checks.

## Public disclosure review

### DIFF
--- a/sdk/typescript/tests-ts/cli-patch-results.test.ts
+++ b/sdk/typescript/tests-ts/cli-patch-results.test.ts
@@ -13,7 +13,7 @@ afterEach(async () => {
     await rm(directory, { recursive: true, force: true });
 });
 
-async function repositoryFixture() {
+async function repositoryFixture({ initializeGit = true } = {}) {
   const directory = await mkdtemp(join(tmpdir(), "patch-results-"));
   directories.push(directory);
   const runRepositoryCommand: NonNullable<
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
@codex review

Please review the current head, `7aba78aa4c81d5feff7dd10884fde37d5a21211e`.

mldangelo-oai:
@codex review

Please review the current head, `1a1869ef6a50a7c0164ddde27baf52f663ad3e90`. The follow-up creates plain directory fixtures for the non-Git cases and retains the strict failure-reason assertions.

PR #925, opened 2026-09-14
ci: declare native release workflow permissions

mldangelo-oai:
## Summary

The native job in the npm release workflow calls a reusable workflow without declaring its token permissions. Add an explicit `contents: read` permission at the call site to match the native artifact workflow and its platform builds.

## Changes

- Declare `contents: read` on the native release job.

## Testing

- `git diff --check` passed.
- Prettier check of `.github/workflows/node-release.yml` passed.
- Parsed the workflow before and after the change and verified that only the native job permission map changed.
- Recursively checked the four called native workflows: their permissions fit the caller's `contents: read` ceiling.

## Risk and rollout

This is a workflow configuration change with no CLI or SDK API changes. The called native workflows already use `contents: read`. The release workflow runs on release tags or manual dispatch; local validation was static and did not invoke a release.

## Public disclosure review


mldangelo-oai:
@codex review

PR #946, opened 2026-09-16
feat(dedupe): support host-provided records reviews

kmbroai:
## Summary

Dedupe currently requires a saved scan, a Findings API, and local model execution. Add a records entry point that lets a host supply findings and execute reviews while Codex Security validates the answers and returns persistence decisions.

## Changes

- Export `deduplicateRecords` and a serializable review-runner contract. Reuse the existing screening, independent pair review, grouping, and domain validators with serial execution.
- Add `codex-security dedupe --records`, used alone, for one versioned bidirectional JSON-RPC run over stdin/stdout. Correlate replies, handle cancellation and disconnects, and never retry host execution.
- Return new groups, existing canonical matches, or unresolved observations. Add protocol documentation and a runnable Python fake host.
- Keep host metadata in the records adapter; reuse Finding validation without adding a second algorithm or changing saved-scan checkpoints.

## Testing

- 30 focused SDK/CLI records tests passed, including real subprocess pipes and failure handling.
- Type checks, changed-file formatting, SDK build, and installed-package smoke checks passed, including public NodeNext types.
- The documented Python fake host passed against the built CLI with local Codex unavailable.
- Full SDK tests with seed 12345 and randomized order are still running. The seeded run has 11 authentication-fixture failures caused by pre-existing Git metadata at a hard-coded temporary path; all 11 reproduce on unchanged main. Policy fixtures pass with an isolated temporary directory.

## Risk and rollout

The new flag defaults to false; saved-scan dedupe keeps its existing runner and checkpoints. Records mode performs no candidate retrieval, local model execution, or persistence. A review failure leaves the batch unresolved, and a group connecting multiple existing canonicals requires host reconciliation. The host owns remote cancellation and persistence. Validation uses a fake host; no live remote backend integration was tested.

## Public disclosure review


ianw-oai (approved):

sdk/typescript/src/deduplication/records.ts
```diff
@@ -0,0 +1,196 @@
...
+    observations.get(id)!,
+  );
+  const reviewer = new CodexDeduplicationReviewer({
+    async run<T>({ validate, ...review }: CodexReview<T>): Promise<T> {
+      return await abortable(async () => {
+        options.signal?.throwIfAborted();
+        return validate(
+          await options.reviewRunner.run(
+            {
+              ...review,
```
mldangelo-oai: The `schema` sent to the host here is the local tool schema. In the emitted request, screening uses `oneOf`, and pair review uses a top-level `oneOf` plus `mergedFinding.additionalProperties: {}`. Those shapes are outside [OpenAI's strict Structured Outputs subset](https://developers.openai.com/api/docs/guides/structured-outputs), so passing the schema directly as the protocol docs recommend fails before the review can run. Could this adapter emit a schema the host can use directly, with a test for the emitted schema? I checked the emitted JSON locally; I did not send a live model request.

sdk/typescript/src/deduplication/records-protocol.ts
```diff
@@ -0,0 +1,188 @@
...
+                method: "review.run",
+                params: review,
+              }).catch(outputError);
+            });
+          },
+        },
+      }).then(
+        async (result) => {
+          if (finished) return;
+          finished = true;
```
mldangelo-oai, line 151: Setting `finished` before `send` flushes makes the cancellation handler return while the write is still pending. With the built CLI and a synthetic 50-anchor run, I left stdout unread and sent SIGTERM; the process stayed alive until SIGKILL. Could cancellation remain effective until the final write completes, without sending a second response?

sdk/typescript/src/deduplication/records.ts
```diff
@@ -0,0 +1,196 @@
...
+    options.signal,
+    1,
+  );
+  let groups: string[][];
+  try {
+    const decisions = await algorithm.run(
+      sourceFindings.map((finding) => finding.findingId),
+    );
+    options.signal?.throwIfAborted();
+    if (decisions.refusals?.length)
```
mldangelo-oai: If an SDK runner throws the exported `DeduplicationReviewError` with category `refusal`, `FindingDeduplicator` catches it and continues scheduling; this check only sees it after the remaining work has run. In a two-neighborhood test the first screening refused, the second still ran, and the entire batch returned `unresolved`. Since subsequent remote reviews cannot contribute to the result, could records mode stop scheduling on the first refusal?

sdk/typescript/src/deduplication/records-protocol.ts
```diff
@@ -0,0 +1,188 @@
+import { createInterface } from "node:readline";
+import { type Readable, Writable } from "node:stream";
+import { z } from "incur";
+import { deduplicateRecords, type DeduplicateRecordsInput } from "./records.js";
+
+const rpcId = z.union([z.string(), z.number().int()]);
```
mldangelo-oai, line 6: The protocol docs say request IDs can be integers, but `z.number().int()` accepts only JavaScript safe integers. Sending `9007199254740993` as the first run ID returned `-32600` with `id: null` and a mismatched-review-response error, so a host using a large numeric ID cannot correlate the response. It may be simplest to document the safe range and require strings outside it, or preserve larger IDs if they are meant to be supported.

sdk/typescript/src/deduplication/records.ts
kmbroai: Clarified the contract in ffc0616e. `schema` remains the existing tool-submission schema; the docs and SDK comment now explicitly distinguish it from strict Structured Outputs. The documented function-calling path uses `strict: false` and returns tool arguments for domain validation. Hosts requiring strict outputs must adapt their submission format while preserving the result contract and arbitrary finding evidence. This is a documentation correction, not a change that makes the emitted schema strict-compatible.

sdk/typescript/src/deduplication/records-protocol.ts
kmbroai: Fixed in ffc0616e. Cancellation remains effective while a terminal response is pending, without sending a second response. Records-mode CLI now exits explicitly after protocol writes have flushed or been canceled, since Node stdout ignores destroy(). Added regression coverage for blocked success/error responses and a child process with unread stdout. The built Node CLI reproduction with 50 anchors now exits 143 on SIGTERM without SIGKILL.

sdk/typescript/src/deduplication/records.ts
kmbroai: Fixed in ffc0616e at the records adapter boundary. A typed refusal now stops the shared scheduler immediately, and every anchor remains unresolved with no groups returned. Removed the redundant post-run refusal check; saved-scan refusal handling is unchanged. Screening and pair-review tests assert that no further reviews run after a refusal. The two-neighborhood reproduction now makes one host call instead of two.

sdk/typescript/src/deduplication/records-protocol.ts
kmbroai: Documented in ffc0616e: numeric request IDs must be safe integers from -9007199254740991 through 9007199254740991, and larger numeric identifiers must be sent as strings. Runtime parsing and correlation behavior are unchanged; no arbitrary-precision parsing was added.

sdk/typescript/src/deduplication/records.ts
```diff
@@ -0,0 +1,203 @@
...
+              {
+                ...review,
+                requestId: randomUUID(),
+                trustedInstructions: [
+                  sourceReviewInstructions,
+                  review.stage === "screening"
+                    ? screeningInstructions
+                    : pairReviewInstructions,
+                  review.stage === "screening"
+                    ? screeningFindingFormatInstructions
```
mldangelo-oai, line 135: Follow-up simplification: `screeningPrompt`/`pairReviewPrompt` already include the stage and finding-format instructions, and `trustedInstructions` adds the same text again here. I measured about 3.7k identical characters for screening and 4.8k for pair review on every call. For host reviews, could `trustedInstructions` carry those rules once and `prompt` carry only the assignment and records? The saved-scan prompt can keep its existing behavior. This would remove repeated text from every host request.

sdk/typescript/docs/dedupe-records.md
```diff
@@ -0,0 +1,305 @@
...
+The CLI never resubmits a review, including after this error. Invalid model
+answers, missing decisions, and host errors make the whole batch unresolved;
+previous successful reviews in that run do not become persistence decisions.
+The existing algorithm treats a validated `DISTINCT` as a nonmatch, when
+supplied finding content is insufficient to establish a shared correction. If the host considers the review
+inconclusive, it must return an error rather than manufacture a `DISTINCT` answer.
+No new `INCONCLUSIVE` model decision is introduced.
+
+The existing screening, independent pair review, contradiction-aware grouping, and
+severity-based representative selection are reused. No local review checkpoints
```
mldangelo-oai, line 178: Follow-up docs cleanup: could the host-ownership and failure rules be stated once? The opening, review, persistence, cancellation, and fake-host sections repeat parts of the same guidance; the lines here about reusing the algorithm and saved-scan checkpoints are also implementation details an adapter author does not need. I'd keep the wire examples, error/exit codes, and the new non-strict tool guidance, and cut the repeated caveats and internal workflow notes. That would make the actual host contract easier to find.

sdk/typescript/tests-ts/dedupe-records.test.ts
```diff
@@ -0,0 +1,650 @@
...
+          if (scenario === "malformed") return "not an object";
+          if (scenario === "inconclusive") return { decision: "INCONCLUSIVE" };
+          if (scenario === "unknown-canonical") {
+            if (review.stage === "screening") return answer(review);
+            return {
+              ...decision(assigned(review)),
+              canonicalFindingId: "unknown",
+            };
+          }
+          const valid = answer(review) as {
```
mldangelo-oai, line 205: Follow-up test reduction: `finding-deduplication.test.ts` already covers missing/extra screening slots and canonical IDs outside the assigned pair. Here I would keep one bad screening result, one bad pair result, and the thrown/refused-host cases to prove that records mode returns all anchors unresolved and does not retry; the other validator-specific scenarios and their dispatch branches could go. That would shorten this test while keeping coverage for the behavior this adapter adds.

sdk/typescript/src/deduplication/records.ts
```diff
@@ -0,0 +1,203 @@
...
+        .object({
+          observationId: id,
+          candidateObservationIds: z.array(id),
+        })
+        .strict(),
+    ),
+  })
+  .strict();
+
+export interface DeduplicateRecordsInput {
```
mldangelo-oai, line 42: Follow-up simplification: this interface repeats the input schema immediately above. Could the schema infer its own type and the exported input derive from `z.output<typeof deduplicateRecordsInputSchema>` instead? The finding transform already returns `Finding`; I checked that the inferred fields match this public type. That removes the duplicated shape and gives future input changes one place to update.

Issue #956, opened 2026-09-17
onWorkerStatus can omit dispatch events when delegated workers run


gpapakyriakopoulos:
## Summary

`onWorkerStatus` can emit a valid preflight status but omit every `dispatch`
status even when a successful Standard committed-diff scan actually uses delegated
workers.

The [SDK documentation](https://learn.chatgpt.com/docs/security/sdk#track-or-cancel-a-scan)
describes `onWorkerStatus` as being called when worker preflight or dispatch status
changes. In `0.1.28`, dispatch status is instead inferred from a marker in an
`agent_message`; it is not derived from the actual worker lifecycle. If the scan
agent omits or misformats that marker, the callback is silent.

## Environment

- `@openai/codex-security` `0.1.28`
- sealed manifest producer: `codex-security-plugin` `0.1.95`
- Node.js 24 on GitHub Actions Linux
- Standard committed-diff scan using `DiffTarget.refs(...)`
- `gpt-5.6-sol`, `low` reasoning effort

## Observed reproduction

Collect lifecycle events during a normal SDK scan:

```ts
const workerStatuses = [];
const progress = [];

const result = await security.run(repository, {
  target: DiffTarget.refs({ base, head }),
  outputDir,
  onWorkerStatus(status) {
    workerStatuses.push(status);
  },
  onProgress(update) {
    progress.push(update);
  },
});
```

The successful scan emitted this worker status:

```json
{"kind":"preflight","delegation":"available","configuredSlots":null}
```

It then emitted normal committed-diff progress through discovery, validation, and
reporting, completed with sealed artifacts and complete coverage, and its report
recorded the validation mode as:

```text
Static source review of complete base/head files and exact diff; two independent
delegated batches plus parent consistency review.
```

However, `workerStatuses` contained no event with `kind: "dispatch"` for
`file_review`, validation, or any other phase.

This is not a case where worker delegation was unavailable: preflight reported it
available and the completed scan recorded delegated batches.

## Expected behavior

When the runtime attempts a delegated phase, `onWorkerStatus` should receive a
runtime-derived dispatch event with the actual `phase`, `planned`, and `started`
counts, including `started: 0` when dispatch fails or capacity is unavailable.

If dispatch events are intentionally best-effort, the API documentation and types
should state that explicitly and the SDK should expose a separate deterministic
worker-lifecycle event suitable for CI control logic.

## Why this appears to be an SDK implementation gap

The current parser sends `command_execution` events to deterministic preflight
[...]

PR #962, opened 2026-09-17
fix(sdk): report observed worker sessions without status markers

mldangelo-oai:
## Summary

Worker dispatch status depends on assistant text markers, so SDK callers can miss successful delegation. Add `onWorkerEvent` to report persisted worker sessions discovered by the existing SDK session tracker, independently of model text. This works with the currently bundled Codex 0.154.0 and requires no runtime changes or dependency updates.

Refs #956. This provides session observation; phase counts, failed-dispatch reporting, and a pre-review control gate remain outside this callback's contract.

## Changes

- Export `ScanWorkerEvent` and add `ScanOptions.onWorkerEvent`, emitting only `{ kind: "observed", worker: number }` once per discovered session per run.
- Reuse session attribution and worker numbering from cost/activity tracking, including nested workers and associated validation or Deep Scan sessions. Enable polling when this is the only observer.
- Keep prompts, raw thread IDs, and session contents out of the callback. Route synchronous and asynchronous observer failures through `onObserverError` without stopping the scan.
- End observations after the first final session read, including cancellation during post-scan work. Document persistence/polling delay, saved workers on resume, and the post-scan boundary.
- Preserve `onWorkerStatus` behavior and explicitly document its dispatch variant as best-effort model reporting. No public CLI changes.

## Testing

- Marker-independent session regression: failed before implementation and passed afterward for all three supported parent metadata shapes.
- Final randomized SDK suite (seed `2454836527`): **3,083 passed, 57 skipped, zero failures**. Seed `12345` also passed (3,082 passed, 57 skipped) before adding the final cancelled-follow-up case.
- Focused API/session/status suite: 292 passed, 4 skipped. The subsequent cancellation/observation regression set passed all 22 cases; all three metadata-fixture variants passed after the race correction.
- SDK types, formatting, production build, and CI build: passed.
- npm archive and installed-package smoke checks on Node 24.15.0: passed, including strict NodeNext public types, CLI and SDK lifecycle, all 142 bundled plugin files, and nested worker startup. Installation used the existing dependency cache after an unchanged development dependency failed a registry lookup.
- Final compiled SDK with **unmodified Codex 0.154.0**: a synthetic Standard `DiffTarget.refs` scan created two real delegated sessions and delivered both observations before the mock model was allowed to complete. Verified matching session worker numbers, absent status markers, non-fatal throwing observers, sealed complete artifacts, and no observation for a capacity failure that created no session.
- Three independent native reviews plus independent verification passed on the final commit after correcting a cancelled-follow-up lifecycle edge and a test-fixture race.

The integration uses a localhost mock model, synthetic credentials, and a harmless temporary repository. Local execution coverage is macOS arm64; Linux and Windows remain CI coverage.

## Risk and rollout

This is an optional SDK observer backed by the same session discovery used for cost tracking. It is compatible with the existing dependency pins. Saved workers can be reported again on resume, and missing notifications do not establish that delegation was skipped. The event does not mean a worker just started or file review has begun, does not report failed spawn attempts or planned counts, and cannot gate dispatch. Observation ends before `postScanPrompt`; `maxCostUsd` and `AbortSignal` remain the cost/cancellation controls.

## Public disclosure review


mldangelo-oai:
@codex review

Please review head `b2e08d718d9ee1bf171f05f224550fb386b98e70`. This PR remains draft pending the paired runtime release and dependency pin update.

PR #970, opened 2026-09-18
fix(cost): report an inaccessible session log once, not on every poll

### LINKED ISSUE #223: An unreadable Codex session log fails every cost poll for the rest of the scan, and can fail a scan that succeeded
## Summary

`ScanCostTracker` quarantines a session log it cannot *parse*, but not one it cannot *open*. An unopenable file under `$CODEX_HOME/sessions` therefore fails on every poll for the life of the scan, and — because the failure happens before the session's thread id is known — it escapes the thread-tree filter that exists precisely to ignore unrelated sessions.

## Environment

- Reproduced against the exported `ScanCostTracker` with a `chmod 000` fixture; no network or credentials required

## The asymmetry

`readSessionUsage` in `sdk/typescript/src/cost.ts` has two failure exits:

```ts
if (session.unreadable) return;
let file;
try {
  file = await open(path, "r");
} catch (error) {
  if (isMissingFile(error)) return;
  throw error;                      // <-- nothing recorded
}
...
      } catch (error) {
        session.unreadable = true;  // <-- quarantined
        session.pendingLine = [];
        session.pendingLineBytes = 0;
        throw error;
      }
```

The oversized-event exit marks the session `unreadable`, so the guard at the top skips it on every later poll: one error, then the tracker moves on. The `open()` exit records nothing, so the next poll reopens the same file and fails identically, every `COST_POLL_INTERVAL_MS`.

`#readSessions` amplifies it. A failure is only deferred — and thus filtered by the `included` thread-tree walk — when the thread id is already known:

```ts
} catch (error) {
  if (session.threadId === null) throw error;
  unreadable.push({ session, error });
}
```

An oversized *unrelated* session gets deferred, because its `session_meta` line was parsed before the oversized line failed. An *unopenable* session never got a thread id at all, so the error escapes the filter entirely.

The suite already pins the tolerant half of this: `tests-ts/cost.test.ts` has "ignores oversized events from unrelated prior credential sessions". An unrelated session that is too big is ignored; the same unrelated session that merely cannot be opened aborts the scan.

## Trigger

Any file under `$CODEX_HOME/sessions` the process cannot open — most plausibly a root-owned rollout left behind by a single `sudo codex` run, but also a shared home or restrictive ACLs.

## Impact, via `src/api.ts`

- **With `--max-cost`**: `onError: (error) => costAbortController.abort(error)` aborts the scan on the first poll, effectively before it starts.
- **Without a cost limit**: `onFinalize` awaits `tracker.stop(usage)`, which rejects — so a scan that **completed successfully** is reported as failed.
- **For SDK consumers whose `onError` does not abort**: `#readSessions` throws before the usage aggregation runs, so `#snapshot` never updates and `onCost` never fires. The `--max-cost` ceiling is then never evaluated at all, while the same error repeats every poll.

## Secondary: the documented fallback is unreachable

`stop(fallbackUsage)` calls `await this.refresh()` unguarded, so when refresh rejects, `stop()` rejects and the caller-supplied `fallbackUsage` — the completed turn's own authoritative usage — is discarded. The test named "falls back to the completed turn when session logs are unavailable" documents the intent, but the fallback cannot be reached whenever the sessions directory exists and cannot be scanned.

## Expected behavior

An unopenable session log should be reported once and then quarantined, exactly like an unparseable one, rather than failing every poll. A scan whose turn completed should not be reported as failed because of it.

I have a fix and will open a PR referencing this issue.

## One related gap, not covered by that fix

`sessionFiles` has the same shape at directory level: a non-ENOENT `readdir` error has no per-directory state to quarantine, so it would also repeat on every poll. It is outside the trigger above (`sudo codex` leaves directories readable and only the rollout file root-owned) and fixing it needs new state rather than reuse of the existing flag, so I left it alone. Happy to follow up if you want it covered.

feiiiiii5:
## Summary

An inaccessible session log under `$CODEX_HOME/sessions` fails on every cost poll for the rest of the scan, and can fail a scan that already completed.

`readSessionUsage` has two failure exits and only one of them records that the session is bad. A log it cannot parse sets `session.unreadable`, so the guard at the top of the function skips that file on later polls: one error, then the tracker moves on. A log it cannot **open** records nothing, so the next poll opens the same file and fails identically every `COST_POLL_INTERVAL_MS`.

The failure also escapes the thread-tree filter in `#readSessions`. A deferral is filtered by `included` only when the thread id is known, and an unopenable log never produced one, so `if (session.threadId === null) throw error;` rethrows instead of deferring — for a file that can never be attributed to this scan.

Because `stop()` awaits `refresh()` unguarded, the final poll can reject after the turn already completed, discarding the caller's `fallbackUsage` and reporting a successful scan as failed.

Reproduced on `main` at `70d5b2ed` with a real `chmod 0o000` rollout file: four `onError` reports within 300 ms, and `stop()` rejecting with `EACCES: permission denied, open '.../sessions/2026/07/26/rollout-prior-thread.jsonl'` from `src/cost.ts:191`.

Closes #223. #224 proposed a fix for this and was closed by its author on 2026-09-01 with no maintainer decision recorded; this is an independent implementation against current `main`, and it differs in the respect described under Changes.

## Changes

- `readSessionUsage` now quarantines the session when `open()` fails with an access refusal the process cannot clear by retrying (`EACCES`, `EPERM`), mirroring the existing parse-failure exit. The error is still thrown, so it is still reported — once.
- Both failure exits now share a `quarantineSession()` helper. The parse-failure path is unchanged in behavior.
- Every other open failure keeps the current retry-and-report behavior. A process-wide shortage such as `EMFILE` clears on its own, and retiring a session for it would stop observing usage that could still be read, which would weaken the `--max-cost` enforcement the tracker exists for. No attempt counter or other new limit is introduced.

Not changed, deliberately: `stop()` still awaits `refresh()` unguarded, so a failure that first appears exactly at the final refresh would still discard `fallbackUsage`. With inaccessible logs quarantined, the reported failure mode no longer reaches that path. Hardening `stop()` against unrelated refresh failures is a separate question and is left out of this change.

## Testing

From `sdk/typescript` on macOS arm64, node 24.19.0, bun 1.3.14, pnpm 11.11.0:

- `bun test --timeout 30000 tests-ts/cost.test.ts -t inaccessible` at `70d5b2ed` **without** the `src/cost.ts` change: **2 fail** — `expect(received).toHaveLength(expected) / Expected length: 1 / Received length: 4`, and `stop()` rejecting with `EACCES: permission denied, open '/…/rollout-prior-thread.jsonl'` at `async stop (src/cost.ts:191:16)`.
- The same command **with** the change: **2 pass**.
- `bun test --timeout 30000 tests-ts/cost.test.ts`: **74 pass, 0 fail** (72 before the two new tests).
- Mutation check: deleting only the new `quarantineSession(session)` call from the open-failure path makes both new tests fail again; restoring it makes them pass. So the tests pin this change rather than the surrounding code.
- `bun test --timeout 30000 tests-ts/cost-context.test.ts tests-ts/cost-model.property.test.ts tests-ts/cli-show-cost.test.ts`: **24 pass, 0 fail**.
- `pnpm run types` rc=0, `pnpm run lint` rc=0, `pnpm run format` rc=0 (`All matched files use Prettier code style!`).
- `bun test --timeout 30000 tests-ts/api.test.ts`: **54 pass, 4 skip, 109 fail, identical with and without this change** — the same three counts measured at base and at head. Those failures are `ModuleNotFoundError: No module named 'workbench_scan_usage'` from the generated `_bundled_plugin` payload, which I could not build here: `pnpm run build:plugin` stops at `plugins/codex-security/native/prebuilt/COPYRIGHT-library.html`, and preparing that payload needs the pinned Rust toolchain. For the same reason `pnpm run test` was not run locally; the focused module suites above were run instead.

The two new tests are gated `test.skipIf(process.platform === "win32" || process.getuid?.() === 0)`, following the existing permission fixture in `tests-ts/security-policy-helper.test.ts:203`, because the fixture is a POSIX permission bit. Both restore the mode in a `finally` block so the shared temporary-directory cleanup can remove the fixture.

## Risk and rollout

No public CLI or SDK surface changes, and no new limit, flag, or configuration. Behavior changes only for a session log that cannot be opened: it is reported once and then skipped, instead of being reported on every poll for the life of the scan. A quarantined session is not re-read if it later becomes accessible — that is the contract the parse-failure path already has. Missing files (`ENOENT`) are unchanged, and transient open errors keep retrying as before. No release, migration, or rollout ordering impact.

## Public disclosure review


feiiiiii5:
Flagging an overlap I should have caught before opening this.

Staff PR #465 — `fix(cost): enforce verifiable scan spending limits`, open since 2026-08-15 — already covers this root cause. Its `sdk/typescript/src/cost.ts` diff adds the same predicate and the same helper this PR introduces:

```ts
function isSessionAccessDenied(error: unknown): boolean {
  return (
    isRecord(error) && (error["code"] === "EACCES" || error["code"] === "EPERM")
  );
}
```

and calls `quarantineSession(session, error)` at both sites this PR touches: the `open()` failure inside `readSessionUsage` (`if (session.threadId === null && isSessionAccessDenied(error))`), and the `unreadable` loop in `#readSessions` (`else if (isSessionAccessDenied(error))`). It also reworks `session.unreadable` from a boolean into `{ error }` and adds `requireReadableSessions` / `readFailures`, which looks like it reaches the `stop()` interaction I listed under "Not changed, deliberately". I have not traced that path end to end, so treat that last part as unverified.

To be precise about the current state: `main` as of writing has no `quarantineSession`, no `isSessionAccessDenied` and no `EACCES` handling in `sdk/typescript/src/cost.ts`, so #223 is still unfixed and this PR is not stale today.

If #465 lands first, though, this one is redundant and should be closed rather than rebased. I have not closed it myself because #465 is still open, so the fix is not on `main` yet. Happy to close on a maintainer's word, or to narrow this to whatever #465 leaves uncovered if that is more useful than closing it.

The note in the description about #224 stands: it was closed by its author on 2026-09-01 with no maintainer decision recorded.

Issue #538, opened 2026-08-18
CLI: failed structured history commands emit an ok: true envelope


soyeon-oai:
## Summary

A failed structured history command exits non-zero and reports an error on stderr, but still emits an `ok: true` envelope on stdout.

## Reproduction

```sh
codex-security scans show 51e818d --format json --full-output
```

The seven-character scan ID prefix is invalid because prefixes must contain at least eight characters.

## Actual behavior

Exit code:

```text
2
```

stderr:

```text
codex-security: Could not read Codex Security scan history: Scan ID prefixes must be at least eight characters.
```

stdout:

```json
{
  "ok": true,
  "meta": {
    "command": "scans show",
    "duration": "97ms"
  }
}
```

## Expected behavior

A failed command must not emit an `ok: true` structured response. It should emit either an `ok: false` error envelope or no structured stdout, consistently with the CLI error contract.

## Impact

Automation that relies on structured stdout can incorrectly classify the failed command as successful.

## Environment

- `@openai/codex-security`: `0.1.14`
- Reproduced twice on PR #439 head `34e687e5beda004dc9af585d8f375c5147187ffe`
- Also reproduced on base `37bf87a692fc72d41f7312cc48808d699d204fba`, so this is a pre-existing issue

pasmud:
Fixed by #562: failed structured history commands now emit an `ok: false` error envelope instead of `ok: true`.

soyeon-oai:
Verified fixed on current `main` at `c61b46e16a2cbba53be8a93c24706deaa4b073a4` using a packed source build under Node.js 24.11.1.

Re-running the reported seven-character scan ID prefix case now produces:

- exit code `2`
- the expected “Scan ID prefixes must be at least eight characters” error on stderr
- empty structured stdout, rather than an `ok: true` envelope

Closing as fixed by #562.

PR #971, opened 2026-09-18
fix(cli): report full-output scan failures as errors

thegoodengineer:
## Summary

Since #709, a failed `scan` with `--json`, `--format json`, or `--format jsonl` prints the documented `SCAN_FAILED` object on stdout. The handler does this by returning the object as command data, so when `--full-output` is also set, Incur wraps it in a success envelope. A failing `codex-security scan . --json --full-output` exits with `2` but prints:

```json
{
  "ok": true,
  "data": {
    "status": "failed",
    "code": "SCAN_FAILED",
    "message": "..."
  },
  "meta": { "command": "scan", "duration": "4ms" }
}
```

Automation that reads the envelope sees `ok: true` for a failed scan. #538 set the rule for this case: a failed command must not emit an `ok: true` structured response. With this change the failure uses the standard error envelope instead:

```json
{
  "ok": false,
  "error": { "code": "SCAN_FAILED", "message": "..." },
  "meta": { "command": "scan", "duration": "4ms" }
}
```

## Changes

- In the `scan` handler, when the output format is JSON or JSONL and `--full-output` is present, return the failure through Incur's error path instead of as data. The message still goes through `safeErrorMessage` because it is written to stdout.
- In `main()`, keep that error envelope on stdout with a `scanStructuredError` flag, the same way `patch` keeps its `--full-output` failures (`patchStructuredError`).
- Output without `--full-output` is unchanged: JSON and JSONL still print `{"status": "failed", "code": "SCAN_FAILED", "message": ...}`, and the other formats still print nothing on stdout. Exit codes are unchanged, including `130` and `143`.
- Document the `--full-output` failure shape in the SDK README, next to the existing structured failure section.
- Add regression tests for `--json`, `--format json`, and `--format jsonl` with `--full-output`, plus a credential redaction check for the envelope message.

## Testing

- The two new tests fail on `main` (`envelope.ok` is `true` and there is no `error` object) and pass with this change.
- A local check with the existing `FakeSignals` fixture: `scan . --json --full-output` interrupted by `SIGINT` and `SIGTERM` exits with `130` and `143` and prints the `ok: false` envelope. Plain `--json` output for the same interruptions is unchanged.
- `bun test --timeout 120000` on `tests-ts/cli.test.ts`, `cli-authentication.test.ts`, `cli-project-config.test.ts`, `cli-policy.test.ts`, `cli-patch.test.ts`, and `cli-patch-results.test.ts` (Windows 11, Node 22.17, Bun 1.3.14): 269 pass, 1 skip, 99 fail. Unmodified `main` gives 267 pass, 1 skip, and the same 99 failures in this environment, because the bundled plugin is not built locally (it needs the CI native payload). No test fails only with this change.
- `pnpm run types`: passed.
- `pnpm run format`: passed.
- `pnpm run build:ci`: passed.
- `git diff --check`: passed.

## Risk and rollout

Low. Only failed scans that combine `--full-output` with JSON or JSONL output change: they now get Incur's `ok: false` error envelope with the same code and message, which matches how `patch` and `policy` already report `--full-output` failures. No new flags, schema changes, or plugin changes.

Draft #463 proposes a broader rework of structured scan failures. This change is limited to the behavior already on `main`.

## Public disclosure review


mldangelo-oai (approved):
Thanks for catching this and for covering the JSON and JSONL forms along with credential redaction. I verified that execution failures now emit `ok: false` and still exit with code 2, while the existing plain JSON shape is preserved. This is a useful, focused fix. I’ll follow up separately on incomplete-coverage results, which already used a success envelope on main.

